PR #1, opened 2016-10-15
refactor and eupsify repo

jchiang87:
These changes enable this package to be setup using eups.   I've refactored the parsing code, adding some validation of the input commands, and added unit tests.


SConstruct
```diff
@@ -0,0 +1,4 @@
+# -*- python -*-
```
cwwalter: This is probably just my ignorance but why do we need SContruct files for a python only program?  Do we need this fore eups somehow?

bin/imsim.py
```diff
@@ -0,0 +1,120 @@
...
+                        help='output directory for eimage file')
+    arguments = parser.parse_args()
+
+    # Get the number of rows to read from the instance file.  Use
+    # default if not specified.
+    numRows = arguments.numrows
+    logger.info("Reading %i rows from the instance catalog %s.",
+                numRows, arguments.file)
+
+    # The PhoSim instance file contains both pointing commands and
```
cwwalter: This check:

   # Get the instance file name. A valid file must be specified.
    try:
        with open(arguments.file) as instanceFile:
            print 'Using', instanceFile.name, "as an instance file."
            pass
    except IOError as errorMessage:
        print errorMessage
        print "That instance file cannot be opened. Program will end.\n"
        sys.exit()

was left out from my code.  This means that if you run the program with no arguments it crashes.

python/desc/imsim/imSim.py
```diff
@@ -1,39 +1,43 @@
...
-This is the imSim program, used to drive GalSim to simulate the LSST.  Written
-for the DESC collaboration and LSST project.  This version of the program can
-read phoSim instance files as is. It leverages the LSST Sims GalSim interface
-code found in sims_GalSimInterface.
-"""
-import os
 import sys
-import logging
-import argparse
+from collections import namedtuple
```
cwwalter: Since this file now just contains the parser should it have Parser type name (instead if imSim.py) instead?  Or is the intention to put a lot of code in here?

python/desc/imsim/imSim.py
```diff
@@ -198,105 +206,5 @@ def parsePhoSimInstanceFile(fileName, numRows):
...
-    # Write out the fits files
-    output_dir = arguments.outdir
-    if not os.path.isdir(output_dir):
-        os.makedirs(output_dir)
-    phoSimGalaxyCatalog.write_images(nameRoot=os.path.join(output_dir, 'e-image_')+visitIDString)
-
-if __name__ == "__main__":
-    main(sys.argv)
+    return PhoSimInstanceCatalogContents(commandDictionary, phoSimObjectList,
+                                         phoSimHeaderCards, phoSimSources)
```
cwwalter: Is there a reason to return more than the commandDictionary and phoSimDataBase?  I notice when you call this you use ) [:2] to just get the those.

bin/imsim.py
```diff
@@ -0,0 +1,120 @@
...
+from lsst.obs.lsstSim import LsstSimMapper
+
+from lsst.sims.photUtils import LSSTdefaults
+from lsst.sims.utils import ObservationMetaData
+
+from lsst.sims.GalSimInterface.galSimCatalogs import GalSimBase
+from lsst.sims.GalSimInterface import GalSimStars
+from lsst.sims.GalSimInterface import GalSimGalaxies
+
+import desc.imsim
```
cwwalter: I noticed if I didn't actually set this up (with eups) and just tried to run it, this caused the interpreter to fail.  Is that expected?

bin/imsim.py
```diff
@@ -0,0 +1,120 @@
+#!/usr/bin/env python
```
cwwalter: On Mac this is causing the program to fail.  I believe this is do to SIP stripping the LIBRARY_PATH from the sub process.  We went through a lot of this before.  @timj can probably tell us the "approved" DM solution.  In the mean time running this as "python bin/imsim" works.

bin/imsim.py
```diff
@@ -0,0 +1,120 @@
...
+    commandDictionary, phoSimObjectList = \
+        desc.imsim.parsePhoSimInstanceFile(arguments.file, numRows)[:2]
+
+    # Access the relevant pointing commands from the dictionary.
+    visitIDString = str(int(commandDictionary['obshistid'][0]))
+    mjd = commandDictionary['mjd'][0]
+    rightAscension = commandDictionary['rightascension'][0]
+    declination = commandDictionary['declination'][0]
+    rotSkyPosition = commandDictionary['rotskypos'][0]
+    bandpass = 'ugrizy'[int(commandDictionary['filter'][0])]
```
cwwalter: Unfortunately I can't figure out how to comment on code that wasn't actually changed but notice in monkeyPatchedGalSimBase.py this code:


    # Add bandpasses to simulate over.
    self.bandpassNames = ['r']
    # self.bandpassNames = self.obs_metadata.bandpassName

This means no matter what currently the r band will be used.  We need to tie it to the bandpass passed through the observationmetadata.

python/desc/imsim/imSim.py
jchiang87: The other two items were `global` in the original code so that the caller could have access. In order to avoid that, I made them part of the returning tuple so that they were still available.

bin/imsim.py
jchiang87: Exactly. If the code is going to crash in this function, I'd rather it throw the exception and have the standard exception handling output the diagnostics, and if uncaught, produce a non-zero return code so that job handlers like lsf can report that the code crashed.  Also, since this is now a function in a separate module, the calling code has the option of catching the IOError and dealing with it or not.

python/desc/imsim/imSim.py
```diff
@@ -1,39 +1,43 @@
...
+vistime""".split())
 
 def parsePhoSimInstanceFile(fileName, numRows):
     """
-    Read a PhoSim instance catalog into a Pandas dataFrame.  Then use the
-    information that was read-in to build and return a command dictionary and
-    object dataFrame.
+    Read a PhoSim instance catalog into some Pandas dataFrames.  Then
+    use the information that was read-in to build and return a
+    GalSimCelestialObject.
```
jchiang87: ok. I'll update it.

bin/imsim.py
jchiang87: ok.  We can update the README with this info, though it would be good to get a proper fix for this.

bin/imsim.py
jchiang87: Yes, I saw that after I committed.  I try to update this too.

bin/imsim.py
jchiang87: It should fail because the user's PYTHONPATH won't contain the directory with that module, but if you do
```
export PYTHONPATH=<imSim dir>/python/desc/imsim:${PYTHONPATH}
```
then I think it should work.

cwwalter:
OK.  This all looks good.  I still have some questions on some of the issues above but I can ask you about them on slack.  I'll go ahead an merge.

PR #2, opened 2016-10-18
speed up the parser and simulation code

jchiang87:
Summary of changes:
* In `parsePhoSimInstanceFile`, I replaced the loop over rows of the `phoSimSources` df with array operations instead in `extract_objects`.  It now takes a few seconds to do a 320000 line instance catalog vs hours using the loop.
*  Added a command line option to specify a single sensor to simulate (this will crash in a non-obvious way if the selected sensor doesn't have any sources on it...I'll fix that later.)  Doing a single sensor speeds things up by a factor of several for the Twinkles Run1.1 instance catalogs, and reduces the memory footprint substantially.  I think that the memory usage was the cause of the crashes (on a machine with 16GB memory) that I saw when I tried to run a Run1.1 catalog.
* Added some instrumentation to track memory usage in `get_phoSimInstanceCatalog`.  To use it, one needs the `psutil` python module.  If that module isn't available, the memory tracking is disabled.

@cwwalter Please have a look.


cwwalter:
This all looks great.  I'll merge to master.

jchiang87:
Running this code on the v200-fr visit for Twinkles Run1.1, using
```
imsim.py phosim_input_200.txt --outdir=v200-fr --sensor "R:2,2 S:1,1"
```
everything runs in about ~45mins on a slac batch worker (hequ0108) and uses max. memory of 9.7 GB.
Here's a comparison with the Run1.1 image generated with phosim 3.4.2:
galsim_phosim_comparison_v200_twinkles
The panel on the left is `phosim`, the right `imsim.py`.  Note that I took the xy transpose of the `imsim.py` image since `phosim` swaps those.   Note also that the galaxy position angles are different, as @danielsf stated via email.

jchiang87:
These files are available at SLAC.   The `imsim.py` file is
```
/nfs/farm/g/desc/u1/users/jchiang/desc_projects/imSim/tests/v200-fr/e-image_200_R_2_2_S_1_1_r.fits
```
The phosim Run1.1 file (at least the one that seems to be the one we used) is at
```
/nfs/farm/g/desc/u1/Pipeline-tasks/Twinkles-phoSimII/phosim_output/000000/output/lsst_e_200_f2_R22_S11_E000.fits.gz
```

sethdigel:
The Twinkles phosim 3.5.2 run of this visit took about 10x longer, 29.4 ks CPU time on a hequ (i.e., reasonably fast) host.  phosim includes sources from the instance catalogs that are within a flux-dependent buffer region around the sensor being simulated.  The minimum buffer size is 100 pixels.  imsim.py is simulating only sources that are within the (active area?) of the sensor, so the speed up is not quite as much as it looks like.

jchiang87:
It seems to me that not simulating photons that aren't really needed is a perfectly valid way to speed up a simulation.

sethdigel:
Right.  I was just pointing out that the comparison was not quite apples to apples in terms of the numbers of sources.  I can't say that I've studied what size buffer region is needed, or how to define 'need' in terms of accuracy of the simulation.  John maintains that the definition of the buffer region in phosim is not conservative, but I don't know by what measure.

cwwalter:
In terms of the issue of the brightness being too low..   @danielsf should comment:  We are not dealing with a true chromatic object here right?  The normalization is calculated in in calcADU in

https://github.com/lsst/sims_photUtils/blob/b3f515d2d9f5d31223238acd3c7bdc5946c95e2c/python/lsst/sims/photUtils/Sed.py

by doing a integral over the bandpass.  What wavelength is actually used for the simulation?  A weighted average?   I will follow up with Ji Won about the chromatic photon shooting code.  The python version is supposed to be ready now.

dkirkby:
Do these images show a single chip?  What is the exposure time and filter?  The mean sky ~800 ADU/pix would be about right for 30s in r-band.  Sources in the PhoSim image seems significantly brighter / larger (could be either with this much clipping), but the sky levels are much closer (with PhoSim perhaps a bit darker?)

cwwalter:
> Do these images show a single chip? What is the exposure time and filter? The mean sky ~800 ADU/pix would be about right for 30s in r-band. Sources in the PhoSim image seems significantly brighter / larger (could be either with this much clipping), but the sky levels are much closer (with PhoSim perhaps a bit darker?)

This is a single chip in u-band.  @danielsf is it for a 30 second exposure?  Where is that set in the SED calculation code?

dkirkby:
30s of u-band should be closer to 74 ADU/pix of sky (assuming 22.9 mag/sq.arcsec.)

jchiang87:
Actually they are just part of a single chip, the central 1/4 or so.   They are r-band, based on the filenames,`e-image_200_R_2_2_S_1_1_r.fits` and `lsst_e_200_f2_R22_S11_E000.fits.gz`, and 30 sec exposures each.  The phosim image has about 800ADU/pix for the sky background.   The imsim.py image has about 400ADU/pix/

cwwalter:
Ah... sorry I was remembering our conversation last night.  So, this is driven by what is in the PhoSim file filter # which is 'r' in this case I guess.

cwwalter:
Maybe this is 15 seconds?

Sky bkg is set by:

  self.noise_and_background = ExampleCCDNoise(addNoise=True,
                                                addBackground=True)

jchiang87:
The instance catalog and FITS header both say the exposure time is 30s. I'd have to dig into the code to see how that gets transmitted to the sky bkg code.  The galaxies are about a factor of 2 fainter, so maybe there is some bug and these really are 15 sec exposures.

dkirkby:
If the colorbar applies to both images, then the RH sky looks slightly brighter then the LH sky, not a factor of 2 darker.

cwwalter:
And that code is here:

https://github.com/lsst/sims_GalSimInterface/blob/56d725e0a73b4a97d8b3c5581aeb95ea19feeaab/python/lsst/sims/GalSimInterface/galSimNoiseAndBackground.py

jchiang87:
@dkirkby Probably the color bar doesn't apply the same to both images. I'm never sure how ds9 handles these things if a scale is set in one frame and one reads in another image in another frame.  In any case, I'm reading the pixel values directly.

cwwalter:
The sky background is calculated in 

calcSkyCountsPerPixelForM5

in

https://github.com/lsst/sims_photUtils/blob/b3f515d2d9f5d31223238acd3c7bdc5946c95e2c/python/lsst/sims/photUtils/SignalToNoise.py

We are setting the m5 level by passing  m5(bandpass) from LSSTDefaults() to the observationMetadata.

jchiang87:
@cwwalter This totally slipped by me, but I don't see the exposure time being set anywhere in the imSim code.  I have no idea how the EXPTIME header keyword got set to 30 sec, but the fact that the galaxies are also have a factor of 2 fewer counts leads me to believe that these are 15s exposures.  @danielsf Does this sound reasonable?

dkirkby:
Its ok for this test, but I don't think you want to use `calcSkyCountsPerPixelForM5` for normal simulations, since it sets the sky level in terms of a desired SNR rather than a sky SED model, and requires assumptions about the PSF which might not be consistent with the simulated PSF.

dkirkby:
I was going to look at the FITS files Jim mentioned above, but I don't know how to access files under `/nfs/farm/g/desc/`.  Is this volume web accessible or how do DESC members get the necessary SLAC account?

jchiang87:
you would need a slac account, but I will put them in my nersc account.  I'll post here when they are available.

There's a sign up procedure for getting a slac account, but I'll have to look for the page....it's somewhere in the desc confluence area.

danielsf:
In CatSim, exposure time is carried around in a class called `PhotometricParameters`.  In `sims_GalSimInterface`, the `galSimCatalog` has a `self.photParams` member that is an instantiation of this class.  I haven't had a chance to look at the ImSim code, so I'm not sure what you chose to do about that, but: any time you pass around a `PhotometricParameters` object, you should be passing around exposure time.

jchiang87:
I've put the files at NERSC in `/global/u1/j/jchiang8/imSim`.  The directory and files have permissions for the `lsst` group.  Let me know if you have access problems.

The page with the signup for a slac account for desc members is

https://confluence.slac.stanford.edu/display/LSSTDESC/Getting+a+SLAC+Computing+Account

cwwalter:
> In CatSim, exposure time is carried around in a class called PhotometricParameters. In sims_GalSimInterface, the galSimCatalog has a self.photParams member that is an instantiation of this class. I haven't had a chance to look at the ImSim code, so I'm not sure what you chose to do about that, but: any time you pass around a PhotometricParameters object, you should be passing around exposure time.

Hmm... I didn't notice that in any of the examples so didn't do anything with it.  We are probably getting whatever the default is.

danielsf:
The FITS header is written in `sims_GalSimInterface/.../galSimDetector.py`.  The value in the fits header is set to (number of exposures)*(exposure time).  I think the question of what to do about the fact that we are taking two snaps and stacking them has remained unresolved in CatSim (the `calcADU` method in `sims_photUtils/.../Sed.py` also takes nexp*exptime).

danielsf:
The default will be 2 15 second exposures

cwwalter:
> The default will be 2 15 second exposures

Stacked?

danielsf:
Yes

If everything is off by a factor of 2, note: the default for sims_GalSimInterface is to produce images in terms of ADU, not electrons, so there should be a factor of gain=2.3 difference between ImSim images and PhoSim e-images

jchiang87:
@danielsf good point.  The phosim images should be electrons.

jchiang87:
I will set the exposure and nsnaps to match the instance catalog, but we should also set gain=1, since we really expect eimages to be electrons/pix.   It would be good to turn off read noise, too, since we plan to run the electronics readout code (where we apply nominal gains, etc.) after generating eimages.

danielsf:
The `PhotometricParameters` instantiation is passed to the `GalSimDetector`, which is a class representing a chip.

PR #4, opened 2016-10-18
set photometric parameters based on instance catalog

jchiang87:


jchiang87:
Here's a comparison of sims of visit 200 from Twinkles Run1.1 for phosim 3.4.2 (left) and imsim.py (right).  The color scale applies to the righthand image.
imsim_vs_phosim_v200_comp
(This time I've zy-transposed the phosim image.)
I've set gain=1, exptime=30s, nsnap=1, and turned off read noise and dark current.

The background is higher in the imsim image, but the counts for the galaxies look roughly consistent between the two images.

The new imsim image is at NERSC:
```
/global/u1/j/jchiang8/imSim/e-image_200_R_2_2_S_1_1_r_photpars.fits
```

@danielsf Could you look at how I set the photometric parameters [here](https://github.com/DarkEnergyScienceCollaboration/imSim/pull/4/files#diff-4977a14fa564921b8b7b4d88d9a3cce2R206) and [here](https://github.com/DarkEnergyScienceCollaboration/imSim/pull/4/files#diff-0d54e1cdc3757c4f6e224fb8dbbc10a5R128)  and comment please?

If you could also comment on how we are applying the sky noise,

https://github.com/DarkEnergyScienceCollaboration/imSim/blob/issue/3/set_photometric_parameters/python/desc/imsim/monkeyPatchedGalSimBase.py#L98

that might help resolve the differences with the phosim image.

danielsf:
@jchiang87 I think you are setting-up `PhotometricParameters` correctly.

`ExampleCCDNoise` is a thin wrapper of GalSim's `CCDNoise` class.  The way that I think it should be working is: it calculates a mean number of sky counts with the `sims_photUtils` method `calcSkyCountsForM5`, and then uses `CCDNoise` to add Poisson noise on top of that.

Have we verified that the m5 values for the two images agree.  One difficulty comparing background values between PhoSim and GalSim is that PhoSim generates backgrounds from its own pseudo-random sky brightness model, so you cannot tune it to "agree" with GalSim.  It is possible that PhoSim just generated a sky brightness that is dimmer than GalSim's.  It is also possible that something is wrong in sims_GalSimInterface.

dkirkby:
I am getting `Permission denied` on
```
/global/u1/j/jchiang8/imSim/e-image_200_R_2_2_S_1_1_r_photpars.fits
```

jchiang87:
Maybe you're not in the `lsst` group?  In any case, I've enabled `others` access to the directory and its files.

dkirkby:
Thanks @jchiang87, I can read the files now.

A few observations from comparing this new galsim image with the original phosim image:
* Source centroids in the phosim image are shifted by about 15 pixels along the axis with length 4000 (relative to the galsim image).
* The median galsim sky is 991e, compared with 802e in phosim (galsim / phosim = 1.24).
* The ratio of the galsim / phosim source flux is 0.78 (so in the opposite direction to the sky discrepancy)
* Phosim fluxes are all exactly integer, as expected if no dark current or readout noise is included, but ~34% of galsim fluxes have a non-integral residual with an RMS ~ 0.09e.
* Sky fluctuations are consistent with Poisson for both images, i.e., (sky - sky0)/sqrt(sky0) has RMS=1.

dkirkby:
For reference, here is a 1D plot that shows the 15pix shift and the different sky and source responses, which should be easy to reproduce as we move towards better agreement:
```
pimg = fitsio.read('phosim.fits').transpose()
gimg = fitsio.read('galsim.fits')
plt.plot(gimg[390:470, 130], 'r-', label='GalSim')
plt.plot(pimg[390:470, 130], 'b-', label='PhoSim')
```
slice

The PSFs don't look very similar here.  Was anything done to try and match them?

cwwalter:
For the shift.  @danielsf and @SimonKrughoff should comment but there is some subtle issue where the phosim positions are not really ICRS.  It has to do with whether the positions are before or after refraction.  This causes a problem when doing the astrometry.   I don't yet have  a good enough handle on this to understand yet whether there should be a relative difference between PhoSim and GalSim.

> The PSFs don't look very similar here. Was anything done to try and match them?

Not really.  First of all, remember you don't actually set a PSF in PhoSim. It is emergent after all of the ray tracing in the atmosphere and optics.   So it is a function of the atmospheric parameters etc at some level but you don't explicitly set a size etc.

For imSim it is using a GalSim PSF. In this case I set it to the PSF that Scott provided as a default for the GalSim interface:

# Add a PSF.  This one Taken from equation 30 of
# www.astro.washington.edu/users/ivezic/Astr511/LSST_SNRdoc.pdf
self.PSF = SNRdocumentPSF()

You can find the explanation in the paper above and the code here:

https://github.com/lsst/sims_GalSimInterface/blob/56d725e0a73b4a97d8b3c5581aeb95ea19feeaab/python/lsst/sims/GalSimInterface/galSimPSF.py

You can also set the seeing in the metadata (which I set to 0.7" to try to be close to the expected result) but looking this code now it doesn't look like that is used.

rmandelb:
Is there any reason not to run PhoSim first, measure the PSF, and then use the PSF from PhoSim for the GalSim run?  Sorry, I am jumping in after missing most of the conversation, but this seems like a way to make matched images despite the whole "PSF is emergent in PhoSim" feature.

cwwalter:
> Is there any reason not to run PhoSim first, measure the PSF, and then use the PSF from PhoSim for the GalSim run? Sorry, I am jumping in after missing most of the conversation, but this seems like a way to make matched images despite the whole "PSF is emergent in PhoSim" feature.

Yes, that is an interesting thing we can try.  Just to be clear: here I was just answering David's narrower question of what has been done so far, not commenting on whether is was correct :) 

So, we may also get closer by just choosing the right GalSim PSF (or using them properly).

jchiang87:
I set the `seeing` used by imsim.py to the value from the instance catalog.  That is also the `seeing` used by phosim in its atmosphere model.  The input value and the emergent one from phosim are strongly correlated (based on the measured value from the Stack), but the correlation isn't perfect.

cwwalter:
But how is the seeing actually used in the galsim interface?  It doesn't show up in:

https://github.com/lsst/sims_GalSimInterface/blob/56d725e0a73b4a97d8b3c5581aeb95ea19feeaab/python/lsst/sims/GalSimInterface/galSimPSF.py

is it applied later on top of it somehow?

jchiang87:
Based on the `SNRdocumentPSF` constructor, it looks like at [this line](https://github.com/DarkEnergyScienceCollaboration/imSim/blob/issue/3/set_photometric_parameters/python/desc/imsim/monkeyPatchedGalSimBase.py#L103) we should be doing
```
    self.PSF = SNRdocumentPSF(self.obs_metadata.seeing[self.bandpass])
```
instead.

cwwalter:
Ah, I see.  But it looks like the default is 0.6".  So the value should already be a little better than average reality.

cwwalter:
@dkirkby 

> The ratio of the galsim / phosim source flux is 0.78 (so in the opposite direction to the sky discrepancy)

Looking at your plot it looks like it is more like 20%... Is there also a shift in the other direction so that we are not really seeing the peak of the galsim source in this plot?

drphilmarshall:
It also looks as though the input galaxy ellipticity being interpreted
differently between GalSim and PhoSim: perhaps different conventions are
being followed?

cwwalter:
@drphilmarshall: @danielsf Already mentioned there is a bug which causes the rotation angle to be incorrect which will be fixed.

dkirkby:
@cwwalter I measured the source flux ratio of ~0.78 by integrating over this source (in 2D) and subtracting the mean sky, which should give an answer that is independent of the PSF and ellipticity.

I would expect some discrepancy in the PSF (although not this much), but the discrepancies in sky level and integrated source flux are pointing to inconsistent simulation inputs that we should be able to track down.

If the ~15 pixel shift is due to different treatments of atmospheric refraction, that implies that the sensor +y axis is aligned with the altitude axis, with zero (or 90deg?) rotator angle.  We should be able to confirm that refraction is responsible by rotating the sensor by 45deg and/or observing at a different elevation angle (what was the elevation angle for this simulated observation?)  Note that 15pix is only 3", so closer to the scale of differential chromatic refraction rather than total refraction (assuming this is not observed at ~zenith).

jchiang87:
@dkirkby I recalled last night that we set the nominal sky brightness by hand via
https://github.com/DarkEnergyScienceCollaboration/Twinkles/blob/master/workflows/PhoSim/twinkles_I_physics_override.txt#L11
since phosim 3.4.2 had some issues with the sky backgrounds which have been fixed in 3.5.x (I believe).  So that may be part of the discrepancy in the sky levels.   Unfortunately, we still used that `zenith_v` value in our more recent runs with 3.5.3, so we can't make a meaningful comparison yet.   The differences in the source fluxes are still an issue of course.

Also, there is a bug in the instance catalogs vis-a-vis the phosim 3.4.2 command parser such that the altitude in the instance catalog (70.8 deg in this case) was ignored and the phosim default value of 89 deg was used instead.   This problem is also fixed in 3.5.3 and addresses in the current catsim code.

cwwalter:
So that @danielsf  won't have to retype the same stuff again let me paste in part of an email he sent me about the refraction issue:

> RA, Dec in a PhoSim InstanceCatalog is not the traditional definition of RA, Dec in the International Celestial Reference System (ICRS).  It is ICRS RA, Dec with precession, nutation, aberration but not refraction added in.  I will refer to these coordinates as 'unrefracted RA/Dec'.  Unfortunately, the contents of the PhoSim InstanceCatalog are still more complicated than just that.

> When PhoSim creates a FITS image it reports the WCS as being in ICRS, even though PhoSim itself has no information about how to go from this unrefracted RA, Dec back to ICRS.  We therefore ended up having to compute the unrefracted RA, Dec of the bore site and all objects to get their relative positions correct, and then undo precession, nutation, etc. so that the PhoSim header's rightascension and declination arguments are, in fact the ICRS values (so that the output WCS is correct), and the objects are all reported in the same correct relative position to the bore site.

> So, to summarize, in a PhoSim InstanceCatalog:

> rightascension and declination of the pointing as reported in the header are actually ICRS values

> RA and Dec of objects are not ICRS values.  They are not even unrefracted RA and Dec values.  They are the RA, Dec values that have the same relative position to the pointing RA, Dec as they would if everything (pointing and objects) were in unrefracted RA, Dec.

> This was how we got the LSST stack to do the correct astrometric fits on PhoSim images.

> There do not currently exist methods to go from these coordinates to camera position (the RA, Dec to pixel position methods all take true ICRS coordinates).  I can probably add something to go form 'raPhoSim', 'decPhoSim' to camera coordinates.

dkirkby:
If I understood @danielsf 's description, the phosim (ra,dec) are defined so that photon direction vectors at the top of the atmosphere (-refraction) at the time of the observation (-precession,etc) can be calculated with simple trig, which makes sense (although I would prefer to have any non-standard coord systems hidden within the implementation and not exposed in input files).  We use the [astropy.coordinates](http://docs.astropy.org/en/stable/coordinates/) package for these transformations in the [DESI sims](http://specsim.readthedocs.io/en/latest/guide.html).  What code is used for the PhoSim input files?

dkirkby:
Do we have magnitudes for the sources in these images so we can compare absolute zero points and dig more into the source flux discrepancy?

jchiang87:
I'll look into running `processEimage.py` on the imsim.py file.  This may take a while to sort out since the FITS header will likely need modifications.

danielsf:
Note: the end result of all of the RA, Dec shenanigans should be that, in the actual image, GalSim and PhoSim put objects on the same pixel.  If we are comparing the locations of sources on actual images (which it sounds like we are) and there is still a 15 pixel shift, that would constitute an unexpected problem.

danielsf:
The coordinate transformations in CatSim are all ultimately based on PALPY.  They are wrapped by methods that are a part of the LSST Simulations stack.

dkirkby:
So, to summarize:
- galsim takes ICRS input coords, so is correct by definition (given the output WCS).
- phosim uses non-standard input coords, but should place photons in the same pixel as galsim after refracting them through the atmosphere.

If this is correct (?) then I would guess that the 3" offset is due to a small difference in how phosim refracts and the calculation used to back out the refraction to calculate phosim input coords.

Of course, there might be a simpler explanation for the offset that has nothing to do with refraction, so it would be worth doing a comparison with a different altitude and or rotator angle to rule this out.

cwwalter:
> Do we have magnitudes for the sources in these images so we can compare absolute zero points and dig more into the source flux discrepancy?

Yes.  The input file follows the source format defined here:

https://bitbucket.org/phosim/phosim_release/wiki/Instance%20Catalog

Under "Astrophysical inputs".  So we would need Jim to upload the instance file used for this exposure.  We might also consider writing out the source data frame that comes from the parsing.  This would give a convenient way to look at things.

sethdigel:
The rotator angle (rotSkyPos) was 256.4 deg for this simulation, so I'd guess that a shift due to refraction would be unlikely to be aligned with the CCD orientations.

jchiang87:
I've uploaded the instance catalog to NERSC:
```
[edison11] pwd
/global/homes/j/jchiang8/imSim
[edison11] ls -l phosim_input_200.txt 
-rw-r--r-- 1 jchiang8 jchiang8 64032541 Oct 19 09:57 phosim_input_200.txt
[edison11] 
```

cwwalter:
For doing the sort of test David is talking about it probably would be good to make a very simple catalog with 10 sources and fixed magnitudes or something.

jchiang87:
yes. and place them strategically, e.g., one at the nominal pointing direction.

dkirkby:
Could they spell out "LSST" so we can have our [IBM moment](https://en.wikipedia.org/wiki/IBM_(atoms)).

dkirkby:
The phosim input catalog does not provide the relevant magnitudes directly.  Instead, it specifies the SED, its normalization at 500nm, and how to apply host extinction.  I know there is lsst.sims code to do the necessary calculations, but it might be easier to save the magnitudes (or fluxes) at the point where they are now being calculated.

dkirkby:
@sethdigel I guess rotSkyPos = 256.4deg rules out a simple refraction explanation for the 15pix shift, assuming we are confident that the simulation actually used this rotation angle.

jchiang87:
> it might be easier to save the magnitudes (or fluxes) at the point where they are now being calculated.

It could also save on memory.  I've only scanned the sims code, but it appears to be caching each of the SEDs in case they get reused.  My guess is that is the bulk of the memory usage.  Saving the fluxes would be more compact. @danielsf Am I guessing right?

dkirkby:
If SED caching is expensive, perhaps it could be eliminated by iterating over sources grouped by SED?

danielsf:
@jchiang87 Yes: the SEDs get cached in case they are used again.

In the case of ImSim, though, the SEDs are cached before dust extinction is applied, so saving the magnitudes/fluxes won't really help us, since the dust extinction is applied as a function of wavelength across the whole SED.

dkirkby:
Is there a galsim equivalent of this phosim input catalog, or is all the translation done on the fly?  What coordinate system is used for the source centroids passed to galsim?

danielsf:
@dkirkby That will require more careful thought, since it will break the existing `sims_GalSimInterface` code, which is designed to connect directly to fatboy.  The fatboy queries are not designed to group sources by SED.

dkirkby:
@danielsf Wouldn't it just require an additional `GROUP BY` sql clause?  (but this only matters if the caching is expensive).

danielsf:
GalSim centroids are handled directly in focal plane coordinates.  CatSim has methods to go from RA, Dec to position on the focal plane.

It is true that the CatSim refraction model is not very well-developed.  We just take the refraction at 500nm and assume that is good enough.  We should probably calculate a different effective wavelength for each SED.  That, of course, would not lead to a uniform 15 pixel shift.

danielsf:
I don't know how expensive doing that extra Group By would be.

cwwalter:
> Is there a galsim equivalent of this phosim input catalog, or is all the translation done on the fly? What coordinate system is used for the source centroids passed to galsim?

It is done on the fly.  The parser reads in the PhoSim input file and builds a binary data frame with all of the source info.  These sources are then passed off to GalSim as Scott explains above.

dkirkby:
@cwwalter It might be worth writing out an equivalent galsim catalog for quicker and more direct comparisons, since some of the discrepancies are likely already introduced at that stage.

dkirkby:
@danielsf The atmospheric differential refraction over the visible bands is on the order of 1 arcsec (5 pix), so that definitely needs revisiting.  The plot below (from [this paper](https://arxiv.org/abs/1506.04790)) shows ADR vs wavelength and altitude:

adr

sethdigel:
> @sethdigel I guess rotSkyPos = 256.4deg rules out a simple refraction explanation for the 15pix shift, assuming we are confident that the simulation actually used this rotation angle.

In its output for this run, phosim reports having used this angle, and based on matching star locations in phosim output with instance catalogs for other runs, I am confident that phosim is actually making this rotation in our simulations.

python/desc/imsim/imSim.py
```diff
@@ -178,3 +202,32 @@ def extract_objects(df):
     phosim_galaxies = \
         phosim_galaxies.query("halfLightSemiMajor >= halfLightSemiMinor")
```
cwwalter: "halfLightSemiMajor >= halfLightSemiMinor"

Was an error handling case.  There was only one of these in the 80000 item test catalog that Scott made.  So really, it should be fixed in the production program.

But, for here, it doesn't feel right to me to just put it into the query like this.  There might be other mistakes that we want to check for too.  So, if we want to include the selection here maybe it would be better to put these in a string called "validGalaxyCondition" or something.

bin/imsim.py
```diff
@@ -98,24 +100,24 @@ def main():
...
+        # Trim the input catalog to a single chip.
+        raICRS = phoSimObjectList['ra'].values
+        decICRS = phoSimObjectList['dec'].values
+        phoSimObjectList['chipName'] = chipNameFromRaDec(raICRS, decICRS,
+                                                         camera=camera,
+                                                         obs_metadata=obs,
+                                                         epoch=2000.0)
         starDataBase = \
-            phoSimObjectList.query("galSimType == 'pointSource' and magNorm < 50 and chipName=='%s'" % arguments.sensor)
+            phoSimObjectList.query("galSimType=='pointSource' and magNorm<50 and chipName=='%s'" % arguments.sensor)
```
dkirkby: I wouldn't expect `magNorm < 50` objects to cause a crash or to have a noticeable impact on the timing.

cwwalter: Right... but Jim originally added these to try to get rid of a crash (which came from a mean 0 counts for Poisson distribution).  I'm not sure if this was really the reason it happened and I was asking whether that was why it was there now or if there was something else.

jchiang87: It is definitely needed for the Twinkles instance catalogs.  Those can have tens of sources with magnorms ranging from 50 to 1000.  I didn't have the patience to find out exactly at which value the mean counts was zero, so I just picked 50.  Unless there is a practical reason for not having this cut, I don't see why we shouldn't retain it.

python/desc/imsim/imSim.py
jchiang87: I included it to be faithful to your code, but in fact, it isn't violated for the instance catalogs I checked, including the example from Scott.  I'm perfectly happy to remove it entirely.

jchiang87: That's right, it was on linux where it didn't crash.  Using the old code, I did run into that message complaining about a single galaxy at the 1e-22 level, but with the new code, `query("halfLightSemiMajor >= halfLightSemiMinor")` has no effect, i.e., all rows are passed and the code runs fine.  So we can add a check, but it wouldn't do anything on linux.   What other validations would you suggest?

bin/imsim.py
jchiang87: We aren't doing infinite precision math. If we pick a magnitude value that's large enough, it will underflow to zero eventually.   That point probably is closer to 1000 than it is to 50, logarithmically speaking, but since our zero points seem to be around ~30-33, I figured 50 would be fine.

sethdigel: In its output summary, phosim lists the numbers of sources and the numbers of photons they generated by magnitude range.  For the Twinkles runs, no source with magnitude > 37 generates any recorded photons, and only ~10 sources that faint or fainter are in the instance catalogs (see, e.g., [this log file](http://srs.slac.stanford.edu/Pipeline-II/exp/LSST-DESC/log.jsp?pi=41888463) for ObsHistID 200.

cwwalter: OK thanks Jim and Seth.  I understand now.

python/desc/imsim/imSim.py
jchiang87: That's fair enough.  I'll add a function to make these checks and report any issues if they occur.

jchiang87:
@cwwalter I updated the code along the lines we discussed.  If it all looks good to you, please go ahead and merge.  If you cd to the `tests` subdir, you can see object validation in action on the tiny instance catalog there:
```
$ imsim.py tiny_instcat.txt 
Reading all rows from the instance catalog tiny_instcat.txt.
/nfs/farm/g/desc/u1/users/jchiang/desc_projects/imSim/python/desc/imsim/imSim.py:230: UserWarning: Omitted 2 suspicious objects from the instance catalog satisfying:
1  (galSimType=="sersic" and halfLightSemiMajor < halfLightSemiMinor)
1  (magNorm > 50)

  warnings.warn(message)
11572 [0x7fc3109f4700] DEBUG daf.persistence.LogicalLocation null - Input string: .
11572 [0x7fc3109f4700] DEBUG daf.persistence.LogicalLocation null - Copy to: .
11578 [0x7fc3109f4700] INFO CameraMapper null - Unable to locate registry registry in root: ./registry.sqlite3
11579 [0x7fc3109f4700] INFO CameraMapper null - Unable to locate registry registry in current dir: ./registry.sqlite3
11580 [0x7fc3109f4700] INFO CameraMapper null - Loading Posix registry from .
WARNING: ErfaWarning: ERFA function "taiutc" yielded 1 of "dubious year (Note 4)" [astropy._erfa.core]
ErfaWarning: ERFA function "taiutc" yielded 1 of "dubious year (Note 4)"
17499 [0x7fc3109f4700] DEBUG afw.image.Mask null - Number of mask planes: 16
20302 [0x7fc3109f4700] DEBUG afw.image.Mask null - Number of mask planes: 16
$ 
```

This PR also fixes #13 .

cwwalter:
This all looks good.  I can merge it.  Just two quick questions first:

- It might be good to list the object ID for the rejected  cases so you can go and look at them.

- What is causing the daf and afw warnings?

cwwalter:
BTW for everyone... if there are any outstanding issues in this PR, make a new issue and refer back to it since this is about to close. I already made one for the object shift issue.

cwwalter:
In case it wasn't clear the reason I asked about the daf and AFW warnings was that I never saw those in my original version. I was curious if it was a  new Linux effect or if something new was being called I missed.

jchiang87:
I thought about listing the object ids but I was worried about the cases where there were a lot of rejected rows.  The validator function can be run with the parser separately to get a df with the rejects.  I don't know what's with the daf and afw related messages.  They're debug lines, so I would not expect them to be printed by default.

cwwalter:
> I thought about listing the object ids but I was worried about the cases where there were a lot of rejected rows. The validator function can be run with the parser separately to get a df with the rejects. 

OK.

>I don't know what's with the daf and afw related messages. They're debug lines, so I would not expect them to be printed by default.

OK, I was just surprised since the only message I saw before was about the registry. I assume it is still related to that and the camera.  Maybe you are just running a newer version of the stack than me (I am still on V12).  I will go ahead and merge.

PR #15, opened 2016-10-23
fixes for eimage files

jchiang87:
Two items:
* In order for `ingestSimImages.py` (from `obs_lsstSim`) to process the eimages, the `obshistID` needs to be set in the header (see [here](https://github.com/lsst/sims_GalSimInterface/blob/56d725e0a73b4a97d8b3c5581aeb95ea19feeaab/python/lsst/sims/GalSimInterface/galSimDetector.py#L540) and [here](https://github.com/lsst/sims_GalSimInterface/blob/56d725e0a73b4a97d8b3c5581aeb95ea19feeaab/tests/testFitsHeaders.py#L79)) and the eimage filename has to start with `lsst_e` (see [here](https://github.com/lsst/obs_lsstSim/blob/1da903b9cd7e96231fe809d684ba3b73af1cb495/python/lsst/obs/lsstSim/ingest.py#L87)).
* The exposure time should be the visit time divided by the number of snaps.  The `sims_GalSimInterface` code adds the snaps together to produce a single eimage.

### DIFF
--- a/bin/imsim.py
+++ b/bin/imsim.py
@@ -103,6 +103,9 @@ def main():
                               m5=defaults.m5(bandpass),
                               seeing=seeing)
 
+    # Set the OpsimMetaData attribute with the obshistID info.
+    obs.OpsimMetaData = {'obshistID': visitID}
+
     # Now further sub-divide the source dataframe into stars and galaxies.
     if arguments.sensor is not None:
         # Trim the input catalog to a single chip.
@@ -148,7 +151,7 @@ def main():
     outdir = arguments.outdir
     if not os.path.isdir(outdir):
         os.makedirs(outdir)
-    phoSimGalaxyCatalog.write_images(nameRoot=os.path.join(outdir, 'e-image_') +
+    phoSimGalaxyCatalog.write_images(nameRoot=os.path.join(outdir, 'lsst_e_') +
                                      str(visitID))
 
 if __name__ == "__main__":
--- a/python/desc/imsim/imSim.py
+++ b/python/desc/imsim/imSim.py
@@ -253,7 +253,8 @@ def photometricParameters(phosim_commands):
     The effects from all three of those will be added by the
     electronics chain readout code.
     """
-    return PhotometricParameters(exptime=phosim_commands['vistime'],
+    exptime = phosim_commands['vistime']/float(phosim_commands['nsnap'])
+    return PhotometricParameters(exptime=exptime,
                                  nexp=phosim_commands['nsnap'],
                                  gain=1,
                                  readnoise=0,
--- a/python/desc/imsim/imSim.py
+++ b/python/desc/imsim/imSim.py
@@ -1,7 +1,7 @@
 """
 Base module for the imSim package.
 """
-from __future__ import absolute_import, print_function
+from __future__ import absolute_import, print_function, division
 import warnings
 from collections import namedtuple
 import numpy as np


cwwalter:
This looks OK.

> The exposure time should be the visit time divided by the number of snaps. The sims_GalSimInterface code adds the snaps together to produce a single eimage.

What does PhoSim do?  Do you get a sequence of snaps? I'm a bit confused because somethings seem to have changed.  For example I have things like

SIM_VISTIME 15.0
SIM_NSNAP 1

in instance files but the new documentation doesn't have the "SIM_" anymore.

jchiang87:
phosim writes each snap to a separate file.  I'm a little uncertain how phosim handles `vistime` vs `exptime`. I think effectively it does `vistime = 2*exptime + readout_time` where `readout_time` is something like 3 seconds.  How these various values get set consistently is unclear to me.

danielsf:
from an email exchange I had with John Peterson

----
the vistime = nsnap * (exposure time) + (nsnap -1) * (readouttime + shutter delay)

right now in v3.5 the readout time + shutter delay is 3.0 seconds

so if you want 15 seconds exposures you’d specify 33.0 seconds.
----

hope that helps

cwwalter:
> so if you want 15 seconds exposures you’d specify 33.0 seconds.

Just to be extra clear: This means "if you want two 15 second exposures" correct?

danielsf:
Yes (though you also have to set the NSNAPS parameter equal to 2, I think)

SimonKrughoff:
FWIW, it would be fine with me to just go with a single 30 second image rather than two 15 sec exposures.  We don't currently have a snap difference CR rejection algorithm, so the separate snap processing just adds more overhead.

dkirkby:
(Remember to scale up the readout noise if you do this)

cwwalter:
> FWIW, it would be fine with me to just go with a single 30 second image rather than two 15 sec exposures. We don't currently have a snap difference CR rejection algorithm, so the separate snap processing just adds more overhead.
> 
> I believe this is what we did for previous runs anyway.

Is this what is effectively happening now anyway since as @jchiang87 says "The exposure time should be the visit time divided by the number of snaps. The sims_GalSimInterface code adds the snaps together to produce a single eimage"?

jchiang87:
For nsnap=1 and vistime=30, the answer is yes.  But when one runs the catsim code to generate the instance catalogs, those values have to be communicated to catsim.  For Twinkles, they are actually written into the instance catalog generation code.

To get `imSim` to be consistent with what phosim is doing, we need to implement the equation that Scott quotes from John, `vistime = nsnap * (exposure time) + (nsnap -1) * (readouttime + shutter delay)`, and as David notes, make sure the read noise scales with the number of snaps.

cwwalter:
> For nsnap=1 and vistime=30, the answer is yes. But when one runs the catsim code to generate the instance catalogs, those values have to be communicated to catsim. For Twinkles, they are actually written into the instance catalog generation code.
> 
> To get imSim to be consistent with what phosim is doing, we need to implement the equation that Scott quotes from John, vistime = nsnap * (exposure time) + (nsnap -1) * (readouttime + shutter delay), and as David notes, make sure the read noise scales with the number of snaps.

So for right now we will go with one 30 second snap correct?  

Can someone enter adding the more "proper" general behavior into an issue?

Also, if we set the 30 second snap up properly do we need to increasing the read noise as David pointed out? Or will that already be handled properly but the lsst-galsim interface?

jchiang87:
I'll post the issue.  Yes, I think we've agreed that a single 30 sec snap is the way to go.  Read noise will be handled by the electronics readout code.  I've added my code for that in a [branch](https://github.com/DarkEnergyScienceCollaboration/imSim/tree/issue/6/realistic_electronics_chain). I'm hoping to make a PR in the next few days.

Issue #21, opened 2016-10-25
Computing Infrastructure enhancements


jchiang87:
- [x] Enable the package to build and run on [OS X El Capitan](http://dmtn-001.lsst.io/en/master/). #18 
- [x] Connect repo to CI systems. #20
- [x] Add command/configuration file capability. #11

Issue #11, opened 2016-10-19
Add command file capability


cwwalter:
Add the ability to read a command file which will set non-visit specific configuration options.

We need to think both about what options go in it and the format/configuration language.

jchiang87:
closed by #24

PR #24, opened 2016-10-27
Issue/21/computing infrastructure

jchiang87:
This isn't quite ready yet, but I wanted to open the PR so that the building on El Capitan could be tested by someone with that platform (it works on linux).

I'd also like to add the command/configuration file work (#11) to this PR.  I'll update the epic (#21) accordingly.


cwwalter:
What actually needs gcc here?

jchiang87:
`scons` does a bunch of configuration stuff, including checking that gcc 4.8 is available.  This is part of the cost of doing the `bin.src` build.

cwwalter:
> I'd also like to add the command/configuration file work (#11) to this PR. I'll update the epic (#21) accordingly.

Is there a reason not to do this in separate PRs?

cwwalter:
> This isn't quite ready yet, but I wanted to open the PR so that the building on El Capitan could be tested by someone with that platform (it works on linux).

OK, I will check (on Sierra).

jchiang87:
Dealing with #11 allows me to fix the unit test failure "correctly" by putting that readout time in the putative configuration file, so that this PR actually passes the CI builds (or at least the Jenkins build).   If you feel like it should be in a separate PR, that's fine.  We'd just have to merge it before this one.

cwwalter:
> This isn't quite ready yet, but I wanted to open the PR so that the building on El Capitan could be tested by someone with that platform (it works on linux).

I checked this on Sierra.  I typed 'scans' it passed the checks, did the build and wrote a imsim.py file into bin/ with the proper shebang line.

cwwalter:
> Dealing with #11 allows me to fix the unit test failure "correctly" by putting that readout time in the putative configuration file, so that this PR actually passes the CI builds (or at least the Jenkins build). If you feel like it should be in a separate PR, that's fine. We'd just have to merge it before this one.

Ah, I see.  OK, I don't have a strong feeling.  It's a bit nicer for tracking to have them separate but do what ever is easiest for you.

cwwalter:
I can't comment in an informed fashion on the travis.yml file but all of the other changes here look fine.

cwwalter:
The usage line in the README is no longer correct.  -f is no longer a optional argument.

README.md
```diff
@@ -1,16 +1,25 @@
...
+execute the `setup imsim` command, and that command can be issued from
+any directory.  The `scons` build step needs only to be re-run if a
+new command line executable is added to the `bin.src` folder.
 
 ## Usage
+The executables in the `bin` folder should be in your path and so
+should be runnable directly from the command line:
 ```
 $ imsim.py --help
 usage: imsim.py [-h] [-v] [-f FILE] [-n NUMROWS] [--outdir OUTDIR]
```
jchiang87: Chris comments that https://github.com/DarkEnergyScienceCollaboration/imSim/pull/24#issuecomment-256755146

jchiang87:
this is ready for review.

cwwalter:
This code all looks fine to me.  I do have two questions on the config file part.

- Currently there is no way to set the file, it just always reads the default in data.  This is fine, I just wanted to mention it on the off chance other code hasn't been checked in yet or something.

- Right now the read_config() command is called inside of photometricParameters which is called from imsim.py.

Are there likely to be config options (for the sensor, telescope, sky etc) that won't be photometric parameters?  Or will every relevant thing we want to do or pass to GalSim really always be put in a *Catalog.photParams class?

cwwalter:
One other thing to think about for the config file...

We will have different sorts of inputs: PhoSim instance, CatSim driver?, CCD Sensor driver etc.  

I imagine all of these may have somewhat different commands. I imagined that the command type itself would be in the file.  In this case the parsing likely needs to happen at a higher level.  On the other hand, this could be driven by a command line option and then each option could call read_option itself.

jchiang87:
I've added the option to override the default config at the command line for `imsim.py` and the code now reads the config just once in the main executable.   Once we have more concrete use cases, we can add the desired functionality.  If everything looks ok, I'd like to merge this PR since I have other work which would benefit from having this in master.

PR #30, opened 2016-11-05
subclass derived classes of GalSimBase instead of replacing base class functions

jchiang87:
Here's a proposal to use subclassing instead of using the `monkeyPatchedGalSimBase.py` code.  It's all on the client-side, i.e., it doesn't require any sims code to be modified, which is probably a good thing.

I've only exercised it on `tests/tiny_instcat.txt`, so I need to do more testing as well as add real unit test code.  Opening this PR so people can comment on the approach.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -87,3 +87,6 @@ ENV/
 
 # Rope project settings
 .ropeproject
+
+# backup files
+*~
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,7 @@ install:
   - ln -sf /home/travis/miniconda/lib/libcrypto.so.1.0.0 /home/travis/miniconda/lib/libcrypto.so.10
   - export PATH="$HOME/miniconda/bin:$PATH"
   - source eups-setups.sh
+  - conda install --es astropy=1.1.2
   - pip install coveralls
   - setup lsst_sims
   - eups declare -r . imsim -t current
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,7 @@ install:
   - ln -sf /home/travis/miniconda/lib/libcrypto.so.1.0.0 /home/travis/miniconda/lib/libcrypto.so.10
   - export PATH="$HOME/miniconda/bin:$PATH"
   - source eups-setups.sh
-  - conda install --es astropy=1.1.2
+  - conda install --yes astropy=1.1.2
   - pip install coveralls
   - setup lsst_sims
   - eups declare -r . imsim -t current
--- a/bin.src/imsim.py
+++ b/bin.src/imsim.py
@@ -8,16 +8,9 @@ code found in sims_GalSimInterface.
 from __future__ import absolute_import, print_function
 import os
 import argparse
-
 from lsst.obs.lsstSim import LsstSimMapper
-
-from lsst.sims.photUtils import LSSTdefaults
-from lsst.sims.utils import ObservationMetaData
 from lsst.sims.coordUtils import chipNameFromRaDec
-
-from lsst.sims.GalSimInterface import GalSimStars
-from lsst.sims.GalSimInterface import GalSimGalaxies
-
+from lsst.sims.GalSimInterface import GalSimStars, GalSimGalaxies
 import desc.imsim
 
 def main():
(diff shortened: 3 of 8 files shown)


cwwalter:
Interesting... OK, I didn't know enough about factory classes to think of doing this.

I'm not against doing this but just to be clear I considered the monkey patching to be a temporary solution that would go away when the interface classes were refactored.  I still think it makes sense to do that.  Depending on how we use them maybe they would actually even be housed together. I think the input interface and galsim interface parts should be separated so the code could be used somewhat as I did after the patching.

So, if this is all going to go away, I wonder if it makes a lot of sense to spend effort designing other ways to do it. OTOH, I could imagine this might be somewhat safer if it sticks around for a while since it is probably a bit more resilient to changes in the base classes.   Alternatively, the galsim interface could be redesigned with this kind of extension actually in mind.

What do you think of these points Jim?

cwwalter:
Also @danielsf should comment.

jchiang87:
If I were doing this from scratch, I would have either 
* subclassed CatalogDBObject to serve up the data extracted from the instance catalogs (perhaps using pandas underneath), or figure out how to stuff that data into an existing CatalogDBObject subclass.  In this case, the existing GalSimBase subclasses could simply be used as is.
* subclassed the GalSimBase subclasses as I've done here.  I've used some metaclass methods to avoid code duplication, but ultimately what I've done isn't different from subclassing each GalSimBase subclass explicitly.  Fundamentally, the only thing that's needed is overloading .column_by_name, so the subclasses themselves would be/are simple extensions of the GalSimBase versions.

In general, I'd say that removing the code that was copied from sims_GalSimInterface is a desirable thing to do.  So even if this new code is ultimately made unnecessary by a CatalogDBObject solution (my preferred solution, but it's not clear to me how to do this), or because the sims code itself is otherwise modified, I think these changes are worth adopting in the near term.  It took about a day of my effort, and I learned a lot about the sims code as a result.  I don't consider that wasted effort.

> I think the input interface and galsim interface parts should be separated so the code could be used somewhat as I did after the patching.

I don't understand this statement.

cwwalter:
>> I think the input interface and galsim interface parts should be separated so the code could be used somewhat as I did after the patching.

> I don't understand this statement.

Oh, I just meant what you wrote above.  @danielsf 's original code was designed from the start to interface with the CatSim database and carry its own database for the columns.  Scott and I actually talked about having pandas pull into the internal database instead but we decided that this was specific enough to the CatSim application that it was better to bypass it.

I just mean that because of the current design the GalSim interface code is doing two things: getting information out of the CatSim database (which we are replacing with the code here), and driving calls to GalSim. I was just saying that I think those two things need to be separate.  The code to drive GalSim should be agnostic to where the information to drive it comes from.

> In general, I'd say that removing the code that was copied from sims_GalSimInterface is a desirable thing to do.

OK, if you think it is best then let's change it.  Probably best to do it ASAP since, (for example) I have a branch related to Peter's sky model that modifies what is there now.

jchiang87:
ok. I don't have any more changes for this branch, so I agree that it should be put into master so other work can proceed.  We can always change things later if necessary.

cwwalter:
OK, please merge it whenever you think it is ready.

cwwalter:
So this results in this for me if I try to run on the tiny_instcat.txt file:

```
  File "/Users/walter/LSST/imSim/bin/imsim.py", line 158, in <module>
    main()
  File "/Users/walter/LSST/imSim/bin/imsim.py", line 141, in main
    phoSimStarCatalog.get_fitsFiles()
  File "/Users/walter/LSST/imSim/python/desc/imsim/monkeyPatchedGalSimBase.py", line 155, in get_phoSimInstanceCatalog
    objectNames = self.phoSimDataBase['objectID'].values
  File "/Users/Walter/LSST/Anaconda/lib/python2.7/site-packages/pandas/core/frame.py", line 1997, in __getitem__
    return self._getitem_column(key)
  File "/Users/Walter/LSST/Anaconda/lib/python2.7/site-packages/pandas/core/frame.py", line 2004, in _getitem_column
    return self._get_item_cache(key)
  File "/Users/Walter/LSST/Anaconda/lib/python2.7/site-packages/pandas/core/generic.py", line 1350, in _get_item_cache
    values = self._data.get(item)
  File "/Users/Walter/LSST/Anaconda/lib/python2.7/site-packages/pandas/core/internals.py", line 3290, in get
    loc = self.items.get_loc(item)
  File "/Users/Walter/LSST/Anaconda/lib/python2.7/site-packages/pandas/indexes/base.py", line 1947, in get_loc
    return self._engine.get_loc(self._maybe_cast_indexer(key))
  File "pandas/index.pyx", line 137, in pandas.index.IndexEngine.get_loc (pandas/index.c:4154)
  File "pandas/index.pyx", line 159, in pandas.index.IndexEngine.get_loc (pandas/index.c:4018)
  File "pandas/hashtable.pyx", line 675, in pandas.hashtable.PyObjectHashTable.get_item (pandas/hashtable.c:12368)
  File "pandas/hashtable.pyx", line 683, in pandas.hashtable.PyObjectHashTable.get_item (pandas/hashtable.c:12322)
KeyError: 'objectID'
```

jchiang87:
I think you have the old `imsim.py` hanging around.  Try
```
scons -c
scons
```
that will do a clean and rebuild.

cwwalter:
ah.. forgot about the bin.src thing.  Yes, that fixed it.  Thanks.

Issue #23, opened 2016-10-26
Allow or ignore missing phosim instance file commands


cwwalter:
Some valid phosim instance file commands (like rotskytel) are not in the list of valid commands.

imSim should either use these, or more gracefully ignore them.  The current behavior results in a exception.

Also allowing commenting out lines with a '#' should be supported.

jchiang87:
I think "gracefully" ignoring things like this is what caused problems with the Run1.1 phosim output wrt the `altitude` keyword.  IMO, having things fail loudly with an exception helps make certain that the input parameters and required parameters are the same.  I think the real issue is deciding what parameters are *required* and which are *optional*.  The current test in the code imposes a 1-1 onto mapping, i.e., all the ones listed are required and no optional ones allowed.

cwwalter:
OK fair enough.  It means we can't run with certain files but you are saying that all twinkles file would be missing this keyword?

Commenting out is still important no matter what.

jchiang87:
I agree on commenting out.  I had to add rotskytel by hand to use imsim.  not sure why it wasn't in those files.

danielsf:
What is rotskytel?  I know about rotskypos and rottelpos.  I have never heard of rotskytel.

cwwalter:
I mistyped .  The one that caused the problem was rotskypos.  But they both do exist.

Look here:

https://bitbucket.org/phosim/phosim_release/wiki/Instance%20Catalog

under Telescope Operator.

cwwalter:
Forget it... I was just reading too fast.

That one doesn't exist. Just a typo.

jchiang87:
rottelpos was the one I had an issue with.  and this was using old instance catalogs from Run1.1.  So the most recent ones from catsim may have rottelpos.

danielsf:
I agree with Jim's statement.  If new InstanceCatalogs don't have rottelpos, let me know, because they should

jchiang87:
This is mostly handled by #35.   Possibly left to do:

- [ ] Make list of required phosim commands configurable rather than hard-wired.

danielsf:
I think this is done.  `parsePhoSimInstanceFile` reads all of the header entries in the InstanceCatalog and just emits a warning if any optional parameters are present.

PR #35, opened 2016-11-21
allow comments and optional commands

jchiang87:
This is basically done.   One additional change to consider would be to put the list of expected phosim commands in `data/default_imsim_configs` or perhaps some other configuration file that could be overridden at runtime instead of [hard-wiring them](https://github.com/DarkEnergyScienceCollaboration/imSim/blob/master/python/desc/imsim/imSim.py#L33).

### DIFF
--- a/python/desc/imsim/imSim.py
+++ b/python/desc/imsim/imSim.py
@@ -29,7 +29,7 @@ class PhosimInstanceCatalogParseError(RuntimeError):
 PhoSimInstanceCatalogContents = namedtuple('PhoSimInstanceCatalogContents',
                                            ('commands', 'objects'))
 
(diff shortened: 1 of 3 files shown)


cwwalter:
Nice. Looks good. 

I'll merge.  Do you want to leave #23 open, or add a new issue for the possibility of not hardwiring the list?  I think hardwiring isn't so bad for now since I think the phosim instance file will be a fixed compatibility input for this program eventually.

jchiang87:
Thanks, Chris.  We might as well leave #23 open for now.

cwwalter:
OK... sounds good.  Can you put a note in there about what is left to do?

PR #36, opened 2016-11-23
Issue/31/instance catalog extinction handling

jchiang87:
This is ready to go.  In addition to the unit tests, I ran on a Twinkles Run3 instance catalog (one that has the correct ordering of the internal and Galactic extinction columns) and it ran fine and compares as well as before with the corresponding phosim image using the same instance catalog and SEDs.


cwwalter:
OK. Thanks!

PR #38, opened 2016-12-05
Issue/33/remove deprecated stack calls

jchiang87:
Also removed `python/desc/imsim/monkeyPatchedGalSimBase.py`.

This is ready to go.


cwwalter:
OK this looks good.

Issue #5, opened 2016-10-18
Implement realistic sky model


cwwalter:
It would be good to use the work from @yoachim and @dkirkby.

yoachim:
https://github.com/lsst/sims_skybrightness

cwwalter:
Thanks!

Can this be easily used as is to add the proper sky model in our program? 

Also, have you thought about coordinating with @dkirkby to incorporate the material he showed about the sky at the site?

yoachim:
It should be good to go--give it an MJD and sky coordinates and it can return the spectrum of the background sky.  

I'm not currently planning on adding more to the sky model unless there's a strong demand for it.

yoachim:
If you use the setRADecMjd method, it assumes you're at the LSST site and computes moon and sun position for you.  There is also an under-documented setParams method where you can set all the relevant parameters manually if you really want to. There's no sanity checking with setParams, so you can put a full moon at zenith at sunset if you want to.

cwwalter:
@jchiang87 and @sethdigel:  In the phosim files the moon position is set correct? Or is that just an override if you don't want do something like Peter just described?

dkirkby:
The ESO model does not reproduce details of actual BOSS sky spectra, but its not clear how much that matters for LSST simulations.  Here are some plots I showed at the June SSIM meeting:

Typical BOSS sky: (flux vs wavelength in Angstroms)
![boss-sky](https://cloud.githubusercontent.com/assets/185007/19613237/28789be4-97a0-11e6-8bec-453939c52a80.png)

ESO model matched to this BOSS spectrum:
![eso-model](https://cloud.githubusercontent.com/assets/185007/19613257/4a2b1ed8-97a0-11e6-9026-af7d2bb41f03.png)

The most obvious discrepancy is the bump around 6000A, which is also seen at Cerro Paranal (so not due to APO light pollution) and attributed to airglow in [this paper](https://arxiv.org/abs/1205.2003) (see Fig. 16).

cwwalter:
> I'm not currently planning on adding more to the sky model unless there's a strong demand for it.

Based on @dkirkby's work, it sounds like the airglow feature at 6000Å might be relevant/important.

cwwalter:
> It should be good to go--give it an MJD and sky coordinates and it can return the spectrum of the background sky.

So if we want this to vary over the chip, we should call it at every pixel position?  

- Do we want that? Or do we want to set a background per chip?

[...]

PR #40, opened 2016-12-19
Issue/5/realistic sky model

cwwalter:
This pull request contains code to replace the exampleCCDNoise background class from the GalSim interface which uses a M5 level with @yoachim's ESO based sky model.  Peter's model returns the sky level in mag/arcsec^2 and @dkirkby's function skyCountsPerSec converts that into a count level using the exposure time in the instance catalog stored in photParams.exptime.  Note that David has calculated the zero points himself for a set of LSST filters. This should be changed to use the photUtils code.

Right now this just sets the whole skylevel based on the central pointing.  I think as an intermediate step we want to this per chip, and then probably eventually build a healpix table completely outside of the loop so we can assign each pixel.  But, I was thinking it might be useful to get this code in first and try it to confirm it gives reasonable signal/background levels and possibly put some options in for sky background choice.

Note there are also several small changes to meet PEP8 DM guidelines.

### DIFF
--- a/python/desc/imsim/imSim_class_factory.py
+++ b/python/desc/imsim/imSim_class_factory.py
@@ -7,7 +7,7 @@ import gc
 from lsst.sims.utils import pupilCoordsFromRaDec
 
 from lsst.sims.GalSimInterface import GalSimStars, GalSimGalaxies
-from lsst.sims.GalSimInterface import ExampleCCDNoise, SNRdocumentPSF
+from lsst.sims.GalSimInterface import SNRdocumentPSF
 from desc.imsim.skyModel import ESOSkyModel
 
 __all__ = ['ImSimStars', 'ImSimGalaxies']
@@ -57,11 +57,14 @@ def imSim__init__(self, phosim_objects, obs_metadata, catalog_db=None):
     self.db_obj = type('DummyDB', (), dict(epoch=2000))
 
     # Add noise and sky background
-#    self.noise_and_background = ExampleCCDNoise(addNoise=True,
-#                                                addBackground=True)
-
-    # We need to know more than the basic info for Peter's sky model.
-    # Pass obs_metadata, chip etc...
+    # The simple code using the default lsst-GalSim interface would be:
+    #
+    #    self.noise_and_background = ExampleCCDNoise(addNoise=True,
+    #                                                addBackground=True)
+    #
+    # But, we need a more realistic sky model and we need to pass more than
+    # this basic info to use Peter Y's ESO sky model.
+    # We must pass obs_metadata, chip information etc...
     self.noise_and_background = ESOSkyModel(obs_metadata, addNoise=True,
                                             addBackground=True)
 
--- a/python/desc/imsim/imSim_class_factory.py
+++ b/python/desc/imsim/imSim_class_factory.py
@@ -3,9 +3,12 @@ Code to generate imSim subclasses of GalSimBase subclasses.
 """
 from __future__ import absolute_import, print_function, division
 import gc
-from lsst.sims.GalSimInterface import GalSimStars, GalSimGalaxies, \
-    ExampleCCDNoise, SNRdocumentPSF
+
 from lsst.sims.utils import pupilCoordsFromRaDec
+
+from lsst.sims.GalSimInterface import GalSimStars, GalSimGalaxies
+from lsst.sims.GalSimInterface import ExampleCCDNoise, SNRdocumentPSF
+from desc.imsim.skyModel import ESOSkyModel
 
 __all__ = ['ImSimStars', 'ImSimGalaxies']
 
@@ -54,8 +57,13 @@ def imSim__init__(self, phosim_objects, obs_metadata, catalog_db=None):
     self.db_obj = type('DummyDB', (), dict(epoch=2000))
 
     # Add noise and sky background
-    self.noise_and_background = ExampleCCDNoise(addNoise=True,
-                                                addBackground=True)
+#    self.noise_and_background = ExampleCCDNoise(addNoise=True,
+#                                                addBackground=True)
+
+    # We need to know more than the basic info for Peter's sky model.
+    # Pass obs_metadata, chip etc...
+    self.noise_and_background = ESOSkyModel(obs_metadata, addNoise=True,
+                                            addBackground=True)
 
     # Add a PSF.  This one is taken from equation 30 of
     # www.astro.washington.edu/users/ivezic/Astr511/LSST_SNRdoc.pdf .
--- a/python/desc/imsim/imSim_class_factory.py
+++ b/python/desc/imsim/imSim_class_factory.py
@@ -9,6 +9,7 @@ from lsst.sims.utils import pupilCoordsFromRaDec
 
 __all__ = ['ImSimStars', 'ImSimGalaxies']
 
+
 def imSim_class_factory(galsim_subclass):
     """
     Return a subclass of galsim_subclass that takes a pandas DataFrame
@@ -22,6 +23,7 @@ def imSim_class_factory(galsim_subclass):
                              ('__name__', imSim_class_name)]))
     imSim_class.__imSim_class__ = imSim_class
     return imSim_class
+
 
 def imSim__init__(self, phosim_objects, obs_metadata, catalog_db=None):
     """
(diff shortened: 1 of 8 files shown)


python/desc/imsim/imSim_class_factory.py
```diff
@@ -52,8 +57,16 @@ def imSim__init__(self, phosim_objects, obs_metadata, catalog_db=None):
...
-                                                addBackground=True)
+    # The simple code using the default lsst-GalSim interface would be:
+    #
+    #    self.noise_and_background = ExampleCCDNoise(addNoise=True,
+    #                                                addBackground=True)
+    #
+    # But, we need a more realistic sky model and we need to pass more than
+    # this basic info to use Peter Y's ESO sky model.
+    # We must pass obs_metadata, chip information etc...
+    self.noise_and_background = ESOSkyModel(obs_metadata, addNoise=True,
```
jchiang87: I'm wondering of we should allow `self.noise_and_background` and `self.PSF` to be settable through the constructor, i.e.,
```
def imSim__init__(self, phosim_objects, obs_metadata, catalog_db=None, noise_and_background=None, PSF=None):
```
and set them to the values here if the default `None` values are passed.

python/desc/imsim/skyModel.py
```diff
@@ -0,0 +1,136 @@
...
+# use instead of ExampleCCDNoise
+class ESOSkyModel(NoiseAndBackgroundBase):
+    """
+    This class wraps the GalSim class CCDNoise.  This derived class returns
+    a sky model based on the ESO model as implemented in
+    """
+
+    def __init__(self, obs_metadata, seed=None, addNoise=True, addBackground=True):
+        """
+        @param [in] addNoise is a boolean telling the wrapper whether or not
```
jchiang87: I know that this way of describing function arguments is what's used in the sims code, but for Twinkles and other DESC repos, we've decided that the numpy style docstrings are preferable.   I don't have a super strong opinion and don't really care if we mix styles, but I do think that the numpy style is easier to read.

python/desc/imsim/skyModel.py
```diff
@@ -0,0 +1,136 @@
+'''
+Classes to represent realistic sky models.
+Note that this extends the default classes located in
+sims_GalSimInterface/python/lsst/sims/GalSimInterface/galSimNoiseAndBackground.py
+'''
+
+from __future__ import absolute_import, division
```
jchiang87: I added this line so that the python 3 checks in the travisc-ci builds would pass.

python/desc/imsim/skyModel.py
danielsf: Actually, we did make the decision (about a year ago) to adopt numpy style doctrings in the sims stack.  We just haven't made the decision to budget the time to implement that decision, so all of our older code still has the '@param' format docstrings.

cwwalter: Can you summarize the difference?

cwwalter: This function is this way since I am deriving from a Sim's one that uses this function.

jchiang87: > Can you summarize the difference?

I think that the parameter names are easier to scan when reading the numpy style, plus the old sims way looks like it was based on Doxygen and so implies some C/C++-ism that aren't really appropriate for python code, e.g., function arguments should always be "[in]", i.e., assumed to be passed by value and not modified or set for output.   There are numpy style examples in this repo, but I usually look at https://github.com/numpy/numpy/blob/master/doc/HOWTO_DOCUMENT.rst.txt for guidance.

cwwalter: Ah, OK so that style is for DOxygen... I didn't know about the structured text documentation on the other style of docstrings.  Thanks.

cwwalter: Anyway, consistency is better. I'm not sure whether it is more consistent to be the same as the sims_interface (especially if it will change) or the rest of DESC code.  Probably rest of DESC code.  Feel free to change this now if you can do it quickly.  Otherwise we can do it later.  I think this code is probably going to be radically restructured when we make it work per pixel so I don't think leaving it now is terrible.

python/desc/imsim/skyModel.py
cwwalter: OK thanks.  I will try to remember this for the future.

cwwalter:
I need to update the zero point values with the new information from Javier.  I will do that in a bit.

python/desc/imsim/imSim_class_factory.py
cwwalter: So, in practice it would show up here right:

phoSimStarCatalog = desc.imsim.ImSimStars(starDataBase, obs_md)

and would look something like 

phoSimStarCatalog = desc.imsim.ImSimStars(starDataBase, obs_md, PSF=special_PSF)

This sounds OK except it seems to me that that PSF is not really a feature of the source and you will have separate entries like above for Galaxies etc.  Perhaps it should be part of the common obs_metadata (which is already repeated)?

jchiang87: I'm not sure what the right answer is.   I might have expected PSF and noise_and_background to be class-level variables so that they would automatically consistent across object types.  I'm happy to defer this question for now.  If someone really wants to change it, they can do so explicitly for each instance by setting the attribute directly.

cwwalter: OK, how about adding an issue for now to make these class level variables and we can do it separately in the future?

jchiang87: sure, I'll make one, though that would imply changes to the sims code.

danielsf: Creating images with different object types is always going to be awkward because of the way objects are stored on fatboy (different tables for stars, galaxies, asteroids, etc.).

`sims_GalSimInterface` includes a method `copyGalSimInterpreter()` in `GalSimCatalog` which allows you to copy the `self.galSimInterpreter` (which actually does the image generation) from catalog to another to ensure that all of the objects land in the same FITS files.  We could amend that method so that it also copies `self.PSF` and `self.noise_and_background` from one `GalSimCatalog` to another.

python/desc/imsim/skyModel.py
```diff
@@ -0,0 +1,136 @@
...
+import galsim
+from lsst.sims.GalSimInterface.galSimNoiseAndBackground import NoiseAndBackgroundBase
+
+
+# Code snippet from D. Kirkby.  Note the use of astropy units.
+def skyCountsPerSec(surface_brightness=21, filter_band='r',
+                    effective_area=33.212*u.m**2, pixel_size=0.2*u.arcsec):
+    # Lookup the zero point corresponding to 24 mag/arcsec**2
+    B0 = 24.
+    s0 = (dict(u=0.732, g=2.214, r=1.681, i=1.249, z=0.862, y=0.452)[filter_band] *
```
jchiang87: If these numbers may change, then perhaps they should be put in the configuration file, https://github.com/LSSTDESC/imSim/blob/master/data/default_imsim_configs, e.g., 
```
[skyModel_params]
B0 = 24.
u = 0.732
g = 2.214
r = 1.681
i = 1.249
z = 0.862
y = 0.452
```

cwwalter: That might be good (how hard is it to add?).

In the longer term I wanted to change this to using the simulation project calls to return these numbers. Now, these come from David's external calculations.  That's one of the things I was hoping to learn about on Thur's SSim meeting.

jchiang87: it's straight-forward.  I can do it, if you want.

cwwalter: OK, please do. For right now it is better than hard coding it.

cwwalter: According to @fjaviersanchez the new values are supposed to be:

```
zp_u =  0.282598538804 1 / (m2 s)
zp_g =  1.56492206349 1 / (m2 s)
zp_r =  1.3488785992 1 / (m2 s)
zp_i =  0.998783620839 1 / (m2 s)
zp_z =  0.699897205858 1 / (m2 s)
zp_y =  0.32665889564 1 / (m2 s)
```

cwwalter:
OK, I looked at the latest changes and tried doing a run (see issue #5). I think other than the question of if the value is correct I think this is ready to merge.  Do you agree Jim?

jchiang87:
yes, let's do it....I just hit the approval button.

PR #44, opened 2017-01-13
Issue/42/sersic2d pa in radians

jchiang87:
This is ready for review.


cwwalter:
Does this have to be matched with the updated galSim interface to work?

jchiang87:
Not really.  This is the correct thing to do given how those angles are fed to GalSim.

cwwalter:
OK looks good then.

Issue #25, opened 2016-10-28
Consider how to refactor the lsst-galsiminterface library and how it relates to this code.


cwwalter:
Currently, some routines in the interface library  are monkey patched here in order to separate the source input mechanism from the GalSim interface code itself.

We should consider the class layout more generally and the relationship between this driving code and the library.  We can also drive other sorts of lookups (e.g. CatSim directly) but perhaps other things might use the library too.

This is a placeholder just to remind us this is an issue.

cwwalter:
Jim and Scott met at SLAC this week to discuss the strategy to do this.

jchiang87:
This is still open.  Scott and I agreed that we should proceed with his plan from the DESC meeting, but I wanted to wait until after DC1 production to work on it again.

jchiang87:
This was addressed by #95.

PR #53, opened 2017-01-31
initialize _column_cache in imSim__init__

danielsf:


jchiang87:
Regarding the missing `_column_cache` attribute:  The problem is that the superclass' `__init__` can't be called in the default case because we don't have a `CatalogDBObject` to pass it (i.e., see the if block at the top of `imSim__init__`: https://github.com/LSSTDESC/imSim/blob/master/python/desc/imsim/imSim_class_factory.py#L44) so those attributes were never set.   Ideally, we should call the superclass constructor so that any other attributes that get used down the road will be available.  This goes back, of course, to discussions at #25 and #30.

cwwalter:
> Ideally, we should call the superclass constructor so that any other attributes that get used down the road will be available

I remember that one of the changes in Python 3 had to do with how Super was implemented/used.  Would it be different there?

cwwalter:
Just curious BTW, as you say the real solution is related to discussions in #25 etc.

jchiang87:
I'm not sure...the differences in super for python3 look largely syntactic...I don't think it would fundamentally change the issue regarding passing a CatalogDBObject.

PR #54, opened 2017-03-24
Fix seeing

danielsf:
This pull request allows us to toggle between PSF models from the command line

### DIFF
--- a/bin.src/imsim.py
+++ b/bin.src/imsim.py
@@ -10,6 +10,8 @@ import os
 import argparse
 from lsst.obs.lsstSim import LsstSimMapper
 from lsst.sims.coordUtils import chipNameFromRaDec
+from lsst.sims.GalSimInterface import SNRdocumentPSF
+from desc.imsim.skyModel import ESOSkyModel
 import desc.imsim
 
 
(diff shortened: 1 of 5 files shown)


bin.src/imsim.py
```diff
@@ -32,6 +39,12 @@ def main():
     parser.add_argument('--log_level', type=str,
                         choices=['DEBUG', 'INFO', 'WARN', 'ERROR', 'CRITICAL'],
                         default='INFO', help='Logging level. Default: "INFO"')
+    parser.add_argument('--psf', type=str, default='DoubleGaussian',
+                        choices=['DoubleGaussian', 'Kolmogorov'],
```
cwwalter: Do we want the default to be doubleGaussian or Kolmogorov?  I thought we wanted Kolmogorov...

bin.src/imsim.py
```diff
@@ -130,6 +131,8 @@ def main():
         #
         # https://confluence.slac.stanford.edu/pages/viewpage.action?spaceKey=LSSTDESC&title=SSim+2017-03-23
 
+        airmass = 1.0/np.cos(np.radians(OpsimMetaData['altitude']))
```
dkirkby: This is only accurate near zenith.  Perhaps use the better approximation in eqn (3) of Krisciunas & Schaefer:
```
airmass = (1 - 0.96 * sin(90 - altitude) ** 2) ** -0.5
```

bin.src/imsim.py
jchiang87: Sure, let's go ahead and make the default Kolmogorov.   Otherwise, everything looks good to me.

PR #55, opened 2017-04-06
set allowed_chips

jchiang87:
I've been using this for the DC1 production runs and it's been working fine.

### DIFF
--- a/bin.src/imsim.py
+++ b/bin.src/imsim.py
@@ -89,6 +89,9 @@ def main():
 
     # First simulate stars
     phoSimStarCatalog = desc.imsim.ImSimStars(starDataBase, obs_md)
+    # Restrict to a single sensor if requested.
+    if arguments.sensor is not None:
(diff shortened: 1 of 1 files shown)


cwwalter:
This looks fine.. but this just sets the list of chips.  Did we previously check in something that uses it?

jchiang87:
I'm not sure I understand your question.  Every time you run imsim.py with a single sensor selected, it will be used.  If this code isn't present, occasionally FITS images for chips that were not requested will be generated because of a bright source near its edge.

cwwalter:
I just mean all the PR does is add:

> phoSimStarCatalog.allowed_chips = [arguments.sensor]

I can't remember where the code that uses this variable is.  I'm just checking you are not forgetting to check something in.

jchiang87:
The code that uses that is in the sims_GalSimInterface code.

cwwalter:
Ah, OK got it.

PR #57, opened 2017-05-03
Fix ConfigParser to work with Python 3.

cwwalter:
This change is needed for Python 2/3 compatibility. As of version 0.16 of python-future, the ConfigParser package name alias is no longer functional. This code block is based on the way EUPS now handles this.

Could someone also check this still works with python 2?

### DIFF
--- a/python/desc/imsim/imSim.py
+++ b/python/desc/imsim/imSim.py
@@ -8,7 +8,16 @@ import warnings
 from collections import namedtuple, defaultdict
 import logging
 import gc
-import ConfigParser
+
+# python_future no longer handles configparser as of 0.16.
+# This is needed for PY2/3 compatabiloty.
+try:
+    from configparser import ConfigParser
+    from configparser import SafeConfigParser
+except ImportError:
+    from ConfigParser import ConfigParser
+    from ConfigParser import SafeConfigParser
+
 import numpy as np
 import pandas as pd
 import lsst.log as lsstLog
@@ -516,7 +525,7 @@ def read_config(config_file=None):
         config_file.
     """
     my_config = ImSimConfiguration()
-    cp = ConfigParser.SafeConfigParser()
+    cp = SafeConfigParser()
     cp.optionxform = str
     if config_file is None:
         config_file = os.path.join(lsstUtils.getPackageDir('imsim'),
--- a/python/desc/imsim/imSim.py
+++ b/python/desc/imsim/imSim.py
@@ -528,7 +528,7 @@ def read_config(config_file=None):
         config_file.
     """
     my_config = ImSimConfiguration()
-    cp = configparser.SafeConfigParser()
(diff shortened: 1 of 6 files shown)


python/desc/imsim/imSim.py
```diff
@@ -8,7 +8,16 @@
 from collections import namedtuple, defaultdict
 import logging
 import gc
-import ConfigParser
+
+# python_future no longer handles configparser as of 0.16.
+# This is needed for PY2/3 compatabiloty.
+try:
```
jchiang87: I'd prefer to do something like this in order to retain namespace scoping below:
```
try:
    import configparser
except ImportError:
    # python 2 backwards-compatibility
    import ConfigParser as configparser
```

cwwalter: OK. Happy to do it this way.  See new commit below.

cwwalter:
It stupidly did not occur to me that I could just run the unit tests manually (I ran the main executable on the test file).  But when I did I noticed two things:

1) For some reason I am seeing this in the unit tests:
[I run as python tests/something.py]

```
tests/test_config_reader.py:20: DeprecationWarning: The SafeConfigParser class has been renamed to ConfigParser in Python 3.2. This alias will be removed in future versions. Use ConfigParser directly instead.
  cp = configparser.SafeConfigParser()
```
But I do *not* see that when I just run imSim.py

2)  There is a python 3 issue in two of the unit tests (config_reader and skyModel).  cp.write(output) is resulting in:

```
  File "tests/test_config_reader.py", line 25, in setUp
    cp.write(output)
  File "/Users/Walter/LSST/Anaconda/lib/python3.6/configparser.py", line 916, in write
    self._sections[section].items(), d)
  File "/Users/Walter/LSST/Anaconda/lib/python3.6/configparser.py", line 920, in _write_section
    fp.write("[{}]\n".format(section_name))
TypeError: a bytes-like object is required, not 'str'
```

jchiang87:
I think it will be fine to replace `SafeConfigParser` with `ConfigParser`.  We don't use the "magical interpolation" feature anyways, so `ConfigParser` should be fine.   Regarding the `TypeError`, can you try changing the line
```
        with open(self.test_config_file, 'wb') as output:
```
to
```
        with open(self.test_config_file, 'w') as output:
```
I don't know why I made the mode `'wb'`, but it's clearly not needed in these cases.

cwwalter:
> I think it will be fine to replace SafeConfigParser with ConfigParser. We don't use the "magical interpolation" feature anyways, so ConfigParser should be fine.

I did this.  Actually I was mostly curious why I wasn't seeing the error in the main script..

I also made the changes for the TypeError in the test programs.  Everything now runs OK on Python 3.

jchiang87:
Looks good to me.  I ran the unit test and the `imsim.py` executable and everything checks out on python 2.

PR #61, opened 2017-06-21
compute apparent magnitudes from input catalog info

jchiang87:


cwwalter:
I was doing a check of old emails to make sure there was nothing I missed and came across this. 

Somehow, I missed noticing it at the time.  The code looks fine but I have a quick question:

 I guess you used this in the pipeline to produce the true magnitudes in the star and galaxy catalog you made from the instance catalog. Is that right?

If we instead produce the catalogs directly from CatSim (like in LSSTDESC/SSim_DC1#41) would will still need this?  And do you still want to add it any case?

jchiang87:
> I guess you used this in the pipeline to produce the true magnitudes in the star and galaxy catalog you made from the instance catalog. Is that right?

Yes, that's right.

> If we instead produce the catalogs directly from CatSim (like in LSSTDESC/SSim_DC1#41) would will still need this? And do you still want to add it any case?

It doesn't hurt to merge anyways, even if it's functionality is superseded later.

cwwalter:
OK, this looks good. I will merge.

PR #66, opened 2017-08-09
Issue/6/realistic electronics chain

jchiang87:
I think this is ready to be merged.  There is a lot still to do, but it seems to have the minimum level of functionality.  Once it is merged, additional work can be broken down into separate issues and developed along their own branches.   Some of that work will involve changes to other code in this repo, so working off of master will be cleaner.


cwwalter:
> Some of that work will involve changes to other code in this repo, so working off of master will be cleaner.

Sounds great. I will merge.

PR #69, opened 2017-09-22
implement correct raft-level NOAO keywords in multi-amp FITS files

jchiang87:


cwwalter:
I'm afraid I don't have enough knowledge of fits header keywords to evaluate this.  I think you are the expert Jim.  Could Simon or someone also take a look?

I do have two simple questions though:

- Has the output been already checked that it runs though the pipeline tasks OK?

- Pierre-Francois Leget at Stanford (no GH account I can find in DESC) saw that he couldn't use the e-image output from multiple sensors from the DC1 output in mosaic mode in DS9.  @RobertLuptonTheGood suggested that this was an issue with TAN-SIP being handled incorrectly in DS9.  Someone (Lauren?) had a workaround in HSC.   My understanding is that this is totally different issue and will just handle properly making the mosaics for the 16 amplifiers and will not fix the other problem at all.  Is that right?

jchiang87:
The Camera team does have a specification on this, so the work would be to make sure that spec is satisfied by the implementation here.  Beyond that, I don't think we need additional input from the DM team (they have been pointed at that spec and used it to implement obs_comCam.)

jchiang87:
BTW, this work isn't ready to merge yet.

cwwalter:
> BTW, this work isn't ready to merge yet.

OK got it.

jchiang87:
@cwwalter This is ready to go.   The code produces keywords and amplifier geometries that agree with actual TS8 files.

bin.src/write_amp_image_file.py
```diff
@@ -11,11 +11,17 @@
 
 parser = argparse.ArgumentParser()
 parser.add_argument("eimage_file", help="eimage file to process")
+parser.add_argument('--log_level', type=str,
```
cwwalter: Just to check to make sure I know what is going on: This work on adding loggers is just to add this functionality (which exists in imsim.py) to the electronics code correct? I mean there is nothing specific about the header keywords and it could in principle be in it's own PR.

python/desc/imsim/camera_readout.py
```diff
@@ -15,17 +15,19 @@
...
+import warnings
+from collections import namedtuple, OrderedDict
 import numpy as np
 import astropy.io.fits as fits
 import astropy.time
 import lsst.afw.geom as afwGeom
 import lsst.afw.image as afwImage
 import lsst.utils as lsstUtils
 from .focalplane_info import FocalPlaneInfo, cte_matrix
+from .imSim import get_logger
```
cwwalter: I've noticed some inconsistency in the use of this relative import notation. For example in imSim we have:

from desc.imsim import CosmicRays

Should we care about doing it in a consistent way?

python/desc/imsim/camera_readout.py
```diff
@@ -273,12 +291,19 @@ def get_amplifier_hdu(self, amp_name):
             Image HDU with the pixel data and header keywords
             appropriate for the requested sensor segment.
         """
-        hdu = fits.ImageHDU(data=self.amp_images[amp_name].getArray(),
-                            do_not_scale_image_data=True)
+        hdu = fits.ImageHDU(data=self.amp_images[amp_name].getArray().astype(np.int32))
```
cwwalter: I see you lost: "do_not_scale_image_data=True"

Is that because you are explicitly setting the type here with as type so you don't have to worry?

python/desc/imsim/camera_readout.py
```diff
@@ -273,12 +291,19 @@ def get_amplifier_hdu(self, amp_name):
...
 
         # Copy keywords from eimage primary header.
-        for key in self.eimage[0].header.keys():
-            hdu.header[key] = self.eimage[0].header[key]
+        with warnings.catch_warnings():
+            warnings.simplefilter("ignore")
+            for key in self.eimage[0].header.keys():
+                try:
+                    hdu.header[key] = self.eimage[0].header[key]
+                except ValueError as eobj:
```
cwwalter: Does this mean in those cases the copy will fail and we won't have the header? Should we "fix" them instead or is this a rare occurrence?

python/desc/imsim/camera_readout.py
```diff
@@ -448,3 +473,205 @@ def set_phosim_bboxes(amp):
     amp.setRawPrescanBBox(afwGeom.Box2I(afwGeom.Point2I(0, 1),
                                         afwGeom.Extent2I(4, 2000)))
     return amp
+
+PixelParameters = namedtuple('PixelParameters',
+                             ('''dimv dimh ccdax ccday ccdpx ccdpy gap_inx
+                                 gap_iny gap_outx gap_outy preh'''.split()))
+pixel_parameters \
+    = {'E2V': PixelParameters(2002, 512, 4004, 4096, 4197, 4200, 28,
```
cwwalter: Should these file level namespace objects be defined at the top of the file? Or is it better to have them closer to the function where you used them?

Also, I'm probably just not understanding, but rather than hardcoding in the pixel parameters, should these values be set from the parameters in the segmentation.txt file available in the amp object (same comment for the old code above in the set_phosim_bboxes function).

python/desc/imsim/camera_readout.py
```diff
@@ -448,3 +473,205 @@ def set_phosim_bboxes(amp):
...
+                              25, 26.5, 25, 10),
+       'ITL': PixelParameters(2000, 509, 4000, 4072, 4198, 4198, 27,
+                              27, 26.0, 26, 3)}
+
+# The image extension headers do not include the CCD vendor so we
+# infer that from the values of the DETSIZE keyword.
+vendors = {'[1:4096,1:4004]': 'E2V',
+           '[1:4072,1:4000]': 'ITL'}
+
+def set_noao_keywords(hdu, slot_name):
```
cwwalter: Maybe we can also put in a definition reference for the keywords?  I found this:

http://iraf.noao.edu/projects/ccdmosaic/imagedef/fitsdic.html

tests/test_camera_readout.py
```diff
@@ -73,6 +75,51 @@ def test_get_amp_image(self):
         self.assertTupleEqual(image.getArray().shape, (2020, 532))
 
 
+class NoaoKeywordTestCase(unittest.TestCase):
+    "TestCase class for raft-level NOAO mosaicking keywords"
```
cwwalter: Does  "mosaicking" refer to "NOAO Mosaic Camera" keywords as in 
http://iraf.noao.edu/projects/ccdmosaic/imagedef/mosaic/MosaicV1.html ?

or keywords related to making mosaics?  The code looks like it is  checking all keywords exists as in TS8 data.  Perhaps "mosaicking" should be left off? I'm not familiar enough to know.

cwwalter:
And one extra question related to something in the history in comment https://github.com/LSSTDESC/imSim/pull/69#issuecomment-331545384  above: Will this make it so that we can properly use the mosaic mode in ds9 to look at all of the amps together? Or is that another issue?

tests/test_camera_readout.py
jchiang87: It's not checking all of the keywords--the ones in the `exclude` tuple are being excluded.  The ones remaining in those harvested headers all pertain to mosaicking.

python/desc/imsim/camera_readout.py
jchiang87: meh.

python/desc/imsim/camera_readout.py
jchiang87: I hard-coded these here because **they** **will** **never** **change** since they are part of the vendors' designs and those are fixed.   It didn't seem worth the effort to make put this in segmentation.txt (and alter the format from phosims) or make up some new configuration file for something that will never change.

python/desc/imsim/camera_readout.py
jchiang87: This only arises for *phosim* eimage files.  We're not going to fixing phosim's file output, but I think we will want to be able to run this script on those eimage files in order to get "amplifier" files that look like real camera data.

python/desc/imsim/camera_readout.py
jchiang87: that option resulted in the files being written as floats, which isn't consistent with the real data and which does not compress as well as ints.

python/desc/imsim/camera_readout.py
jchiang87: I suppose we should care a little.  seems like the changed should be in imSim.py

bin.src/write_amp_image_file.py
jchiang87: right, nothing keyword specific here.  didn't seem worth the trouble to make a separate branch and pr.

jchiang87:
Yes, these keywords were defined and implemented so that tools like ds9 could be used to display the data properly mosaicked.

cwwalter:
OK this looks good to merge.

Issue #68, opened 2017-09-14
update imsim.py to use LSSTCameraWrapper


jchiang87:
This is related to the fixes discussed at https://github.com/LSSTDESC/SSim_DC1/issues/36

jchiang87:
@heather999 Here's the relevant comment on the PR in `sims_GalSimInterface`:  https://github.com/lsst/sims_GalSimInterface/pull/32#issuecomment-329366702

Let me know if you have any questions.

heather999:
slow to notice this @jchiang87, sorry, but I assume this still needs doing? Looking at that comment on the PR, am I incorrect to think this is just a simple one-liner?  It's a good way to get acquainted with the actual code :)
Testing using lsst_sims 2_3_7 might suffice?

cwwalter:
Just pinging @danielsf since I thought the way he wound up changing things everything was already supposed to be consistent (but I want to check).

jchiang87:
No, I had to incorporate these changes in the work on the shears.   I'm
about to do the PR for that this morning and will reference this issue.
Stay tuned.

On Oct 24, 2017 7:41 AM, "Chris Walter" <notifications@github.com> wrote:

Just pinging @danielsf <https://github.com/danielsf> since I thought the
way he wound up changing things everything was already supposed to be
consistent (but I want to check).

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/LSSTDESC/imSim/issues/68#issuecomment-339013915>, or mute
the thread

danielsf:
@heather999 Unfortunately, you are going to need the most up-to-date sims installation (2.4.0 or 2.4.1) to test this.  The underlying changes to sims_GalSimInterface do not exist in 2.3.7

jchiang87:
Closed by #72

Issue #17, opened 2016-10-24
Model weak lensing distortions


jchiang87:
For `sersic2d` objects in the phosim instance catalogs, the 8th, 9th, and 10th columns (the three following the redshift) are the weak lensing distortion parameters, `gamma1`, `gamma2`, and `kappa`.  Currently, we aren't doing anything with them (and catsim doesn't set them, so they are zero in the Twinkles instance catalogs, for example).   So the `sims_GalSimInterface` code needs to be updated to make use of them, and we need to pass those parameters from the instance catalog.  Also, we need a means of filling those values in the instance catalogs given some lensing map.

jchiang87:
@danielsf I'm going to start working on this issue.  Can you point me at where in the sims code the shear parameters can be obtained and also how that info should passed to GalSim?

jchiang87:
It looks like changes in the `sim_GalSimInterface` and `imSim` code need to occur in four places:
* Modify [`GalSimCelestialObject`](https://github.com/lsst/sims_GalSimInterface/blob/master/python/lsst/sims/GalSimInterface/galSimCelestialObject.py#L16) to include the WL shear and convergence parameters.
* Modify [`GalSimBase.get_fitsFiles`](https://github.com/lsst/sims_GalSimInterface/blob/master/python/lsst/sims/GalSimInterface/galSimCatalogs.py#L356) to pass those parameters to the `GalSimCelestialObject` constructor.
* Modify [`GalSimInterpreter.drawSersic`](https://github.com/lsst/sims_GalSimInterface/blob/master/python/lsst/sims/GalSimInterface/galSimInterpreter.py#L355) to apply the lens model parameters to the galaxy model before convolving with the PSF.
* Modify `imSim`'s [`extract_objects`](https://github.com/LSSTDESC/imSim/blob/master/python/desc/imsim/imSim.py#L239) function to add the WL parameters to the `phosim_galaxies` dataframe.

danielsf:
Sorry I didn't see this thread when you first posted it.  I agree that those are the changes we would need.

cwwalter:
@jchiang87 This can be closed correct?

jchiang87:
closed by #72

PR #72, opened 2017-10-24
Apply WL shears

jchiang87:
This addresses #17.   Since I needed to make [corresponding changes to `sims_GalSimInterface`](https://github.com/lsst/sims_GalSimInterface/pull/35), I also made the updates for #68 .

Here's a comparison of the output of `imsim.py` using [shear_test_instcat.txt](https://github.com/LSSTDESC/imSim/files/1412472/shear_test_instcat.txt) (left image) with the corresponding sheared galaxies rendered using direct calls to `galsim` (right image, [galsim_shear_test_py](https://github.com/LSSTDESC/imSim/files/1412480/galsim_shear_test_py.txt)):

shear_comparison

### DIFF
--- a/bin.src/imsim.py
+++ b/bin.src/imsim.py
@@ -12,11 +12,8 @@ import numpy as np
 from lsst.obs.lsstSim import LsstSimMapper
 from lsst.sims.coordUtils import chipNameFromRaDec
 from lsst.sims.GalSimInterface import SNRdocumentPSF
-try:
-    from lsst.sims.GalSimInterface import Kolmogorov_and_Gaussian_PSF
-except ImportError:
-    # in case we are running with an old version of lsst_sims
-    pass
+from lsst.sims.GalSimInterface import LSSTCameraWrapper, GalSimCameraWrapper
+from lsst.sims.GalSimInterface import Kolmogorov_and_Gaussian_PSF
 from desc.imsim.skyModel import ESOSkyModel
 import desc.imsim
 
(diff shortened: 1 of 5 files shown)


cwwalter:
Just to be clear: I can visually see differences.  Should the be the same or different? Or is it just an issue of particular DS9 settings.

jchiang87:
Since we've had a lot of issues with rotations and coordinate systems in the sims code, the purpose of this comparison is just to show that the conventions on the angles at which the galaxies are sheared are consistent between what ultimately comes out of imsim and what one would get from galsim without all of that intervening machinery.   It would take a certain amount of fiddling with fluxes, backgrounds, and placement on the image plane (which would not really be germane to this work) to get images that looked identical in some statistical sense.  In any case, if there is more relevant testing that needs to be done, please suggest something.

cwwalter:
> Since we've had a lot of issues with rotations and coordinate systems in the sims code, the purpose of this comparison is just to show that the conventions on the angles at which the galaxies are sheared are consistent between what ultimately comes out of imsim and what one would get from galsim without all of that intervening machinery.

Ah OK got it, I just wasn't sure from the text if the implication was that they should be the same. 

I guess to really test it we  should do a closed loop test.  Produce a galaxy with the shear parameters filled in and the use HSM to measure it and see if it makes sense.  I'm not sure if we need to do a statical study or just make a perfect gaussian galaxy without noise with shear applied and then see that we see the correct shear.

rmandelb:
I'd do it without noise and with a relatively large shear, and a few cases, like (g1,g2)=(0.5,0), (0,0.5).  Doing shears separately in the two components can be useful to ensure that the coordinate system is being defined with a consistent handedness.  If you're concerned about validating conventions like shear vs. reduced shear, you should also check for consistency in cases with non-negligible magnification.

jchiang87:
I think this is ready to merge.  Here's the "test card" (from Joe Z via Eve) showing the convention for the sheared galaxy orientations in galsim:
galsim_test_card
and here is an image generated using this [instance catalog](https://github.com/LSSTDESC/imSim/blob/a7865c1f87996e22ae6b954dcf9b3f8f5c6c33ea/tests/shear_test_instcat.txt):
shear_test_image
The orientation of the upper two images match the test card wrt the increasing directions of RA and Dec.
Unless I hear otherwise, I will merge this branch in the next day or so.

PR #75, opened 2017-11-02
replaced skyCountsPerSec with calls from photUtils

ellongley:
This is the function and parameters from photUtils that I believe we want to use.

### DIFF
--- a/python/desc/imsim/skyModel.py
+++ b/python/desc/imsim/skyModel.py
@@ -14,21 +14,23 @@ import lsst.sims.skybrightness as skybrightness
 import galsim
 from lsst.sims.GalSimInterface.galSimNoiseAndBackground import NoiseAndBackgroundBase
 
+from lsst.sims.photUtils import Sed
+
 from .imSim import get_config
 
 __all__ = ['skyCountsPerSec', 'ESOSkyModel', 'get_skyModel_params']
 
 
-# Code snippet from D. Kirkby.  Note the use of astropy units.
-def skyCountsPerSec(surface_brightness=21, filter_band='r',
-                    effective_area=32.4*u.m**2, pixel_size=0.2*u.arcsec):
-    pars = get_skyModel_params()
-    # Lookup the zero point corresponding to 24 mag/arcsec**2
-    s0 = pars[filter_band] * u.electron / u.s / u.m ** 2
-
-    # Calculate the rate in detected electrons / second
-    dB = (surface_brightness - pars['B0']) * u.mag(1 / u.arcsec ** 2)
-    return s0 * dB.to(1 / u.arcsec ** 2) * pixel_size ** 2 * effective_area
+def skyCountsPerSec(skySpec, bandpass, photParams):
+	# Calculate the rate in detected electrons / second.
+  
+	skySed = Sed(wavelen = skySpec.wave, flambda = skySpec.spec)
+	
+	skycounts = skySED.calcADU(bandpass, photParams)
+	
+	skycounts_persec = skycounts / photParams.exptime / photParams.nexp / photParams.gain * u.electron / u.s / u.m ** 2
+	
+	return skycounts_persec
 
 
 # Here we are defining our own class derived from NoiseAndBackgroundBase for
@@ -96,15 +98,16 @@ class ESOSkyModel(NoiseAndBackgroundBase):
 
         bandPassName = self.obs_metadata.bandpass
         skyMagnitude = skyModel.returnMags()[bandPassName]
+        skySpec = skyModel.returnWaveSpec()
 
         # Since we are only producing one eimage, account for cases
         # where nsnap > 1 with an effective exposure time for the
         # visit as a whole.  TODO: Undo this change when we write
         # separate images per exposure.
-        exposureTime = photParams.nexp*photParams.exptime*u.s
+        exposureTime = photParams.nexp*photParams.exptime
 
-        skyCounts = skyCountsPerSec(surface_brightness=skyMagnitude,
-                                    filter_band=bandPassName)*exposureTime
+        # bandpass is the CatSim bandpass object
+        skyCounts = skyCountsPerSec(skySpec, bandpass, photParams)*exposureTime*u.s
 
         # print "Magnitude:", skyMagnitude
         # print "Brightness:", skyMagnitude, skyCounts
--- a/python/desc/imsim/skyModel.py
+++ b/python/desc/imsim/skyModel.py
@@ -19,10 +19,10 @@ from lsst.sims.photUtils import Sed
 
 from .imSim import get_config
 
-__all__ = ['skyCountsPerSec', 'ESOSkyModel', 'get_skyModel_params']
+__all__ = ['SkyCountsPerSec', 'ESOSkyModel', 'get_skyModel_params']
 
 
-class skyCountsPerSec():
+class SkyCountsPerSec():
     """
     This is a class that is used to calculate the number of sky counts per
     second.
@@ -47,13 +47,13 @@ class skyCountsPerSec():
 
     def __call__(self, filter_name='u', magNorm=None):
         """
-        This method calls the skyCountsPerSec object and calculates the sky
+        This method calls the SkyCountsPerSec object and calculates the sky
         counts.
 
         @param [in] filter_name is a string that indicates the name of the filter
         for which to make the calculation.
 
-        @param [in] magNorm is an option to calculate the skycounts for a given
+        @param [in] magNorm is an option to calculate the sky counts for a given
         magnitude.  When calculating the counts from just the information in skyModel
         this should be set as MagNorm=None.
         """
@@ -147,7 +147,7 @@ class ESOSkyModel(NoiseAndBackgroundBase):
 
         exposureTime = photParams.nexp * photParams.exptime
 
-        skycounts_persec = skyCountsPerSec(skyModel, photParams, bandPassdic)
+        skycounts_persec = SkyCountsPerSec(skyModel, photParams, bandPassdic)
         skyCounts = skycounts_persec(bandPassName) * exposureTime * u.s
 
         # print "Magnitude:", skyMagnitude
--- a/python/desc/imsim/skyModel.py
+++ b/python/desc/imsim/skyModel.py
@@ -15,28 +15,62 @@ import galsim
 from lsst.sims.GalSimInterface.galSimNoiseAndBackground import NoiseAndBackgroundBase
 
 from lsst.sims.photUtils import BandpassDict
-
 from lsst.sims.photUtils import Sed
 
 from .imSim import get_config
 
 __all__ = ['skyCountsPerSec', 'ESOSkyModel', 'get_skyModel_params']
 
-#galBandpassDict = BandpassDict.loadTotalBandpassesFromFiles(['u','g', 'r', 'i', 'z', 'y'])
+
+class skyCountsPerSec():
+    """
+    This is a class that is used to calculate the number of sky counts per
+    second.
+    """
+    def __init__(self, skyModel, photParams, bandpassdic):
+        """
+
+        @param [in] skyModel is an instantation of the skybrightness.SkyModel
+        class that carries information about the sky for the current conditions.
+
+        @photParams [in] is an instantiation of the
+        PhotometricParameters class that carries details about the
+        photometric response of the telescope.
+
+        @bandpassdic [in] is an instantation of the Bandpassdict class that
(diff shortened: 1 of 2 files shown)


ellongley:
I'm not sure this is right, as I get a different result for counts than the original function, @fjaviersanchez, thank you so much for the help, am I missing a step ? 

Thank you  ! 

Emily

jchiang87:
Hi Emily, I'm looking at your code now.  I'll have a number of comments/suggestions.  In the meantime, could you add a unit test in tests/test_skyModel.py that checks the counts you expect for some set of input parameters?  Any other testing you think would be useful would be good to add there as well.

jchiang87:
@ellongley Which version of the `lsst.sims.skybrightness` package are you using for your work?   I am using a sims stack based on a weekly from September, and the interface appears to be different from the one you are using.   Here is the version info for my setup:
```
In [6]: skybrightness.__version__
Out[6]: '2.3.7.sims+2'
```
and the error I am seeing:
```
ValueError                                Traceback (most recent call last)
/gpfs/slac/kipac/fs1/g/desc/imsim_deep/imSim/tests/skybrightness_example.py in <module>()
      2 
      3 skyModel = skybrightness.SkyModel(mags=True)
----> 4 skySpec = skyModel.returnWaveSpec()

/gpfs/slac/kipac/fs1/g/desc/imsim_deep/DMstack/LSST_Stack_2017-09-11/lsstsw/stack/Linux64/sims_skybrightness/2.3.7.sims+2/python/lsst/sims/skybrightness/skyModel.py in returnWaveSpec(self)
    634         """
    635         if self.mags:
--> 636             raise ValueError('SkyModel set to interpolate magnitudes. Initialize object with mags=False')
    637         # Mask out high airmass points
    638         # self.spec[self.mask] *= 0

ValueError: SkyModel set to interpolate magnitudes. Initialize object with mags=False
> /gpfs/slac/kipac/fs1/g/desc/imsim_deep/DMstack/LSST_Stack_2017-09-11/lsstsw/stack/Linux64/sims_skybrightness/2.3.7.sims+2/python/lsst/sims/skybrightness/skyModel.py(636)returnWaveSpec()
    634         """
    635         if self.mags:
--> 636             raise ValueError('SkyModel set to interpolate magnitudes. Initialize object with mags=False')
    637         # Mask out high airmass points
    638         # self.spec[self.mask] *= 0

ipdb>                                                                           
```
@danielsf Any advice on this?

jchiang87:
Changing the `mags` keyword as the error message suggests, I get this:
```

**Comment by danielsf** [comment]
`sims_skybrightness` hasn't changed in a meaningful way since August, so sims2.3.7 should not be different from the most up-to-date version.

@yoachim, does this failure ring any bells for you?

**Comment by danielsf** [comment]
It looks like `self.wave` gets created by `self._interpSky()`, which only gets called by running `self.setParams()` or `self.setRaDecMjd()`.  Does that help?

**Comment by ellongley** [comment]
Thank you @jchiang87, working on a unit test. I'm not sure what this means I get "12.0-5-g750978f" when I check the version.

**Comment by yoachim** [comment]
Yup, looks like you called for a spectrum before telling the model what RA,Dec and MJD to set to.  I'll add an error message there to make it more clear.

**Comment by jchiang87** [comment]
@ellongley  I'm afraid you may be developing against an outdated version of the sims code.   Do you have access to a sims Stack that is a recent versioned installation, e.g. v14_0?  If you have a NERSC account, you could use one of the installations available there.

**Comment by ellongley** [comment]
Ok, thank you @jchiang87 , I'll make sure I'm using the latest version !

File: python/desc/imsim/skyModel.py [hunk 1]
```diff
@@ -14,21 +14,23 @@
...
-
-    # Calculate the rate in detected electrons / second
-    dB = (surface_brightness - pars['B0']) * u.mag(1 / u.arcsec ** 2)
-    return s0 * dB.to(1 / u.arcsec ** 2) * pixel_size ** 2 * effective_area
+def skyCountsPerSec(skySpec, bandpass, photParams):
+	# Calculate the rate in detected electrons / second.
+  	
+	skySed = Sed(wavelen = skySpec.wave, flambda = skySpec.spec[0,:])
+	
+	skycounts = skySED.calcADU(bandpass, photParams)
```
jchiang87: I don't understand how this line can execute without an error.  `skySED` on this line should be `skySed` as defined on line 27.

python/desc/imsim/skyModel.py
```diff
@@ -14,21 +14,23 @@
...
-    # Lookup the zero point corresponding to 24 mag/arcsec**2
-    s0 = pars[filter_band] * u.electron / u.s / u.m ** 2
-
-    # Calculate the rate in detected electrons / second
-    dB = (surface_brightness - pars['B0']) * u.mag(1 / u.arcsec ** 2)
-    return s0 * dB.to(1 / u.arcsec ** 2) * pixel_size ** 2 * effective_area
+def skyCountsPerSec(skySpec, bandpass, photParams):
+	# Calculate the rate in detected electrons / second.
+  	
+	skySed = Sed(wavelen = skySpec.wave, flambda = skySpec.spec[0,:])
```
jchiang87: The `skySpec` object from line 101 below is a tuple.  See the current master of the `sims_skybrightness` package [here](https://github.com/lsst/sims_skybrightness/blob/master/python/lsst/sims/skybrightness/skyModel.py#L634).
Given the way you are using `skySpec` in this function call, you should just pass `skyModel` directly as the first argument of your function call.

python/desc/imsim/skyModel.py
```diff
@@ -96,15 +98,16 @@ def addNoiseAndBackground(self, image, bandpass=None, m5=None,
 
         bandPassName = self.obs_metadata.bandpass
         skyMagnitude = skyModel.returnMags()[bandPassName]
+        skySpec = skyModel.returnWaveSpec()
```
jchiang87: This line is not needed.  See comment above.

python/desc/imsim/skyModel.py
```diff
@@ -96,15 +98,16 @@ def addNoiseAndBackground(self, image, bandpass=None, m5=None,
...
         # where nsnap > 1 with an effective exposure time for the
         # visit as a whole.  TODO: Undo this change when we write
         # separate images per exposure.
-        exposureTime = photParams.nexp*photParams.exptime*u.s
+        exposureTime = photParams.nexp*photParams.exptime
 
-        skyCounts = skyCountsPerSec(surface_brightness=skyMagnitude,
-                                    filter_band=bandPassName)*exposureTime
+        # bandpass is the CatSim bandpass object
+        skyCounts = skyCountsPerSec(skySpec, bandpass, photParams)*exposureTime*u.s
```
jchiang87: Instead of `skySpec`, you should just pass `skyModel`.

python/desc/imsim/skyModel.py
```diff
@@ -14,21 +14,23 @@
...
-                    effective_area=32.4*u.m**2, pixel_size=0.2*u.arcsec):
-    pars = get_skyModel_params()
-    # Lookup the zero point corresponding to 24 mag/arcsec**2
-    s0 = pars[filter_band] * u.electron / u.s / u.m ** 2
-
-    # Calculate the rate in detected electrons / second
-    dB = (surface_brightness - pars['B0']) * u.mag(1 / u.arcsec ** 2)
-    return s0 * dB.to(1 / u.arcsec ** 2) * pixel_size ** 2 * effective_area
+def skyCountsPerSec(skySpec, bandpass, photParams):
+	# Calculate the rate in detected electrons / second.
```
jchiang87: Instead of tabs, can you use 4 spaces for indentation?  You might want to run `autopep8` on your code to fix this sort of thing, but more generally, you should configure your editor to use standard PEP-8 formatting.

python/desc/imsim/skyModel.py
jchiang87: You are passing `bandpass` here, but the default value of `bandpass` is `None` in the `addNoiseAndBackground` function signature.  However, `.calcADU` expects a `Bandpass` object and crashes if `None` is passed, so `bandpass` either as to be changed to be a required argument (i.e., not a keyword argument) or if it is left as `None`, then an acceptable value of `bandpass` needs to be set and passed to `calcADU`.  More generally though, we need to think somewhat more carefully about where the appropriate `Bandpass` object(s) should be created.  Currently, the bandpass is only expressed as a string, i.e., `'u', 'g', 'r', 'i', 'z', or 'y'`, in the calling code.

tests/test_skyModel.py
```diff
@@ -72,6 +72,24 @@ def test_nexp_scaling(self):
...
+     
+    def test_skycounts_function(self):
+        """
+        Test that the sky counts per sec function gives the right result for hte previously 
+        calculated zero points. (Defined as the number of electrons per second for 
+        a 24 magnitude source, the default for skyCountsperSec will be set to u band and 24 mag.)
+        """
+        
+        desc.imsim.read_config()
+    	instcat_file = os.path.join(os.environ['IMSIM_DIR'], 'tests',
```
jchiang87: This line has incorrect indentation and should crash your test program.  Can you actually execute this code without it failing?  When I try to run this (after fixing the above errors), I get a `TabError`.

ellongley:
Hi @jchiang87 thank you so much for the comments, will make sure test_skymodel runs on my latest copy.

ellongley:
For reference I'm putting here what I'm currently getting from photUtils: 

u: 0.282598605763
g: 1.56492847516
r: 1.34887891881
i: 0.998799323673
z: 0.699899668666
y: 0.326677101868

python/desc/imsim/skyModel.py
ellongley: OK! Is it better to pass skyModel as an argument or have skyCountsPerSec be a function of the ESOSkyModel class?

ellongley: OK I was going to create a bandpass dictionary in the sky counts function using photutils' galBandpassDict function and then grab the right one with the string name still as an argument of the skyCountsPerSec function, or is it better to create the dictionary outside the skyCountsPerSec function ? thank you !

danielsf: The risk with instantiating the dict in the `skyCountsPerSec` function is that, each time that function is called, you will be doing I/O reading in the dict.  Fortunately, in python, functions can have attributes (that may not be what they're called, but they function pretty similarly).  This will allow you to only instantiate the dict once using code like

```
def skyCountsPerSec(*args, **kwargs):

    if not hasattr(skyCountsPerSec, '_bp_dict'):
        skyCountsPerSec._bp_dict = [code to create the dict]
```
you can then refer to the dict as `skyCountsPerSec._bp_dict` within `skyCountsPerSec` and have access to it.

python/desc/imsim/skyModel.py
```diff
@@ -14,32 +14,46 @@
...
+
+
 class ESOSkyModel(NoiseAndBackgroundBase):
     """
     This class wraps the GalSim class CCDNoise.  This derived class returns
     a sky model based on the ESO model as implemented in
     """
 
-    def __init__(self, obs_metadata, seed=None, addNoise=True, addBackground=True):
+    def __init__(
```
jchiang87: It would be good to avoid format changes like this that are non-idiomatic for this repo, especially if the original code already follows the nominal coding conventions (e.g., pep8).

ellongley:
Ok, here are my latest changes.   I found a missing factor of (0.2**2) *(32.4) or the pixel_scale * effective_area = 1.296, which corresponds to 767/ 592 = 1.296. I also changed the function to be a class with a __call__ method.  I also deleted the stray comments and added docstrings, Let me know of any further changes to make / anything else to add to the class or the test. Thank you so much !

python/desc/imsim/skyModel.py
```diff
@@ -14,25 +14,68 @@
...
 __all__ = ['skyCountsPerSec', 'ESOSkyModel', 'get_skyModel_params']
 
 
-# Code snippet from D. Kirkby.  Note the use of astropy units.
-def skyCountsPerSec(surface_brightness=21, filter_band='r',
-                    effective_area=32.4*u.m**2, pixel_size=0.2*u.arcsec):
-    pars = get_skyModel_params()
-    # Lookup the zero point corresponding to 24 mag/arcsec**2
-    s0 = pars[filter_band] * u.electron / u.s / u.m ** 2
+class skyCountsPerSec():
```
jchiang87: Can you change this to
```
class SkyCountsPerSec(object):
```
and replace `skyCountsPerSec` everywhere else with `SkyCountsPerSec`?  This is a PEP-8 convention.

PR #80, opened 2017-11-15
Issue/76/add cosmic rays

jchiang87:


jchiang87:
This is ready to merge. 

Here are two zoomed-in, single amplifier images of 500s darks taken with an ITL sensor and a simulated 500s dark using this code:
cr_image_comparison
The noise, background, and stretches are all roughly the same for the three images.

cwwalter:
Awesome, what rate are you using? The telescope altitude? Or is is set to be what you see in the dewers?

jchiang87:
This is using rate at the ITL test stand in Tucson, so if we wanted to be really precise, we could adjust it for the actual site.  For 30 s exposures, we’d get ~1-2 events per segment.

cwwalter:
> This is using rate at the ITL test stand in Tucson, so if we wanted to be really precise, we could adjust it for the actual telescope site. For 30 s exposures, we’d get ~1-2 events per segment.

I think if we know the number (It is 1-2/sec/segment?) we might as well use that.  Since we are interested in things like false positives for SN, it's good to get the rate right.

jchiang87:
The 1-2 per segment number is for Tucson.  Not sure what it would be at the LSST site...

cwwalter:
How about we make the rate a parameter that can be set in the options file?  Then we could use the setup either for at the site or in test stands.  We can start by just setting it to 1-2/segment/30sec as you already have.

I've asked on slack to try to get a pointer to someone who knows the rate at the Blaco site on DECam which should be the same.

jchiang87:
I've added changes to read the CR rate from the config file.

jchiang87:
@cwwalter I merged current master into this branch so that the unit tests and imsim.py application will run successfully against the recent changes in `sims_GalSimInterface` (this will address the failure you noted on Slack).   Unless I hear otherwise, I will merge this branch into master after 5pm PT today.

cwwalter:
OK sounds good.  Feel free to do it now if you want.  We should make sure there is a issue about setting the rate based on further work from DECam etc.

jchiang87:
Closes #76

Issue #46, opened 2017-01-13
Make the Sky model vary over the focal plane


cwwalter:
Issue #5 implemented the ESO sky model.  However, now the level is set solely by the pointing at the center of the focal plane.

The code should be modified to vary per chip and then per pixel.  This probably means refactoring the code to recalculate the sky level in a healpix map once outside of the chip loop.

dkirkby:
How is `skyModel.ESOSkyModel.addNoiseAndBackground()` being called in the current simulations?  Once per chip, or once per focal plane, or something else?  I suspect that tabulating the sky model on a healpix map will be more trouble than it is worth, given that it is straightforward to evaluate the sky level directly at the center of each chip (or pixel).

jchiang87:
once per chip.

jchiang87:
If we want a healpix map of the sky brightness for a given set of observing conditions, e.g., read in from an instance catalog, I think it makes more sense just to provide a separate application that provides that rather than build one routinely in the `imsim.py` runs on a per-visit basis.

fjaviersanchez:
@jchiang87 I am kind of confused. So imSim calls this function (in the current implementation) once per chip but supplies the `pointingRA` and dec from the center of the focal plane?

jchiang87:
Yes, it's just that the original implementation didn't do the per chip coordinate calculation that Bela is adding.

dkirkby:
> but supplies the pointingRA and dec from the center of the focal plane?

No, it gets the pointing from `self.obs_metadata`

fjaviersanchez:
yes, but `self.obs_metadata` contains the center of the focal plane and not per-chip

belaa:
What would be the most sensible output in this case? Should it give the sky level per chip by passing a new parameter to `skyModel.ESOSkyModel.addNoiseAndBackground()`, or should it simulate the whole focal plane and then each chip can be queried separately?

cwwalter:
Thanks all, and thanks for the original work on this @belaa!

PR #81, opened 2017-11-27
Vary skymodel

belaa:
This PR addresses Issue #46 and modifies `skyModel.py` to give a different sky background for each chip on the focal plane. It includes a new `chipName` parameter in `skyModel.ESOSkyModel.addNoiseAndBackground()` which adds the sky level based on the value at the center of the chip. This is still a work in progress and any feedback is appreciated.


python/desc/imsim/skyModel.py
```diff
@@ -31,6 +39,80 @@ def skyCountsPerSec(surface_brightness=21, filter_band='r',
     return s0 * dB.to(1 / u.arcsec ** 2) * pixel_size ** 2 * effective_area
 
 
+def get_chip_names_centers():
```
jchiang87: It doesn't appear that this function is used anywhere, so it should be deleted.

python/desc/imsim/skyModel.py
```diff
@@ -87,46 +169,62 @@ def addNoiseAndBackground(self, image, bandpass=None, m5=None,
         @param [out] the input image with the background and noise model added to it.
         """
 
-        # calculate the sky background to be added to each pixel
+        #obs_db = 'minion_1016_sqlite_new_dithers.db'
```
jchiang87: Please delete these commented out lines, 172-181.

python/desc/imsim/skyModel.py
```diff
@@ -87,46 +169,62 @@ def addNoiseAndBackground(self, image, bandpass=None, m5=None,
...
+                    yPix=center_y, chipName=chipName, camera=camera, 
+                    obs_metadata=self.obs_metadata, epoch=2000.0, includeDistortion=True)
         mjd = self.obs_metadata.mjd.TAI
-        skyModel.setRaDecMjd(ra, dec, mjd, degrees=True)
-
-        bandPassName = self.obs_metadata.bandpass
-        skyMagnitude = skyModel.returnMags()[bandPassName]
+        skyModel.setRaDecMjd([ra], [dec], mjd=mjd, degrees=True)
+         
+        bandPassName = self.obs_metadata.bandpass # obs_metadata returns y-band
```
jchiang87: This comment, `# obs_metadata returns y-band`, isn't correct.  In this context, `self.obs_metadata.bandpass` should be set to whatever was specified in the input instance catalog.

tests/test_skyModel.py
```diff
@@ -65,10 +65,10 @@ def test_nexp_scaling(self):
         skymodel = desc.imsim.ESOSkyModel(obs_md, addNoise=False,
                                           addBackground=True)
         image_2 = galsim.Image(100, 100)
-        image_2 = skymodel.addNoiseAndBackground(image_2, photParams=photPars_2)
+        image_2 = skymodel.addNoiseAndBackground(image_2, 'R:4,2 S:1,0', photParams=photPars_2)
 
         image_1 = galsim.Image(100, 100)
-        image_1 = skymodel.addNoiseAndBackground(image_1, photParams=photPars_1)
+        image_1 = skymodel.addNoiseAndBackground(image_1, 'R:4,2 S:1,0', photParams=photPars_1)
```
jchiang87: There should be a test comparing the sky background on two different chips which verifies that the median background levels are different.

python/desc/imsim/skyModel.py
```diff
@@ -6,16 +6,24 @@
 
 from __future__ import absolute_import, division
 
+import matplotlib.pyplot as plt
```
jchiang87: Please avoid adding unused `import`s like this.

python/desc/imsim/skyModel.py
```diff
@@ -62,7 +144,7 @@ def __init__(self, obs_metadata, seed=None, addNoise=True, addBackground=True):
         else:
             self.randomNumbers = galsim.UniformDeviate(seed)
 
-    def addNoiseAndBackground(self, image, bandpass=None, m5=None,
+    def addNoiseAndBackground(self, image, chipName, bandpass=None, m5=None,
```
jchiang87: This method is called by `GalSimInterpreter.drawImage` in the `sims_GalSimInterface` package, so we would need to modify that function call in that package to use this code.  We need to discuss with @danielsf  about the implications of this interface change since it would force the `NoiseAndBackgroundBase` class to change.

python/desc/imsim/skyModel.py
```diff
@@ -87,46 +169,62 @@ def addNoiseAndBackground(self, image, bandpass=None, m5=None,
...
+        #c = conn.cursor()
+        #c.execute("SELECT obsHistID, moonPhase, dist2Moon, moonAlt FROM ObsHistory WHERE airmass>0")
+        #rows = c.fetchall()
+	
+	#obsid = 7177
+
+        #gen = ObservationMetaDataGenerator(database=obs_db, driver='sqlite')
+        #obs_md = gen.getObservationMetaData(obsHistID=rows[obsid][0], boundLength=3)[0]
+
+        camera = LsstSimMapper().camera
```
danielsf: There is now a method `lsst.sims.coordUtils.lsst_camera` that is designed such that, the first time you call `lsst_camera()` anywhere, it calls `LsstSimMapper().camera` and stores the result.  All subsequent calls to `lsst_camera()` return the cached result.  We should start using this function, as it will avoid the risk of us accidentally calling the expensive `LsstSimMapper().camera` more than once.

python/desc/imsim/skyModel.py
danielsf: I think this is a great improvement.  If you could just pass `chipName` as a kwarg that defaults to `None`, rather than as a positional arg, we can update `sims_GalSimInterface` so that it expects the kwarg, but places a homogeneous sky background if `chipName=None`.

jchiang87:
@danielsf Could you have another look at this please?  I made the updates and also made a related [PR](https://github.com/lsst/sims_GalSimInterface/pull/47) in sims_GalSimInterface.  Thanks!

python/desc/imsim/skyModel.py
```diff
@@ -127,34 +155,35 @@ def addNoiseAndBackground(self, image, bandpass='u', m5=None,
         PhotometricParameters class that carries details about the
         photometric response of the telescope.  Defaults to None.
 
+        @param [in] chipName is the name of the sensor being considered,
+        e.g., "R:2,2 S:1,1".
+
         @param [out] the input image with the background and noise model added to it.
         """
+        camera = get_obs_lsstSim_camera()
```
danielsf: In the interest of not having to re-implement both `addNoiseAndBackground()` and `getNoiseModel()`, should we pass `chipName` all the way through to `getNoiseModel()` and then do all of the "central pixel to RA, Dec" calculations in `getNoiseModel()`?  Then we would only have to implement that method, which was how I originally intended the `NoiseAndBackground` functions to work.  I'm just worried about this `addNoiseAndBackground()` getting out of sync with the version in `NoiseAndBackgroundBase`.  Feel free to object if you would rather just get on with your life.

jchiang87: I'm not sure I'm understanding why we'd want to move those central pixel calculations to `getNoiseModel()`.  The only data that function needs are `skyLevel` and the read noise, which are passed in the existing interface.   Since this seems to work for now, let's use the current implementation and revisit this later if some divergence with the base class arises.

danielsf: Okay.

danielsf: Go ahead and merge (assuming my review was holding that up)

PR #84, opened 2017-12-06
updates to sky background implementation to include sensor effects

jchiang87:
Opening this PR so that design decisions (in this repo and at sims_GalSimInterface) can be discussed regarding the handling of tree rings and other sensor-specific effects.


python/desc/imsim/skyModel.py
```diff
@@ -64,7 +64,9 @@ def __init__(self, obs_metadata, seed=None, addNoise=True, addBackground=True):
 
     def addNoiseAndBackground(self, image, bandpass=None, m5=None,
                               FWHMeff=None,
-                              photParams=None):
+                              photParams=None,
+                              tree_ring_center=galsim.PositionD(0, 0),
```
jchiang87: @danielsf I'm passing the tree ring info as parameters here, but that's inconsistent with the interface of this method in the base class.   I was thinking that it would make more sense to package the tree ring info with the image data in a detector object and pass that detector object instead of `image`.  This requires more extensive changes in sims_GalSimInterface, however.   What do you think?

For reference, see the [`TreeRings`](https://github.com/rmjarvis/sims_GalSimInterface/blob/ffe3953497b88f1df98d906c689610e17adc8411/python/lsst/sims/GalSimInterface/galSimInterpreter.py#L461) class I added to `sims_GalSimInterface`.

danielsf: I'm not sure how I feel about storing image data in the `detector` object.  I'd originally envisioned `detector` as just a container for parameters pertaining to the physical detector.  I agree that this means that the tree ring parameters should probably live in `detector`.  Could we pass `detector` into `addNoiseAndBackground` alongside `image`?  I imagine that someday there will be even more detector-specific parameters that `addNoiseAndBackground` will need to know about.

jchiang87: Sure, that proposal sounds fine.  When I wrote that comment I was conflating the `GalSimInterface.detectors` attribute with the `GalSimInterface.detectorImages` attribute.  So we'd just need to add the tree ring info to the GalSimDetector class and pass in one of those to `addNoiseAndBackground`.

jchiang87:
@cwwalter @rmjarvis With the changes I just pushed, both the faster and slower versions of the sky background/sensor model code work and produce sensible numbers for the background counts:  ~850 e/pixel for a 30 sec exposure in r-band.  I estimate about ~4 hours on cori Haswell for a single sensor using the faster version.  I'm doing a more precise timing test now.  Not sure where the time is going...the version without photon bundling is only a factor of ~3 slower.   Note that for a full sensor, the non-bundling code wants to create a photon array of size ~1e10 (800\*4000\*4000) which the `galsim.PhotonArray` class rejects as too large (presumably because it expects a 32-bit int as the argument).

rmjarvis:
I don't think that's a 32 bit issue.  It is undoubtedly running out of memory.  1e10 photons need ~10^10 * 3 * 8 Bytes = 240GB of memory at initialization and then that much again to set the angles and wavelengths.

But  we don't have to do all the photons at once.  You can break up the full PhotonArray into chucks and add them separately to keep the memory under control.  Eg. have it check skyCounts * npix > some value and if it is, just recurse calling itself N times each with a skyCounts small enough that the memory is reasonable.

The effect in the silicon layer will be the same, except that it would force a recalculation of the pixel boundaries at the start of each recursed call.

rmjarvis:
Or probably better, just internally in that function loop `niter` times with and set `skyCounts = skyCounts_total / niter`.

jchiang87:
Sure, chunking the `PhotonArray` that way should work.  BTW, here's the error I saw which led me to conclude it was a 32-bit int issue:
```
>>> galsim.PhotonArray(int(1.3e10))
  File "<stdin>", line 1, in <module>
  File "/global/cscratch1/sd/jchiang8/DC2/imSim_tests/galsim_local/python/galsim/photon_array.py", line 95, in PhotonArray_init
    _PhotonArray_empty_init(self, N)
OverflowError: bad numeric conversion: positive overflow
```

rmjarvis:
I see.  Maybe so then.  On my laptop it just started trying to crank away with it until my machine started to freeze up and I killed it.  But that line will involve a cast to whatever the C++ calls `int`.  So if that's 32 bit, I guess maybe Boost.Python notices the overflow and catches it.

rmjarvis:
Oh, nvm.  I actually only tried 10**9, not 10**10.  I do indeed see that error with 10**10.  Which makes sense, since 32 bit max int is around 2e9.

craiglagegit:
I pushed a new version of the tree ring code.  There is a simple Python script in data/tree_ring_data that creates a text file (data/tree_ring_data/tree_ring_parameters_19mar18.txt) with a set of tree ring parameters.  This file is then read in at run time to assign the tree ring parameters to the sensors.  This way the focal plane will always have the same set of tree ring patterns.  Let me know if you want me to split out the code, which is currently in python/desc/imsim/imSim.py into a separate file.  I also wrote a simple tests/test_tree_rings.py that tests that it is successfully reading in the parameters and assigning them to the detector, but this test is not integrated into the unit tests.

cwwalter:
> I pushed a new version of the tree ring code. There is a simple Python script in data/tree_ring_data that creates a text file (data/tree_ring_data/tree_ring_parameters_19mar18.txt) with a set of tree ring parameters. This file is then read in at run time to assign the tree ring parameters to the sensors. This way the focal plane will always have the same set of tree ring patterns. Let me know if you want me to split out the code, which is currently in python/desc/imsim/imSim.py into a separate file. I also wrote a simple tests/test_tree_rings.py that tests that it is successfully reading in the parameters and assigning them to the detector, but this test is not integrated into the unit tests.

Hi Craig,

Thanks! I think it would feel better to me if it was structured like the cosmic ray code above it in imsim.py

have a 

from desc.imsim import TreeRings

Have a function add_treering(gs_interpreter, other parameters): [basically what is there now but renamed]

and then let it call member functions of the class TreeRings that is in 

imSim/python/desc/imsim/tree_rings.py

the inner functions  in imsim.py now don't feel very elegant.

Jim do you have a comment?

craiglagegit:
Chris,

Thanks for the comments.  I'll work on structuring it that way.

Craig

craiglagegit:
Chris, Jim,

Pushed a new version with the tree ring code partitioned into a separate file tree_rings.py.  Also cleaned up reading Raft and Sensor code from detector.name. The simple test in tests/test_tree_rings.py runs OK, but I haven't done more extensive tests.

jchiang87:
Thanks, Craig.  I'm going through your additions now and will make some comments.

tests/test_tree_rings.py
```diff
@@ -0,0 +1,43 @@
...
+    detector_list.append(make_galsim_detector(camera_wrapper, sensor, phot_params, obs_md))
+
+gs_interpreter = GalSimInterpreter(detectors=detector_list)
+
+desc.imsim.add_treering_info(gs_interpreter)
+
+r_value = 5280.0
+
+for detector in gs_interpreter.detectors:
+    print("Detector = ", detector.name)
```
jchiang87: If possible, it would be good to re-write this as a unit test that `nosetests` or `py.test` can run as part of the overall suite.  If you look at the other test programs, you'll see that they use the `unittest` package.  So instead of the print statements, I would put some `.assertEqual(<calculated value>, <expected value>)` lines.

craiglagegit:
Jim,

I think I made all of your requested changes.  Let me know if you see something you don't like.  The old file data/tree_ring_data/test_tree_ring_model_16mar18.py was what I used to test the tree ring model in GalSim, and it didn't run in this environment anyway, so I just removed it.  The unittest now runs, but all it does at present is test that the tree_ring_parameters file can be read successfully.

Craig

jchiang87:
Thanks, @craiglagegit .  All of your changes look good to me.

danielsf:
@jchiang87 If you are worried about memory usage: be aware that this code

https://github.com/lsst/sims_photUtils/blob/master/python/lsst/sims/photUtils/Sed.py#L576

caches every SED read in from `sims_sed_library` so that each file only needs to be read in once.  If that proves to be a problem for KNL, we can add a switch to `sims_photUtils` that will turn that behavior off.

jchiang87:
Thanks, @danielsf .  That caching appears to be a relatively minor contribution to the overall memory usage and does help speed up the SED calculations by factors of a few, so I think we should keep it.  The `SedWrapper` class I added was to address the fact that the sensor model handling code needs the object's SED:
https://github.com/rmjarvis/sims_GalSimInterface/blob/treerings/python/lsst/sims/GalSimInterface/galSimInterpreter.py#L335
Storing that info with each `gsObject` at the outset resulted in memory blowing up to 17GB once the instance catalog was read in.    Reading in the SED only when each object is drawn and discarding once done avoids that issue (though [memory seems to blow up for other reasons](https://confluence.slac.stanford.edu/download/attachments/193772618/imsim_update.pdf?version=2&modificationDate=1522079916000&api=v2) when the sensor model is enabled.)

data/tree_ring_data/build_tree_ring_file_19mar18.py
```diff
@@ -0,0 +1,79 @@
...
+    sigmak1 = 10.0 # Standard deviation of spatial freqency in pixels
+    numfreqs2 = 5 # Number of spatial frequencies
+    meank2 = 35.0 # Mean spatial freqency in pixels
+    sigmak2 = 10.0 # Standard deviation of spatial freqency in pixels
+
+    # This generates the parameters for radial tree ring function.
+    # Tree ring amplitude in pixels is given by A + B * r^4
+    # a list of spatial frequencies and phases is generated for each detector
+    if np.random.rand() < bad_tree_ring_fraction:
+        B = np.random.uniform(4.0E-17, 8.0E-17, 1)[0] # Amplitude factor in pixels^-3
```
danielsf: Should this function accept as an argument an instantiation of `np.random.RandomState`?  That way, we can control the seed of the random number generator and have reproducible results.

data/tree_ring_data/build_tree_ring_file_19mar18.py
```diff
@@ -0,0 +1,79 @@
...
+    xy_dist_to_center = 3000.0 # Distance from center of detector to tree ring center
+    dither_radius = 100.0
+    corner = [np.random.choice([-1.0,1.0]), np.random.choice([-1.0,1.0])]
+    origin = np.random.uniform(xy_dist_to_center - dither_radius, xy_dist_to_center + dither_radius, 2)
+    center = np.array(corner) * np.array(origin)
+    
+    return (A, B, center, cfreqs, cphases, sfreqs, sphases)
+
+
+#****************MAIN PROGRAM*****************
```
danielsf: The main program should be contained in an
```
if __name__ == "__main__":
```
block.  That is how python separates the main program from importable methods and classes.

python/desc/imsim/tree_rings.py
```diff
@@ -0,0 +1,90 @@
...
+        self.npoints = int(self.r_max / dr) + 1 # Number of points in tree ring function
+
+        with open(tr_filename, 'r') as input_:
+            self.lines = input_.readlines() # Contents of tree_ring_parameters file
+
+        return
+    
+    def Read_DC2_Tree_Ring_Model(self, Rx, Ry, Sx, Sy):
+        """
+        This function finds the tree ring parameters for a given sensor
```
danielsf: Can you expand the docstring here to define `Rx, Ry, Sx, Sy`?

python/desc/imsim/tree_rings.py
```diff
@@ -0,0 +1,90 @@
...
+                        continue
+                else:
+                    continue
+            # If we reach here, the (Rx, Ry, Sx, Sy) combination was not found.
+            raise TreeRingsError("Failed to read tree ring parameters for Rx=%d, Ry=%d, Sx=%d, Sy=%d"%(Rx, Ry, Sx, Sy))
+        except:
+            raise TreeRingsError("Failed to read tree ring parameters for Rx=%d, Ry=%d, Sx=%d, Sy=%d"%(Rx, Ry, Sx, Sy))
+        return ()
+
+    def tree_ring_radial_function(self, r):
```
danielsf: Can you add a docstring specifying what this function does and what `r` is?

python/desc/imsim/tree_rings.py
```diff
@@ -0,0 +1,90 @@
...
+        return
+    
+    def Read_DC2_Tree_Ring_Model(self, Rx, Ry, Sx, Sy):
+        """
+        This function finds the tree ring parameters for a given sensor
+        and assigns a tree ring model to that sensor.
+        """
+        try:
+            for i, line in enumerate(self.lines):
+                if line.split()[0] == 'Rx':
```
danielsf: Would it be more efficient to do the conversion to `int`s and `float`s once in the constructor, rather than storing the lines of the treering data file as strings and doing the conversion every time we call this method? 

Feel free to point out that this could only ever be called 189 times, so it's probably not worth the work to change it.

tests/test_tree_rings.py
```diff
@@ -0,0 +1,49 @@
...
+    def setUp(self):
+        self.sensors = ['R:2,2 S:1,1', 'R:3,4 S:2,2']
+        self.instcat_file = os.path.join(lsstUtils.getPackageDir('imsim'),
+                               'tests', 'tiny_instcat.txt')
+        self.rtest = 5280.0 # Just a value to test the radial function at
+        self.rvalues = [.0030205, -.0034135] # Expected results
+        self.centers = [(-3026.3, -3001.0), (3095.5, -2971.3)] # Input center values
+
+    def test_read_tree_rings(self):
+        "Check reading of tree_ring_parameters file"
```
danielsf: This should be enclosed in `"""` to be a docstring.

python/desc/imsim/skyModel.py
```diff
@@ -155,62 +180,350 @@ def addNoiseAndBackground(self, image, bandpass=None, m5=None,
...
+                                  wave_type='nm')
+            self._waves = galsim.WavelengthSampler(sed=sed,
+                                                   bandpass=gs_bandpass,
+                                                   rng=self.randomNumbers)
+        return self._waves
+
+    @property
+    def angles(self):
+        if self._angles is None:
+            fratio = 1.234  # From https://www.lsst.org/scientists/keynumbers
```
danielsf: In the future, should we add `fratio` and `obscuration` into `sims.photUtils.PhotometricParameters`, so that ImSim can be extended to non-LSST systems?

jchiang87: Yes, this is a sensible suggestion.

jchiang87:
Thanks for the review @danielsf .   @craiglagegit Do you want to handle Scott's comments on the treering code?  They're all straightforward and fairly minor, so I can address them if you want.

tests/test_tree_rings.py
jchiang87: Thanks for pointing this out! I did not know this...I've been making the same error elsewhere.

craiglagegit:
@jchiang87 I'm kind of swamped right now, so I would appreciate it if you do it.  Let me know if there are issues.  Thanks.

tests/test_tree_rings.py
danielsf: Actually... I was wrong.  What you have is fine.   I have always been using `"""` and just assumed that was because I had to.  Sorry for the officiousness.

danielsf: I guess my misconception was based on a PEP8 convention

https://www.python.org/dev/peps/pep-0257/

jchiang87: ok. [PEP 257](https://www.python.org/dev/peps/pep-0257/) does say to use `"""`, even for one-liners, so I plan to follow that convention.

jchiang87:
> @jchiang87 I'm kind of swamped right now, so I would appreciate it if you do it. Let me know if there are issues. Thanks.

@craiglagegit No problem.  I'll take care of them.  Thanks.

python/desc/imsim/tree_rings.py
jchiang87: I'm going to leave this one be.

python/desc/imsim/tree_rings.py
jchiang87: Done.

danielsf:
Whenever you feel you've addressed my review, go ahead and merge.

jchiang87:
Following @danielsf's suggestion, I updated the `build_tree_ring_file_19mar18.py` script and used a `numpy.random.RandomState` object so that we can regenerate parameter files from a given seed.  I generated a new parameter file, `tree_ring_parameters_2018-04-24.txt`, and updated the `imSim.py` code to use it.   Here is a comparison of the tree rings from that new file (right) versus the tree rings from the older file (`tree_ring_parameters_19mar18.txt`, left) for the R:2,2 S:1,1 sensor and Run1.2p visit 197356:
tree_ring_parameter_comparison

Unless I hear otherwise, I will merge this branch into master at Noon PT tomorrow, April 25.

bin.src/imsim.py
```diff
@@ -89,26 +93,22 @@ def main():
...
-                                       bandpassDict=bp_dict,
-                                       noiseWrapper=noise_and_background,
-                                       seed=arguments.seed)
+    gs_interpreter = make_gs_interpreter(obs_md, detector_list, bp_dict,
+                                         noise_and_background,
+                                         epoch=2000.0,
+                                         seed=arguments.seed,
+                                         apply_sensor_model=apply_sensor_model)
+    chip_name = arguments.sensor if arguments.sensor is not None\
+                else "R:2,2 S:1,1"
```
cwwalter: So this says if we don't specify a sensor then we select the middle sensor in the focal plane.  So if someone gives imSim a instance catalog without specifying the sensors and the sources in the catalog fall on another sensor the output will be this sensor with only the sky correct?  

Is this consistent with the parameter help for --sensor: "'If None, then simulate all sensors with sources on them'"

python/desc/imsim/imSim.py
```diff
@@ -693,3 +675,28 @@ def add_cosmic_rays(gs_interpreter, phot_params):
...
+    gs_interpreter: lsst.sims.GalSimInterface.GalSimInterpreter
+        The object that is actually drawing the images
+
+    Returns
+    -------
+    None
+        Will act on gs_interpreter, adding tree ring information to the detectors.
+    """
+    tr_filename = os.path.join(lsstUtils.getPackageDir('imsim'),
+                               'data', 'tree_ring_data', 'tree_ring_parameters_2018-04-24.txt')
```
cwwalter: Should we hardcode this tree ring file?  Maybe it should be in the config file?

bin.src/imsim.py
jchiang87: Setting `chip_name` here is only used in the next line where the `.sky_bg_per_pixel` attribute is set, and that attribute is only used to set the scale used for the `folding_threshold` that the `galsim` code uses to compute the postage stamp size for point sources.  In that application, a precise value really isn't needed, so the center of the focalplane is good enough.  In all other cases, including when simulating all of the sensors, the code in `skyModel.py` uses the sensor-specific value.  The problem is that we need to set this attribute before calling `.drawObject` below.

python/desc/imsim/imSim.py
jchiang87: I thought about that, but I didn't bother since I figured that we would change this very infrequently.  If we decide we need that sort of runtime flexibility, we can put it into the config then.

PR #89, opened 2018-01-19
Adds support for parsing and drawing a random walk component with Galsim

EiffL:
This pull request allows imSim to parse an instance catalog that contains a GalSim RandomWalk component. This complements the following PRs:
  - https://github.com/lsst/sims_GalSimInterface/pull/40 : Provides the drawing code in the Galsim interface
  - https://github.com/LSSTDESC/sims_GCRCatSimInterface/pull/2 : Enables the creation of an instance catalog with RandomWalk from protoDC2


jchiang87:
@EiffL I did a couple of simulation runs doing `generateInstCat.py` followed by `imsim.py` on a protoDC2 visit both with and without the `_addon_knots` qualifier on the `protoDC2` catalog option.  Here is a comparison of the resulting FITS files:
randomwalk_galaxy_comparison
The image on the left is the `--descqa_catalog protoDC2` case while the image on the right is for `--descqa_catalog protoDC2_addon_knots`.   The extra bright objects in the right image seem to indicate a problem.

For the record, the left image took 31cpu mins on a cori login node, while the right image took 47 cpu mins on the same node.

The catalogs, FITS files, etc. for these tests are all in 
```
/global/cscratch1/sd/jchiang8/DC2/random_walk_tests
```
at NERSC.  I used the current masters of `sims_GalSimInterface`, `gcr-catalogs`, and your forks of `sims_GCRCatSimInterface` and `imSim`.

EiffL:
Hi Jim, thanks for checking! I've been working on a descqa test to validate this and just got some similar results :-/ I'm doing some digging to find out what's going on.

EiffL:
Ok, got it, I thought the fluxes of some galaxies were blowing up but my version of imSim was "just" skipping the stars when the randomwalk catalog was empty. 

The origin of the confusion came from the copyGalSimInterpreter() function (https://github.com/lsst/sims_GalSimInterface/blob/731c209ced25f1cac51973a45f8db0ec44340623/python/lsst/sims/GalSimInterface/galSimCatalogs.py#L430), in the galsim interface, which will not actually copy the interpreter if the catalog (in this case, the randomWalk catalog) is empty and thus never gets initialized. 

I've updated the my fork and here the same plot, no knots on the left, knots on the right:
image

I think there is still an issue on the rescaling of the flux for the disk, but that would come from the GCRCatSimInterface.

cwwalter:
If you do a more of a zoom (or maybe without the sky background) can you clearly see the difference in the sub-structure?

EiffL:
Yes, it's really not obvious with the atmospheric PSF and sky background. Here is a zoom-in without background and with just a double Gaussian PSF:
image
image
The difference is obvious if I substract
image

I'm working on validation, I want to check that the overall flux is preserved and that basic properties like size and ellipticity are not deviating too much.

EiffL:
I have fixed an issue in the GCRCatalog interface which caused the flux of the disk to not be correctly re-scaled because of some caching of the catalog which bypassed some of my modifications. Everything should work as expected now but I still need to check that the knots are generated at the same position in different bands (they should as the random number generator they rely on is seeded by the object uniqueId).

I also wanted to mention that looking at noiseless imsim images, I spotted last night very few instances where the fluxes of some galaxies was changing between realizations. I suspect something is up at the level of the instance catalog generation, I don't think it comes from my RandomWalk modifications but maybe. Just wanted to let you know, I'm checking this this afternoon and will report back.

jchiang87:
Thanks for the update! When things are ready to try again, please let me know, along with all of the packages that need to be updated.

EiffL:
@jchiang87 Ok, done. The suspicious behavior I was seeing with galaxies becoming brighter was due to the current fix in the GCRCatSimInterface to handle nonsensical extinction parameters in protoDC2 (https://github.com/LSSTDESC/sims_GCRCatSimInterface/issues/5). 
This is resolved, and the implementation is working just fine, feel free to test it.

In the meantime, I'll report back on the DC2_repo issue with some results.

EiffL:
@jchiang87 @cwwalter I have merged the latest version of imSim with my branch and everything seems to run ok, it should be ready for a review.

cwwalter:
@EiffL Have you also built some tests for this feature?

danielsf:
@EiffL actually... I may have broken all of your things (sorry)

There was a recent refactor of ImSim.  It now works directly off of PhoSim InstanceCatalogs with no need to call the `imsim_class_factory`.  Since we don't include the RandomWalk objects in the PhoSim InstanceCatalogs, I didn't know how to implement them, so I didn't.  All of which is to say:

How do we want to handle RandomWalk objects in the current state of ImSim `master`

EiffL:
I'm working on it, I will have some by tonight.

cwwalter:
I think the factory class is no longer needed (oops crossed with Scott).

EiffL:
Hummmm.... I ran it and it draw the random walk, I've modified a number of
things this morning and I think it's compatible with your modifications,
but I'll double check

EiffL:
@danielsf ok I see what you mean, this imsim_class_factory didnt serve any purpose anymore, I removed it.
So the way it works is simply parsing the instance catalog and creating `GalSimCelestialObject` of type 'RandomWalk', which are defined in the galsim interface, it's very straightforward. 

The only thing is that to avoid changing too many things, the randomWalk `GalSimCelestialObject`uses the `sersic_index` field to store the number of knots.... It's not super clean but I was wary of changing the definition of `GalSimCelestialObject` to add an additional field, it can easily silently break things

danielsf:
`sims_GalSimInterface` is pretty well covered with unit tests, so you shouldn't feel too hesitant to issue PRs against it.  I think I would actually prefer that to storing something that isn't a Sersic index in `sersic_index`.

We still need to address the question of how to generate ImSim InstanceCatalogs.  One of the original goals of the refactor was to get ImSim to run off of exactly the same InstanceCatalogs as PhoSim.  PhoSim InstanceCatalogs do not have knot objects.

EiffL:
ok, so no problem, I can modify `sims_GalSimInterface`

Now, concerning the instance catalogs... is it an absolute requirement that there should be only one instance catalog ? there could be a simple instance catalog, which can be processed by both imsim and phosim, and a more complex one that, for now, only imsim can process...

Technically, it would be possible to move the code that generates the parameters of the random walk inside imSim, but it's going problematic. It might require less effort to add random walks to phosim !

cwwalter:
> We still need to address the question of how to generate ImSim InstanceCatalogs. One of the original goals of the refactor was to get ImSim to run off of exactly the same InstanceCatalogs as PhoSim. PhoSim InstanceCatalogs do not have knot objects.

Hmm.. I thought we decided that we would put these objects that were not in PhoSim in our instance catalogs.  We were giving up of having them be exactly the same.  We already changed proper motion entries etc.  Are you saying we can't do that anymore?

danielsf:
@cwwalter We certainly *can* still do that.  However, except for the RandomWalk objects, we no longer *have* to do that.  The refactored ImSim reads in a PhoSim catalog and does the right thing.  There is no longer any need to add proper motion, parallax, etc. to the ImSim InstanceCatalogs (assuming the tests I am about to run pass).

cwwalter:
> We certainly can still do that. However, except for the RandomWalk objects, we no longer have to do that.

OK good. That sounds perfect. We always want to be able to add new classes of objects as we get more sophisticated (like @EiffL's work).  

Does this mean we aren't using ICRS anymore for coordinates?

EiffL:
Ok, so I've made the modifs Scott was asking on `GalSimCelestialObject`, which involves updating `sims_GalSimInterface` (https://github.com/lsst/sims_GalSimInterface/pull/46), so until that get merged the tests for ImSim will fail.

I'm running out of time for today and won't have time to add tests until Monday, sorry.

EiffL:
Ok, I've made the requested modifications and added a small validation test of the instance catalog. I can add more tests upon request, just let me know.
Tested against my fork of `sims_GalSimInterface` all the tests pass.

jchiang87:
@EiffL Feel free to merge this when https://github.com/lsst/sims_GalSimInterface/pull/46 is merged.

EiffL:
@jchiang87 Sorry, I don't have merging powers in this repo, but everything should be ready,  lsst/sims_GalSimInterface#46 has been merged

jchiang87:
Thanks, @EiffL.  I've merged your changes.

PR #93, opened 2018-02-04
set the log level of the LsstSimMapper constructor so that the INFO s…

jchiang87:
…tatement is silenced

### DIFF
--- a/bin.src/imsim.py
+++ b/bin.src/imsim.py
@@ -9,10 +9,9 @@ from __future__ import absolute_import, print_function
 import os
 import argparse
 import numpy as np
-from lsst.obs.lsstSim import LsstSimMapper
 from lsst.sims.coordUtils import chipNameFromRaDec
 from lsst.sims.GalSimInterface import SNRdocumentPSF
-from lsst.sims.GalSimInterface import LSSTCameraWrapper, GalSimCameraWrapper
+from lsst.sims.GalSimInterface import LSSTCameraWrapper
 from lsst.sims.GalSimInterface import Kolmogorov_and_Gaussian_PSF
 from desc.imsim.skyModel import ESOSkyModel
 import desc.imsim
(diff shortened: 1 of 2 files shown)


bin.src/imsim.py
```diff
@@ -9,10 +9,9 @@
 import os
 import argparse
 import numpy as np
-from lsst.obs.lsstSim import LsstSimMapper
 from lsst.sims.coordUtils import chipNameFromRaDec
 from lsst.sims.GalSimInterface import SNRdocumentPSF
-from lsst.sims.GalSimInterface import LSSTCameraWrapper, GalSimCameraWrapper
```
cwwalter: Is GalSimCameraWrapper begin removed because it was just superfluous?

python/desc/imsim/imSim.py
```diff
@@ -566,3 +566,13 @@ def get_logger(log_level):
                      eval('lsstLog.%s' % log_level))
 
     return logger
+
+
+def get_obs_lsstSim_camera(log_level=lsstLog.WARN):
+    """
+    Get the obs_lsstSim CameraMapper object, setting the default
+    log-level at WARN in ordert to silence the INFO message about
```
cwwalter: Typo (ordert).

python/desc/imsim/imSim.py
```diff
@@ -566,3 +566,13 @@ def get_logger(log_level):
...
+    "Loading Posix exposure registry from ."
```
cwwalter: Rather than saying this is to stop that particular info message, maybe we should just say we want to be able to set the log level?  The reason we got rid of it can be explained in this pull request but we might actually want to turn this up for debugging and someday that message might disappear in any case.

cwwalter:
A few trivial comments/suggestions but otherwise looks good to merge.

bin.src/imsim.py
jchiang87: yes.

python/desc/imsim/imSim.py
jchiang87: If that's the approach you want to take, then we don't actually need the code in this PR.  The imsim.py executable already has a `--log_level` option, and if you set it to `WARN`, that info message is suppressed.  My feeling was that we want our `INFO` messages available, but that camera mapper message serves no useful purpose so we should just silence it, hence the comment in this function.

cwwalter: Ah, OK.  OK I understand.

That's fine.  Maybe also mentioning in the comment that the overall level is at INFO would make this clearer.

PR #95, opened 2018-02-14
U/danielsf/refactor

danielsf:
This PR implements the refactor Jim and I talked about at the February 2018 DESC Collaboration Meeting.  It liberates imsim from the need to pretend that it is writing out InstanceCatalogs.  It needs to be run against these branches of sims_catUtils and sims_GalSimInterface

https://github.com/lsst/sims_catUtils/tree/u/danielsf/factor_out_phosim
https://github.com/lsst/sims_GalSimInterface/tree/u/danielsf/clean_gsobj

The unit tests still pass *except*:

1) `test_camera_readout.py` does not pass, but it does not pass on `master` for me, either, so I think something more fundamental is going on.

2) `test_imsim_truth.py` does not pass.  Looking at `imsim_truth.py`, I'm not sure that we need it any more.  As a part of this refactor, I moved the `flux_dict` calculations (associating a number of electrons with a bandpass) into the `GalSimCelestialObject` class, which is what is returned by `desc.imsim.sources_from_file`.  If there is some purpose other than calculating the true flux of a source that was meant to be served by `imsim_truth.py`, we can, of course, figure out a way to preserve that functionality.

This is probably not actually ready for merging (at the very least, those `lsst_sims` branches need to be reviewed and merged).  I wanted to open this, though, so that you could comment on them.


jchiang87:
@danielsf Using lsst_sims 2.5.0, the `u/danielsf/clean_gsob` branch of `sims_GalSimInterface`, and the master of `imSim`, `test_camera_readout.py` runs fine for me.  However, when I use `u/danielsf/refactor` of `imSim', I get this error:
```
  File "/global/cscratch1/sd/jchiang8/imSim_prs/imSim/bin/imsim.py", line 170, in <module>
    main()
  File "/global/cscratch1/sd/jchiang8/imSim_prs/imSim/bin/imsim.py", line 90, in main
    numRows=numRows)
  File "/global/cscratch1/sd/jchiang8/imSim_prs/imSim/python/desc/imsim/imSim.py", line 253, in sources_from_file
    ra_appGeo, dec_appGeo = PhoSimAstrometryBase._appGeoFromPhoSim(np.radians(ra_phosim),
AttributeError: type object 'PhoSimAstrometryBase' has no attribute '_appGeoFromPhoSim'
E
```
Other tests using this branch also show this same error.  Is there some other sims package for which I need a more current version that is included in sims 2.5.0?

danielsf:
Sorry.  There is also a pull request in sims_catUtils that Bryce has already reviewed for mer

https://github.com/lsst/sims_catUtils/pull/94

jchiang87:
A couple comments to start:
* I'm getting errors from galsim that I've never encountered before, like
```
  File "/global/cscratch1/sd/jchiang8/imSim_prs/imSim/bin/imsim.py", line 171, in <module>
    main()
  File "/global/cscratch1/sd/jchiang8/imSim_prs/imSim/bin/imsim.py", line 154, in main
    gs_interpreter.drawObject(gs_obj)
  File "/global/cscratch1/sd/jchiang8/imSim_prs/sims_GalSimInterface/python/lsst/sims/GalSimInterface/galSimInterpreter.py", line 294, in drawObject
    centeredObj = self.findAllDetectors(gsObject)
  File "/global/cscratch1/sd/jchiang8/imSim_prs/sims_GalSimInterface/python/lsst/sims/GalSimInterface/galSimInterpreter.py", line 167, in findAllDetectors
    centeredObj = self.createCenteredObject(gsObject)
  File "/global/cscratch1/sd/jchiang8/imSim_prs/sims_GalSimInterface/python/lsst/sims/GalSimInterface/galSimInterpreter.py", line 437, in createCenteredObject
    centeredObj = self.drawSersic(gsObject)
  File "/global/cscratch1/sd/jchiang8/imSim_prs/sims_GalSimInterface/python/lsst/sims/GalSimInterface/galSimInterpreter.py", line 376, in drawSersic
    beta=(0.5*np.pi+gsObject.positionAngleRadians)*galsim.radians)
  File "/global/common/software/lsst/cori-haswell-gcc/stack/w.2018.03_py3/stack/miniconda3-4.3.21-10a4fa6/Linux64/galsim/1.5.1.lsst1/lib/python/galsim/gsobject.py", line 829, in shear
    shear = galsim.Shear(**kwargs)
  File "/global/common/software/lsst/cori-haswell-gcc/stack/w.2018.03_py3/stack/miniconda3-4.3.21-10a4fa6/Linux64/galsim/1.5.1.lsst1/lib/python/galsim/shear.py", line 188, in __init__
    raise ValueError("Cannot use requested axis ratio of %f!"%q)
ValueError: Cannot use requested axis ratio of 1.846340!
```
and 
```
  File "/global/cscratch1/sd/jchiang8/imSim_prs/imSim/bin/imsim.py", line 171, in <module>                                                                                  
    main()                                                                            
  File "/global/cscratch1/sd/jchiang8/imSim_prs/imSim/bin/imsim.py", line 154, in main
    gs_interpreter.drawObject(gs_obj)                                                 
  File "/global/cscratch1/sd/jchiang8/imSim_prs/sims_GalSimInterface/python/lsst/sims/GalSimInterface/galSimInterpreter.py", line 294, in drawObject                        
    centeredObj = self.findAllDetectors(gsObject)                                     
  File "/global/cscratch1/sd/jchiang8/imSim_prs/sims_GalSimInterface/python/lsst/sims/GalSimInterface/galSimInterpreter.py", line 167, in findAllDetectors                  
    centeredObj = self.createCenteredObject(gsObject)                                 
  File "/global/cscratch1/sd/jchiang8/imSim_prs/sims_GalSimInterface/python/lsst/sims/GalSimInterface/galSimInterpreter.py", line 437, in createCenteredObject              
    centeredObj = self.drawSersic(gsObject)                                           
  File "/global/cscratch1/sd/jchiang8/imSim_prs/sims_GalSimInterface/python/lsst/sims/GalSimInterface/galSimInterpreter.py", line 379, in drawSersic                        
    centeredObj = centeredObj.lens(gsObject.g1, gsObject.g2, gsObject.mu)             
  File "/global/common/software/lsst/cori-haswell-gcc/stack/w.2018.03_py3/stack/miniconda3-4.3.21-10a4fa6/Linux64/galsim/1.5.1.lsst1/lib/python/galsim/gsobject.py", line 865, in lens                                                                            
    return self.shear(g1=g1, g2=g2).magnify(mu)                                       
  File "/global/common/software/lsst/cori-haswell-gcc/stack/w.2018.03_py3/stack/miniconda3-4.3.21-10a4fa6/Linux64/galsim/1.5.1.lsst1/lib/python/galsim/gsobject.py", line 829, in shear                                                                           
    shear = galsim.Shear(**kwargs)                                                    
  File "/global/common/software/lsst/cori-haswell-gcc/stack/w.2018.03_py3/stack/miniconda3-4.3.21-10a4fa6/Linux64/galsim/1.5.1.lsst1/lib/python/galsim/shear.py", line 116, in __init__                                                                           
    raise ValueError("Requested shear exceeds 1: %f"%abs(self._g))                    
ValueError: Requested shear exceeds 1: 2.629349                       
```
using either `tests/tiny_instcat.txt` or `tests/data/phosim_combined.txt`.  Printing out the `uniqueId` values and looking at the corresponding entries reveals nothing unusual in the entries, however the reduced shears don't match the value that galsim is complaining about. 
* One of the reasons the current code in master usea the `panda.read_csv` function to ingest the instance catalogs was that the original imSim code read in the catalogs line by line as ascii, and for an instance catalog with ~80k objects, that processing took ~30mins vs a few minutes using the `.read_csv` method.   If timing tests reveal similar performance for the new instance catalog reading code, we may need to revisit this.

danielsf:
My mistake.

All of the lensing shear parameters are zero on fatboy, so I just assigned random values in phosim_galaxies.txt.  Apparently, the values I assigned were non-physical.  I will replace the test catalog with something that runs.  Do you know where the shear values in tiny_instcat.txt came from?

Regarding the pandas.read_csv: I misremembered our conversation.  I thought we had decided it was safe to not rely on pandas. I can work on sticking it back in.

jchiang87:
Ok, thanks, that explains why the instance catalogs in the data subdir don't work, but `tests/tiny_instcat.py` did work with `imsim.py` now it fails with
```
File "/global/common/software/lsst/cori-haswell-gcc/stack/w.2018.03_py3/stack/miniconda3-4.3.21-10a4fa6/Linux64/galsim/1.5.1.lsst1/lib/python/galsim/gsobject.py", line 829, in shear
    shear = galsim.Shear(**kwargs)
  File "/global/common/software/lsst/cori-haswell-gcc/stack/w.2018.03_py3/stack/miniconda3-4.3.21-10a4fa6/Linux64/galsim/1.5.1.lsst1/lib/python/galsim/shear.py", line 188, in __init__
    raise ValueError("Cannot use requested axis ratio of %f!"%q)
ValueError: Cannot use requested axis ratio of 1.846340!
```
There must be some other change somewhere.  I'd hold off on reverting the pandas usage until we can see what the timing is for the new code on an instance catalog with ~100k objects.

danielsf:
The last line of tiny_instcat.txt transposed the semi-major and semi-minor axes.  Maybe the old API automatically corrected such mistakes.   The code in this PR does not.  GalSim was getting upset that the semi-minor axis appeared to be larger than the semi-major axis.

I have fixed that in tiny_instcat.txt.

I have also updated the catalogs in tests/data so that they run as well.  Some warnings are raised about galaxies that are too faint to produce any photons, but imSim runs without crashing

jchiang87:
So those values of major and minor axes were intended in tiny_instcat.txt.  The old code would check all of the entries in the instance catalogs and remove items that would cause galsim to crash.  This was necessary during DC1 because invalid entries would kill jobs that had run for a while, invalidating all of the previous calculations.  Of course, just deleting the offending entry after such a crash wasn't a guarantee that the job wouldn't crash on a later bad entry.  In order to avoid this, even if we believe that the instance catalogs will always be correct, I think we have to have some code that checks for invalid entries and removes them before starting the simulations.

danielsf:
I have several hours on planes tomorrow.  I will try to get the InstanceCatalog reading code back into a state where it had all of the features of the pandas-based version.  I will push some new commits tomorrow evening.

jchiang87:
Here are a couple figures comparing the performance and output of this branch (as of  4d03a91beb) vs master (at 28a4dab9a00).

In the following, I'm using an instance catalog with ~78k objects, visit 197356, generated using the Run1.1 inputs, with AGN and sprinkled sources disabled.

imsim_performance
This plot shows the RSS memory usage vs cpu time.  Data points (obtained via `psutil`) on the running process were written out just prior to calling `.drawObject` in both cases.  In master, the instance catalog was read in and parsed within the first 25 seconds; in the refactor branch, it takes about ~18 minutes to read in the instance catalog.   In the refactor branch, `GalSimObject` creation also occurs as the instance catalog entries are read in, whereas in master, the `GalSimObjects` are created in the loop over `.drawObject`.  The overall difference in cpu time between the two runs is almost exactly 10 minutes, the bulk of which is devoted to reading the instance catalog data in the refactor branch.   Happily, the memory footprints for both versions are ~0.75GB.  Note that these are both explicitly handling only a single sensor, i.e., `--sensor "R:2,2 S:1,1"` was used.

Both runs show a ~3 min gap after the first ~300 objects have been drawn.   I don't yet understand the reason for that.

----

imsim_refactor_rotation
This is a comparison of the images produced by the master and refactor branches.  (The refactor branch image has two prominent cosmic rays on the rhs of the image).  There appears to be a net rotation about the center of the image, and since this is `R:2,2 S:1,1`, the image center is also the center of the full fov.  It would be good to understand this shift.

danielsf:
There was a gross inefficiency in how we calculated the dust extinction parameters in the old `u/danielsf/refactor` commit you compared against, @jchiang.  We were re-instantiating the `a` and `b` arrays for each GalSimObject.  I just pushed a commit that fixes that.  That *might* get rid of the ten minute slowdown between this branch and `master`.  I'm not sure.

Regarding the rotation: I will look into it.  I suspect that we were not as clever as we thought we were undoing the PhoSim coordinate rotations on `master`.  The new unit tests on this branch verify that the pupil coordinates of the `GalSimCelestialObject`s are consistent with what we would expect naively applying the CatSim coordinate transformation routines to the truth catalogs.  That test is here

https://github.com/LSSTDESC/imSim/blob/u/danielsf/refactor/tests/test_instcat_parser.py#L145

I will try to use pandas to speed up the InstanceCatalog reading.  (In all honesty, I suspect that the dust inefficiency I mentioned above was also present in `master`).

danielsf:
This branch

https://github.com/LSSTDESC/imSim/tree/u/danielsf/refactor_pandas

uses pandas to read in the InstanceCatalog.  It does not appear to me to be any faster than the current PR.

jchiang87:
> Looking at imsim_truth.py, I'm not sure that we need it any more.

The original reason for `imsim_truth.py` was to provide a means of getting the top of the atmosphere magnitudes from instance catalogs without having to re-query the UW db.  We used that code to produce the truth tables for DC1 and were planning to use it for the Run1.1x data.   If there is another way to get those magnitudes, e.g., via some other existing CatSim code, that's fine, but we should document that somewhere.

danielsf:
@jchiang87 I added some commits last night here and in the sims_GalSimInterface PR, trying to re-enable your checkpointing scheme.  Do you mind looking at them quickly to make sure they are sensible?

bin.src/imsim.py
```diff
@@ -148,10 +145,12 @@ def main():
 
     if arguments.sensor is not None:
         gs_objects_to_draw = gs_object_dict[arguments.sensor]
-    else
+    else:
```
danielsf: Sorry about all that.  Didn't realize how sloppy those commits were...

jchiang87:
@danielsf I went ahead and fixed the checkpointing (and numRows) code.   From my testing, it looks like the checkpointing is working as before.

@cwwalter I think everything here is ready to merge.

If no one has any objections, I'd like to merge at Noon PT, March 2 (tomorrow).

danielsf:
We are in a bit of a weird state:

Yesterday, DM merged changes to the afwCameraGeom API that broke the dependent sims code.  I have fixes under review that I can merge tomorrow, however, the DM changes have not made it into a weekly, yet, so I can't just issue a new version of lsst_sims, as there will be no weekly to build against.

Ordinarily, I would just go a head and merge everything (the DM-dependent fixes and this refactor, which depends on different changes to lsst_sims repos), accepting that imSim is only going to be able to run against a `master` stack for the next few days.  I suspect, however, that people are actively working on imSim and I don't want to accidentally throw a wrench in anything.

Are there any objections to imSim being in a `master`-only state until Monday, March 5 (with the added caveat that I am traveling to a project meeting on Monday, so the new version of lsst_sims won't be ready until Monday afternoon)

jchiang87:
Which sims packages would be affected by the afwCameraGeom changes?

I'm still using sims 2.5.0 + master of sims_catUtils and the `u/danielsf/clean_gsobj` branch of `sims_GalSimInterface`.  If we merged this and continued to use that configuration, would things still work once the afwCameraGeom-related changes are incorporated?

Others should chime in with what they are doing, but I think it would be better overall to merge these changes in first before merging in the other ongoing work.  Some work will be easier with these changes (e.g., adding the treering info), and other work will have to deal with merge conflicts no matter when we merge this.

danielsf:
The afwCameraGeom change affected sims_coordUtils and sims_GalSimInterface

I can arrange a merge sequence such that we will have a stack that works with your configuration.  It sounds like that is what you want (I'm having a hard time parsing what is meant by "ongoing work" in your last post).

jchiang87:
'ongoing work' == the 5 open PRs in this repo

danielsf:
There is now a 2.5.1.sims version of lsst_sims that incorporates the changes needed to support this PR of imSim.  People running the old version of imSim can still install 2.5.0.sims using `eups dsitrib install lsst_sims -t sims_2_5_0`.   2.5.1.sims is still built against w.2018.03.

Issue #96, opened 2018-02-26
enable reading of gzipped instance catalog components and includeobj entries


jchiang87:
This will be more consistent with phosim behavior and will make the code to write instance catalogs for both imsim and phosim simpler.

cwwalter:
How do you see this interacting with #71?

jchiang87:
I think getting imsim able to read the same format files as phosim wrt gzipped files and `includeobj` is a more pressing concern than the work in #71.   Adding the desired functionality here will result in much less work in instance catalog generation code.  #71 isn't as relevant.

cwwalter:
For future reference for people this issue (raised by Tom, answered by John) has an explicit description of the functionality.

https://bitbucket.org/phosim/phosim_release/issues/6/problem-interpreting-gzipped-instance

PR #99, opened 2018-03-23
handle gzipped instance catalogs and includeobj entries

jchiang87:
This addresses #96 .  I think the unit tests are comprehensive; nonetheless, I've tried it on phosim instance catalogs generated with the code used for Run1.1p, and everything seems fine.


tests/test_fopen.py
```diff
@@ -0,0 +1,51 @@
...
+            os.rmdir(self.test_dir)
+        except OSError:
+            pass
+
+    def test_fopen(self):
+        "Test the fopen function."
+        with desc.imsim.fopen(self.fopen_test_file, mode='rt') as input_:
+            for i, line in enumerate(input_):
+                self.assertEqual(line.strip(), self.lines[i])
+
```
danielsf: The paranoiac in my thinks you should check that `i==4` at this point, just to make sure that all of the lines were yielded.

jchiang87: Excellent point....I've updated the test to check this.

jchiang87:
Unless I hear otherwise, I plan to merge this at 5pm PT today.

PR #105, opened 2018-04-04
Use visit number and detector name to set random seed for generating cosmic rays

jchiang87:
This will ensure that CRs are the same for a given visit and also vary from sensor to sensor.

### DIFF
--- a/python/desc/imsim/cosmic_rays.py
+++ b/python/desc/imsim/cosmic_rays.py
@@ -5,6 +5,7 @@ testing.
 """
 from __future__ import print_function
 from collections import namedtuple, defaultdict
+import hashlib
 import numpy as np
 import numpy.random as random
 import astropy.io.fits as fits
@@ -37,6 +38,7 @@ class CosmicRays(list):
         self.num_pix = 0
         self.exptime = 0
         self.ccd_rate = 0
+        self.rng = random     # Use numpy.random module by default.
 
     def paint(self, image_array, exptime=30., num_crs=None):
         """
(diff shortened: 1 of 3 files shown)


python/desc/imsim/cosmic_rays.py
```diff
@@ -88,12 +89,12 @@ def paint_cr(self, image_array, index=None, pixel=None):
         numpy.array: The input image array with the CR added.
         """
         if index is None:
-            cr = random.choice(self)
+            cr = self.rng.choice(self)
```
cwwalter: I'm sure this a python feature I don't appreciate, but is there a reason this isn't

self.rng.choice(self.image_array)
 
and then self.image_array[index]

?

python/desc/imsim/cosmic_rays.py
```diff
@@ -137,6 +138,47 @@ def read_catalog(catalog_file, ccd_rate=None, extname='COSMIC_RAYS'):
...
+
+        Parameters
+        ----------
+        seed: int
+            The seed must be between 0 and 2**32 - 1
+        """
+        self.rng = np.random.RandomState(seed)
+
+    @staticmethod
+    def get_seed(visit, det_name):
```
cwwalter: Perhaps this should be generate_seed ?

get_seed sounds more like a function to retrieve the current seed.

python/desc/imsim/cosmic_rays.py
cwwalter: OK Jim explained to me that CosmicRay inherits from list so self is also a list.

Also in my comment above image_array wouldn't be the correct choice since that is where the CRs are going.

python/desc/imsim/cosmic_rays.py
jchiang87: Sure, I can change it to that.  I almost called it `make_seed`.

PR #106, opened 2018-04-07
Use sims 2.7.0 in travis-ci builds

jchiang87:
@danielsf Looks like were getting a couple of unit test failures that we weren't seeing with 2.6.0:

https://travis-ci.org/LSSTDESC/imSim/builds/363313116

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,8 @@ dist: trusty
 compiler: gcc
 
 env:
-  - OS_TYPE=centos OS_VERSION=7
+  - OS_TYPE=centos
+  - OS_VERSION=7
   - DOCKER_IMAGE=lsstdesc/stack-sims:w_2018_13-sims_2_7_0
 
 services:
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,7 @@ compiler: gcc
 
 env:
   - OS_TYPE=centos OS_VERSION=7
+  - DOCKER_IMAGE=lsstdesc/stack-sims:w_2018_13-sims_2_7_0
 
 services:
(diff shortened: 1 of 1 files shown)


danielsf:
Well, the "good" news is that I think these tests are failing because the truth catalog the test is comparing to is, in fact, wrong.  We were deprecessing the PhoSim coordinates incorrectly in version 2.6.0.  Now that we have fixed that problem, the relationship between the InstanceCatalog and the truth catalog should have changed.  I will re-generate the test data and issue a PR.

danielsf:
The broken test takes and InstanceCatalog and a truth catalog which contains both RA, Dec and position on the focal plane and verifies that ImSim preserves the relationship between those two coordinate systems.  Version 2.6.0, which was used to generate those catalogs, had an incorrect relationship between the coordinate systems.  Now that we have fixed it, ImSim and the test catalogs disagree on the relationship between RA, Dec and position on the focal plane.  However, it is the input to the test which is wrong, not ImSim.

cwwalter:
Thanks.  I was actually asking to make sure that the instance catalogs being used for 1.2p are OK now since I thought we used 2.6.0 for that but wasn't sure..

danielsf:
Actually, because of a bug in how we were converting ICRS to PhoSim coordinates, Run 1.2p is being run with 2.7.0.sims, which I have validated against some small PhoSim simulations.

danielsf:
This breakage should be fixed by

https://github.com/LSSTDESC/imSim/pull/107

jchiang87:
I'll merge this along with #107 at Noon today unless I hear otherwise.

PR #110, opened 2018-04-21
add bleed trail implementation

jchiang87:


cwwalter:
Awesome! Will look.

Can you comment on two things to help me understand?  You called the afw detect function to get footprints. So if there is a blob of saturated charge in a footprint does all the charge in that footprint go down one column?

Also, should we stop the charge at the midline? Or just let it pass over. It looks like it is just traveling over now.

jchiang87:
> Can you comment on two things to help me understand? You called the afw detect function to get footprints. So if there is a blob of saturated charge in a footprint does all the charge in that footprint go down one column?

Each footprint is composed of "spans" which are ranges of contiguous pixels above threshold along a channel.  The code is finding all the spans and thus all of the channels with pixels above full well, and so a object with several such channels will bleed along each one.  Here's an example:
bleeding_star

> Also, should we stop the charge at the midline? Or just let it pass over. It looks like it is just traveling over now.

I wasn't sure what to do about that, since it wasn't clear to me how the charge should behave once it got to any midline bleed stop.  Also, I thought the behavior for ITL sensors was different than for e2v devices.  If we want to treat a midline stop like an edge and just force all the charge to go the other direction once the channel is full to the midline, that would be any easy change to implement.

cwwalter:
> Each footprint is composed of "spans" which are ranges of contiguous pixels above threshold along a channel. The code is finding all the spans and thus all of the channels with pixels above full well, and so a object with several such channels will bleed along each one. Here's an example:

OK I understand; thanks.

> I wasn't sure what to do about that, since it wasn't clear to me how the charge should behave once it got to any midline bleed stop. Also, I thought the behavior for ITL sensors was different than for e2v devices. If we want to treat a midline stop like an edge and just force all the charge to go the other direction once the channel is full to the midline, that would be any easy change to implement.

So it stops on e2v since there is the midline bloom stop (I haven't really been able to understand what happens to the charge, does it flow back etc) but we can just choose to have it go back.  Craig did some tests on the ITL sensors at Davis since the stop isn't there. He found that it took a while but then eventually it did flow over but that there was something that stopped it for a while and it sort of spreads out.   I think we still need more work to really understand this.

I'm interested in other people's comments (especially Craig). I'm sort of inclined to stop it in both cases for now but am not sure.

cwwalter:
I'm tagging @craiglagegit to get his input.  I've been thinking more about it and I have this proposal:

-If we are going to choose one right now, I think we should have then all stop at the midline.  Our "mixed focal" plane contains e2v after all, and we know those stop.  Based on Craig's recent work it seems we see at least a 'pause' for the ITL.

- For a more complete solution: I think we will eventually want an option to choose either e2v or ITL behavior.  For now it could be either to stop at a midline bloom stop or not.  One thing that might add a nice bit of complexity/realism to the simulations would be to tie this to our 'Pseudo mixed focal plane'.  Do we have a way of knowing in imSim which sensors have the high and low level of tree rings (or is it all in GalSim)? We could match the behavior to that.

craiglagegit:
@cwwalter For the tree rings, the data I got from BNL had 5 ITL sensors - 3 with "low" tree rings and 2 with "high" tree rings, and 1 E2V sensor with low tree rings.  Since there is no statistical difference between 3/5 low and 0/1 low, I set up the focal plane where 40% of the sensors have high tree rings and 60% have low tree rings (10X lower), regardless of vendor.  So we can't use the tree rings to ID the vendor.  Of course this can all be easily changed.  I could have it vary by vendor, but then I would need a map of which rafts are ITL and which are E2V.  For the blooming, recall the ITL blooming didn't "pause" for long at the midline, so I think it'd be a mistake to have it stop at the midline.  I'd like to see data on just how effective the E2V midline bloom stop is.  We are still working on getting and E2V sensor here in Davis to test, but I don't have it yet.

cwwalter:
@craiglagegit Thanks.

Just so I understand.  You said:

> For the blooming, recall the ITL blooming didn't "pause" for long at the midline, so I think it'd be a mistake to have it stop at the midline.

I was looking at your talk at the March 20 SAWG meeting where you showed:

screen shot 2018-04-24 at 9 43 15 am

and it seemed there was something like a minute before it started to flow over.  But was this really a statement about how long it took to fill the well? In the picture it sort of looks like it starts to puddle up before it flows over.  Are you saying in the end it doesn't look any different?

cwwalter:
> I added a midline stop as an option and made it the default. The code is written so that the information whether or not a given detector has a midline stop should be obtained from the GalSimDetector object, i.e., where it would be stored alongside the tree ring info.

OK; Great.

python/desc/imsim/bleed_trails.py
```diff
@@ -0,0 +1,175 @@
...
+        if midline_stop:
+            xmid = imarr.shape[1]//2
+            imarr[ypix, :xmid] = bleed_channel(imarr[ypix, :xmid], full_well)
+            imarr[ypix, xmid:] = bleed_channel(imarr[ypix, xmid:], full_well)
+        else:
+            imarr[ypix, :] = bleed_channel(imarr[ypix, :], full_well)
+    return eimage
+
+
+def find_channels_with_saturation(eimage, full_well):
```
cwwalter: I find "channel" a bit ambiguous.  What does it mean, amp channel, pixel, etc.  Would 'column' be better?so: find_columns_with_saturation etc.

python/desc/imsim/bleed_trails.py
```diff
@@ -0,0 +1,175 @@
...
+
+    numpy.array of pixel values: The channel of pixel data with
+       bleeding applied.  This is a new object, i.e., the input
+       numpy.array is unaltered.
+    """
+    # Construct an afw_image.ImageF to contain the channel data
+    # to enable the use of afw_detect to find the above full-well
+    # footprints.
+    nx = len(channel)
+    image = afw_image.ImageF(nx, 1)
```
cwwalter: Is this guaranteed to always be initialized to zero so when you add the channel data to it, it will be OK?

python/desc/imsim/bleed_trails.py
```diff
@@ -0,0 +1,175 @@
...
+    return imarr[0, :]
+
+
+class BleedCharge:
+    "Class to manage charge redistribution along a channel."
+    def __init__(self, imarr, excess_charge, full_well):
+        """
+        Parameters
+        ----------
+        imarr: numpy.array
```
cwwalter: "imarr" here is used for a 1D column.  But above it is used for a section of a 2D amplifier.

python/desc/imsim/skyModel.py
```diff
@@ -135,7 +135,7 @@ def __init__(self, obs_metadata, seed=None, addNoise=True, addBackground=True):
             self.randomNumbers = galsim.UniformDeviate(seed)
 
     def addNoiseAndBackground(self, image, bandpass=None, m5=None,
-                              FWHMeff=None, photParams=None, chipName=None):
+                              FWHMeff=None, photParams=None, detector=None):
```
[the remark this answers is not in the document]
jchiang87: This is needed to get the test code to work with the current master of `sims_GalSimInterface`, which is what the `treerings` branch also requires.

tests/test_skyModel.py
```diff
@@ -4,6 +4,7 @@
 from __future__ import absolute_import
 import os
 import unittest
+from collections import namedtuple
```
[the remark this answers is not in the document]
jchiang87: Those changes are already in the `treerings` branch.   The changes performed here are what's needed to get the test code to be interface-compatible with the master of `sims_GalSimInterface`.

python/desc/imsim/bleed_trails.py
jchiang87: I wanted to avoid `column` because it implies being parallel to the y-axis, whereas for LSST eimages, the channels are along the x-axis in the Camera Coordinate System.  Also, the `lsst.afw.detection` code has the relevant dimensions of its `span`s indexed in x.  Finally, "channel" seems to be the correct technical term for this feature of CCDs, e.g., see the Janesick book.

python/desc/imsim/bleed_trails.py
jchiang87: Yes, that's the default value:  https://github.com/lsst/afw/blob/master/include/lsst/afw/image/Image.h#L95

python/desc/imsim/bleed_trails.py
jchiang87: In order to use the `afw_detect` code, it needs to operate on a 2D image, that's why `imarr` has this structure.

jchiang87:
I'll merge this at Noon PT today unless I hear otherwise.

Issue #32, opened 2016-11-15
Add a centroid file capability


cwwalter:
PhoSim has a useful feature described as:

'''
Centroid files contain the true numbers of photons that were detector for every source, and the average x and y position. This is useful for crowded fields where the photons from each source can be tagged internally in the simulator.
'''

This may actually be better handled as a GalSim feature as it could be part of the raytracing in either/both the atmosphere and the sensor.

cwwalter:
Closed by #112 and https://github.com/lsst/sims_GalSimInterface/pull/51.

PR #112, opened 2018-05-05
Add basic centroid file capability.

cwwalter:
I am working on adding more pixel level truth to the simulation but thought it would be useful to first add basic 'centroid file' capability (requested in #32).  

My work on adding information based on what each object looks like on the sensor is currently adding a lot of CPU run time and I need more work to optimize it.  So, this initial PR uses only truth information about the objects in the centroid file.   This makes it more difficult to use this to do imSim/PhoSim validation comparisons (because the PhoSim centroid values are average X and Y values on the sensor) but it is better for comparing catalog truth with the sensor since the true values are not affected by being near the edge of the sensor etc. Eventually we may want both. 

This code will make centroid files corresponding to each fits file in the output directory.  For example:

```
wasabi:fits % ls
centroid_197356_R22_S01_r.txt	lsst_e_197356_R22_S01_r.fits
centroid_197356_R22_S10_r.txt	lsst_e_197356_R22_S10_r.fits
centroid_197356_R22_S11_r.txt	lsst_e_197356_R22_S11_r.fits
centroid_197356_R22_S12_r.txt	lsst_e_197356_R22_S12_r.fits
centroid_197356_R22_S21_r.txt	lsst_e_197356_R22_S21_r.fits
```

With a small piece of code this can be converted into a ds9 region file for overlay.  This pandas code example puts a circle at each true object position with the radius scaled by the log of the number of photons for one of the centroid files:

```
import math
import numpy as np
import pandas as pd

centroids = pd.read_csv('fits/centroid_197356_R22_S11_r.txt', delim_whitespace=True, comment="#")

with open('region_file.reg','w') as outfile:
    outfile.write('physical\n')
    centroids.query('Photons>0').to_string(outfile, columns=['xpos','ypos','Photons'], 
                                           header=False, index=False,
                                           formatters=['circle {:10.3f}'.format,
                                                       '{:10.3f}'.format,
                                                       lambda x: '{:10.3f}'.format(math.log(x))])
```

Loading this regions file after loading the fits file results in:

screen shot 2018-05-06 at 12 12 42 am

To my eye, there is a small shift between the circle centers and the objects.  I'm not sure if this is a plotting issue, a misunderstanding on my part of the variables I used, or some small bug in the code.

The imSim code only sets the options to turn this on and sets the file prefix in the config file.  The code to open, close and write to the files is in sims_GalSimInterface. I will open a separate PR there.

Currently, this code only is in the non-sensor version of the GalSimInterpreter class.  I would like some feedback from @jchiang87, @rmjarvis and @danielsf before I add much more to make  sure this approach seems reasonable to people.

I started by making this look like a PhoSim centroid file since I thought this might make things easier on the people running the pipeline but in principle we could also add more information into the file.

### DIFF
--- a/python/desc/imsim/ImageSimulator.py
+++ b/python/desc/imsim/ImageSimulator.py
@@ -34,7 +34,7 @@ class ImageSimulator:
     Class to manage the parallel simulation of sensors using the
     multiprocessing module.
     """
-    def __init__(self, instcat, psf, numRows=None, config=None, seed=267,
+    def __init__(self, instcat, create_centroid_file, psf, numRows=None, config=None, seed=267,
                  outdir='fits', sensor_list=None, apply_sensor_model=True,
                  file_id=None, log_level='WARN'):
         """
@@ -69,6 +69,7 @@ class ImageSimulator:
             Logging level ('DEBUG', 'INFO', 'WARN', 'ERROR', 'CRITICAL').
         """
         self.config = read_config(config)
+        self.create_centroid_file = create_centroid_file
         self.psf = psf
         self.outdir = outdir
         self.obs_md, self.phot_params, sources \
@@ -124,6 +125,10 @@ class ImageSimulator:
                     .restore_checkpoint(self.camera_wrapper,
                                         self.phot_params,
                                         self.obs_md)
+
+            if self.create_centroid_file is True:
+                self.gs_interpreter.centroid_base_name = self.outdir + '/' \
+                    + self.config['persistence']['centroid_prefix']
 
     @staticmethod
     def checkpoint_file(file_id, det_name):
@@ -279,6 +284,9 @@ class SimulateSensor:
         obsHistID = str(image_simulator.obs_md.OpsimMetaData['obshistID'])
         gs_interpreter.writeImages(nameRoot=os.path.join(outdir, prefix)
                                    + obsHistID)
+
+        gs_interpreter.close_centroid_files()
+
 
 #        # The image for the sensor-visit has been drawn, so delete the
 #        # checkpoint file.
--- a/python/desc/imsim/ImageSimulator.py
+++ b/python/desc/imsim/ImageSimulator.py
@@ -285,6 +285,8 @@ class SimulateSensor:
         gs_interpreter.writeImages(nameRoot=os.path.join(outdir, prefix)
                                    + obsHistID)
 
+        # Write out the centroid files if they were made.
+        gs_interpreter.write_centroid_files()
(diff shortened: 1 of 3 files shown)


danielsf:
Regarding the offset between the green circles and the objects in the image above:

Was this with or without the sensor model, @cwwalter ?  The pixel positions you are printing to the centroid files are calculated before the sensor model is applied.  That might have something to do with it.

cwwalter:
> Was this with or without the sensor model, @cwwalter ? The pixel positions you are printing to the centroid files are calculated before the sensor model is applied. That might have something to do with it.

This was without the sensor model (Just calling the drawObject version in GalSimInterpreter).  I also ran with the sensor model turned on which uses GalSimSiliconInterpeter. There I saw what looked to my eye like a smaller shift if it was still there in another direction.

python/desc/imsim/ImageSimulator.py
```diff
@@ -34,7 +34,7 @@ class ImageSimulator:
     Class to manage the parallel simulation of sensors using the
     multiprocessing module.
     """
-    def __init__(self, instcat, psf, numRows=None, config=None, seed=267,
+    def __init__(self, instcat, create_centroid_file, psf, numRows=None, config=None, seed=267,
```
jchiang87: I would prefer to make `create_centroid_file` a keyword argument with a default value of `False` rather than make it a required argument, so that existing scripts won't break.  I would also add it towards the end of the argument list, say after `apply_sensor_model`.

python/desc/imsim/ImageSimulator.py
```diff
@@ -125,6 +126,10 @@ def _make_gs_interpreters(self, seed, sensor_list, file_id):
                                         self.phot_params,
                                         self.obs_md)
 
+            if self.create_centroid_file is True:
+                self.gs_interpreters[det_name].centroid_base_name = self.outdir \
```
jchiang87: It mostly doesn't matter, but it's probably better to use `os.path.join` for concatenating path components.

python/desc/imsim/ImageSimulator.py
```diff
@@ -125,6 +126,10 @@ def _make_gs_interpreters(self, seed, sensor_list, file_id):
                                         self.phot_params,
                                         self.obs_md)
 
+            if self.create_centroid_file is True:
```
jchiang87: `if self.create_centroid_file is True:` -> `if self.create_centroid_file:`

cwwalter:
Requested changes are done in this repo.

PR #113, opened 2018-05-10
calculate FWHMgeom, FWHMeff rather than reading them from the instance catalog

jchiang87:
This will allow imsim to read exactly the same instance catalogs as phosim.


jchiang87:
Thanks, Scott, Chris.   Since these numbers are just read from the opsim db files, I assume that the code to compute these numbers are in OpSim somewhere.  I'll merge this in a couple of hours.

Issue #85, opened 2017-12-14
modify imsim_truth.py to also object magnitudes with Galactic (MW) reddening removed


jchiang87:
This is a request from the LSS group for the truth tables of the galaxy properties

cwwalter:
Is this a DC2 era request from the LSS group?

PR #98, opened 2018-03-23
code to enable sensor simulation in parallel using the multiprocessing module

jchiang87:

PR #114, opened 2018-05-16
Multiprocessing branch with updates to object handling

jchiang87:
This PR supersedes #98 .  In addition to code to enable the use of the `multiprocessing` module, this PR includes pre-trimming of the instance catalogs, deferral of the `GalSimCelestialObject` and `Sed` creation to the sub-processes to save memory and initial set up time, and a re-written `imsim.py` application.

There are a few small items to add for running on KNL:
- [x] Checking for already completed sensor visits on restart.
- [x] Separate log files for each sub-process.
- [x] Different debug levels for imsim vs sims code.


python/desc/imsim/ImageSimulator.py
```diff
@@ -0,0 +1,277 @@
...
+        if processes == 1:
+            # Don't need multiprocessing, so just run serially.
+            for det_name in self.gs_interpreters:
+                if os.path.exists(self.eimage_file(det_name)):
+                    continue
+                simulate_sensor = SimulateSensor(det_name, self.log_level)
+                simulate_sensor(self.gs_obj_dict[det_name])
+        else:
+            # Use multiprocessing.
+            pool = multiprocessing.Pool(processes=processes)
```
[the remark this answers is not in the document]
jchiang87: I think it's not a problem if more cores than sensor-visits are requested.  Un-occupied subprocesses in the pool do show up in `top`, but they don't appear to take up any resources, i.e., neither cputime or memory are used.

tests/test_instcat_parser.py
```diff
@@ -43,7 +43,7 @@ def setUp(self):
         self.extra_commands = 'instcat_extra.txt'
         with open(self.phosim_file, 'r') as input_file:
             with open(self.extra_commands, 'w') as output:
-                for line in input_file.readlines()[:23]:
+                for line in input_file.readlines()[:22]:
```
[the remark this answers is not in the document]
jchiang87: I think for unit tests, it's actually ok to hard-wire in specific numbers, especially for test data that's prepared for the test.  I didn't write this test code, though it was probably based on an earlier test that I did write.   In any case, I think the idea was to have the extra command appear in the `object` entries and have the code check for errant commands that may have been added there.

python/desc/imsim/skyModel.py
```diff
@@ -289,18 +289,18 @@ def addNoiseAndBackground(self, image, bandpass=None, m5=None,
                                       treering_center=detector.tree_rings.center,
                                       transpose=True)
 
-        # Loop over amplifiers to save memory when storing the 36
+        # Loop over 1/2 amplifiers to save memory when storing the 36
```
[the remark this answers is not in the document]
jchiang87: ok.  I'll add this and will update the docstring for the `FastSiliconSkyModel` as you suggest below.

cwwalter:
It would be good to update the description of --sensor (make it --sensors ?) to make it clear a list of sensors is allowed.

python/desc/imsim/instcat_tools.py
```diff
@@ -0,0 +1,176 @@
+"""
```
cwwalter: Does this code replace imsim_truth.py in master?  Also does it close #85?

data/default_imsim_configs
```diff
@@ -13,3 +13,6 @@ catalog = default
 
 [checkpointing]
 nobj = 1000
+
+[wl_params]
+gamma2_sign = -1
```
cwwalter: So for this and the related gamma changes... I think this is right but probably it is good for @rmjarvis to look at it.

My understanding: the instance catalogs are now writing out "the phoSim convention". The entry is labeled_such (but nothing has changed from what it was doing before).  So if that is the case, when we read it in, the - sign is needed because the shear is applied by GalSim. The GalSim algorithm expects the "normal/DES" convention while the PhoSim shear algorithm doesn't.  So, this should give us the same physical orientation on the sensor. Correct?

tests/test_instcat_parser.py
```diff
@@ -289,7 +289,7 @@ def test_object_extraction_galaxies(self):
                                    truth_data['minorAxis'][i_obj], 13)
             self.assertAlmostEqual(gs_obj.majorAxisRadians,
                                    truth_data['majorAxis'][i_obj], 13)
-            self.assertAlmostEqual(gs_obj.positionAngleRadians,
+            self.assertAlmostEqual(2.*np.pi - gs_obj.positionAngleRadians,
```
cwwalter: Angles are defined counter-clockwise?

python/desc/imsim/instcat_tools.py
jchiang87: yes, I suppose it does on both counts.  I'll remove the imsim_truth.py code.

tests/test_instcat_parser.py
jchiang87: I don't think it matters.  The upshot of this change is that the PA sign is flipped and 2\*pi is added to get it back in the (0, 2\*pi) range.

jchiang87:
I'll merge this branch at 2pm PT today, unless I hear otherwise.

PR #123, opened 2018-07-22
Update code that writes amplifier files so that they work with obs_lsstCam

jchiang87:


jchiang87:
This is superseded by #136, but will keep the branch around until `obs_lsstCam` is in `lsst_distrib` so that the updated `write_amp_image_file.py` can be used.

PR #124, opened 2018-07-27
improve imsim.py --help and update README.md

jchiang87:


cwwalter:
Should this get merged so it is in the version that Adrian uses?

jchiang87:
I'll merge this tomorrow at 9am PT unless I hear otherwise.

PR #126, opened 2018-07-30
treering flats

### LINKED ISSUE #120: Add a flat-production script
We should add a imSim command line option, or just a script that will solely call the sensor model with a flat background as the sky. We will need to specify an SED and the magnitude/".  

Should we produce monochromatic flats?

This should run extremely quickly.

hsnee:
this should resolve #120


jchiang87:
Thanks for this Husni.  I'll do a more complete review in the next day or so, but I wanted to suggest now that it would be better to make a standalone `make_flats.py` script rather than adding this functionality to `imsim.py`.  So I would request that you revert the changes to `imsim.py` and add a `make_flats.py` script to the `bin.src` folder.

cwwalter:
This was completed in #132.

PR #127, opened 2018-07-31
Fix scipy.fsolve failure when computing r0_500

jmeyers314:
Fixes a bug where newer versions of scipy.fsolve (sensibly) require the output of the residual function to have the same dimensions as the input.  We were previously producing a vector residual from a scalar input due to broadcasting the vector `L0` parameter.  This PR scalarizes the input `L0`, and also plays a bit more defensively against possible vector inputs that may, for instance, be employed by fsolve itself.

### DIFF
--- a/python/desc/imsim/atmPSF.py
+++ b/python/desc/imsim/atmPSF.py
@@ -81,11 +81,14 @@ class AtmosphericPSF(PSFbase):
     @staticmethod
     def _seeing_resid(r0_500, wavelength, L0, target):
         """Residual function to use with `_r0_500` below."""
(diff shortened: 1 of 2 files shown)


jchiang87:
I'll merge this tomorrow at 9am PT, unless I hear otherwise.

cwwalter:
Ah, forgot to come back to this.  Thanks Jim.  Is the numpy version bumped in the latest release?

jchiang87:
Yes, it's at `numpy 1.14.5` in `lsst_sims 2.10.0`

PR #129, opened 2018-08-07
file handling improvements

jchiang87:
This implements two feature requests from @adrianpope for running on Theta-KNL:
- compression of eimage files
- deleting checkpoint files after eimage files have been written

### DIFF
--- a/data/default_imsim_configs
+++ b/data/default_imsim_configs
@@ -6,6 +6,7 @@ full_well = 1e5
 
 [persistence]
 eimage_prefix = lsst_e_
+eimage_compress = True
 centroid_prefix = centroid_
 
 [cosmic_rays]
--- a/data/default_imsim_configs
+++ b/data/default_imsim_configs
@@ -15,6 +15,7 @@ catalog = default
 
 [checkpointing]
 nobj = 200
+cleanup = True
 
 [wl_params]
 gamma2_sign = -1
--- a/python/desc/imsim/ImageSimulator.py
+++ b/python/desc/imsim/ImageSimulator.py
@@ -296,9 +296,12 @@ class SimulateSensor:
         # Write out the centroid files if they were made.
         gs_interpreter.write_centroid_files()
 
-#        # The image for the sensor-visit has been drawn, so delete the
-#        # checkpoint file.
-#        os.remove(gs_interpreter.checkpoint_file)
+        # The image for the sensor-visit has been drawn, so delete any
+        # existing checkpoint file if the config says to do so.
+        if (gs_interpreter.checkpoint_file is not None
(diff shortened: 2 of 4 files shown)


python/desc/imsim/ImageSimulator.py
```diff
@@ -284,16 +288,44 @@ def __call__(self, gs_objects):
...
 
-#        # The image for the sensor-visit has been drawn, so delete the
-#        # checkpoint file.
-#        os.remove(gs_interpreter.checkpoint_file)
+        # The image for the sensor-visit has been drawn, so delete any
+        # existing checkpoint file if the config says to do so.
+        if (gs_interpreter.checkpoint_file is not None
+            and os.path.isfile(gs_interpreter.checkpoint_file)
+            and image_simulator.config['checkpointing']['cleanup']):
+            os.remove(gs_interpreter.checkpoint_file)
```
cwwalter: This is really adding the ability to *not* delete a checkpoint correct?  I think it was already deleting the checkpoint after the image was written before.  Is there any other feature being added here (just a bit confused since the main PR says it adds the ability per Adrian's request).

data/default_imsim_configs
```diff
@@ -14,6 +15,7 @@ catalog = default
 
 [checkpointing]
 nobj = 200
+cleanup = True
```
cwwalter: Does Adrian needs this default nobj larger now?

tests/test_imageSimulator.py
```diff
@@ -0,0 +1,46 @@
...
+Unit tests for the ImageSimulator module.
+"""
+import os
+import random
+import string
+import unittest
+import gzip
+import desc.imsim
+
+class ImageSimulatorTestCase(unittest.TestCase):
```
cwwalter: Since these tests could in principle be more generic for the ImageSimulator module (testing things like the multiprocessing) I think it would be good to add a comment in setup and teardown that gzip functionality is specifically being tested here. Alternatively, this could be made a test specifically for compression.

python/desc/imsim/ImageSimulator.py
jchiang87: The code being replaced has the call to remove the checkpoint file commented out (and in fact would crash if the checkpointing option were not enabled at runtime).  So this change does enable the checkpoint file to be deleted, *if* it exists and if the config file says it should be.

data/default_imsim_configs
jchiang87: He might, but for running on a KNL node, I think this is still basically the right number.

jchiang87:
Unless I hear otherwise, I'll go ahead and merge this at 9am PT tomorrow.

PR #132, opened 2018-08-11
flat production script

jchiang87:
Using the default parameters, this isn't as fast as one might hope.  For a single flat for `R22_S11`, it takes about 37 minutes on Cori-Haswell.   Here's a png of the image:
flat_161899_r22_s11_r

### DIFF
--- a/bin.src/make_flats.py
+++ b/bin.src/make_flats.py
@@ -1,6 +1,6 @@
 #!/usr/bin/env python
 """
-Script to create imSim flats.  This includes electorstatic sensor effects
+Script to create imSim flats.  This includes electrostatic sensor effects
 such as treerings and brighter-fatter.
 """
 import argparse
--- a/bin.src/make_flats.py
+++ b/bin.src/make_flats.py
@@ -9,7 +9,7 @@ import lsst.afw.cameraGeom as cameraGeom
 from lsst.sims.GalSimInterface import LSSTCameraWrapper, make_galsim_detector
 import desc.imsim
 
-parser = argparse.ArgumentParser("imSim flat production script.")
+parser = argparse.ArgumentParser(description="imSim flat production script.")
 parser.add_argument('instcat', type=str, help="instance catalog file")
 parser.add_argument('--sensors', type=str, default=None,
                     help="sensors to simulate")
@@ -24,15 +24,16 @@ parser.add_argument('--log_level', type=str,
 
 args = parser.parse_args()
 
+config = desc.imsim.read_config()
+
 obs_md, phot_params, _ = desc.imsim.parsePhoSimInstanceFile(args.instcat,
                                                             numRows=30)
 sensor_list = args.sensors.split('^') if args.sensors is not None \
     else args.sensors
 rng = galsim.UniformDeviate(args.seed)
-niter = int(args.counts_total/args.counter_per_iter + 0.5)
+niter = int(args.counts_total/args.counts_per_iter + 0.5)
 counts_per_iter = args.counts_total/niter
 logger = desc.imsim.get_logger(args.log_level, name='make_flats')
-config = desc.imsim.read_config()
 
 visit = obs_md.OpsimMetaData['obshistID']
 
@@ -40,12 +41,13 @@ camera_wrapper = LSSTCameraWrapper()
 
 for det in camera_wrapper.camera:
     det_name = det.getName()
-    logger.info("processing %s", det_name)
     if (det.getType() != cameraGeom.SCIENCE or
             (args.sensors is not None and det_name not in sensor_list)):
         continue
+    logger.info("processing %s", det_name)
     gs_det = make_galsim_detector(camera_wrapper, det_name, phot_params, obs_md)
     desc.imsim.add_treering_info([gs_det])
-    my_flat = desc.imsim.make_flat(gs_det, counts_per_iter, niter, rng)
+    my_flat = desc.imsim.make_flat(gs_det, counts_per_iter, niter, rng,
+                                   logger=logger, wcs=None)
     ccd_id = "R{}_S{}".format(det_name[2:5:2], det_name[8:11:2])
     my_flat.write('flat_{}_{}_{}.fits'.format(visit, ccd_id, obs_md.bandpass))
--- a/bin.src/make_flats.py
+++ b/bin.src/make_flats.py
@@ -21,6 +21,8 @@ parser.add_argument('--counts_per_iter', type=int, default=4e3,
 parser.add_argument('--log_level', type=str,
                     choices=['DEBUG', 'INFO', 'WARN', 'ERROR', 'CRITICAL'],
                     default='INFO', help='Logging level. Default: INFO')
+parser.add_argument('--wcs_file', type=str, default=None,
+                    help='FITS file to use for the WCS (as galsim.FitsWCS)')
 
 args = parser.parse_args()
 
@@ -39,6 +41,8 @@ visit = obs_md.OpsimMetaData['obshistID']
 
 camera_wrapper = LSSTCameraWrapper()
 
+wcs = galsim.FitsWCS(args.wcs_file) if args.wcs_file is not None else None
+
 for det in camera_wrapper.camera:
     det_name = det.getName()
     if (det.getType() != cameraGeom.SCIENCE or
@@ -48,6 +52,6 @@ for det in camera_wrapper.camera:
     gs_det = make_galsim_detector(camera_wrapper, det_name, phot_params, obs_md)
     desc.imsim.add_treering_info([gs_det])
     my_flat = desc.imsim.make_flat(gs_det, counts_per_iter, niter, rng,
-                                   logger=logger, wcs=None)
+                                   logger=logger, wcs=wcs)
     ccd_id = "R{}_S{}".format(det_name[2:5:2], det_name[8:11:2])
     my_flat.write('flat_{}_{}_{}.fits'.format(visit, ccd_id, obs_md.bandpass))
--- a/bin.src/make_flats.py
+++ b/bin.src/make_flats.py
@@ -1,0 +1,51 @@
+#!/usr/bin/env python
+"""
+Script to create imSim flats.  This includes electorstatic sensor effects
+such as treerings and brighter-fatter.
+"""
+import argparse
+import galsim
+import lsst.afw.cameraGeom as cameraGeom
+from lsst.sims.GalSimInterface import LSSTCameraWrapper, make_galsim_detector
+import desc.imsim
+
+parser = argparse.ArgumentParser("imSim flat production script.")
+parser.add_argument('instcat', type=str, help="instance catalog file")
+parser.add_argument('--sensors', type=str, default=None,
+                    help="sensors to simulate")
+parser.add_argument('--seed', type=int, default=42, help="random seed")
+parser.add_argument('--counts_total', type=int, default=8e4,
+                    help="total # electrons/pixel in flat")
+parser.add_argument('--counts_per_iter', type=int, default=4e3,
+                    help="# electrons/pixel per iteration")
+parser.add_argument('--log_level', type=str,
+                    choices=['DEBUG', 'INFO', 'WARN', 'ERROR', 'CRITICAL'],
+                    default='INFO', help='Logging level. Default: INFO')
+
+args = parser.parse_args()
+
+obs_md, phot_params, _ = desc.imsim.parsePhoSimInstanceFile(args.instcat,
+                                                            numRows=30)
+sensor_list = args.sensors.split('^') if args.sensors is not None \
+    else args.sensors
+rng = galsim.UniformDeviate(args.seed)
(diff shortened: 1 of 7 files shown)


tests/test_flats.py
```diff
@@ -0,0 +1,50 @@
...
+    """
+    def __init__(self):
+        self.xMaxPix = 256
+        self.yMaxPix = 256
+        self.xMinPix = 1
+        self.yMinPix = 1
+        self.wcs = galsim.wcs.PixelScale(0.2)
+        self.tree_rings \
+            = TreeRingInfo(galsim.PositionD(0, 0),
+                           galsim.SiliconSensor.simple_treerings(0.26, 47))
```
cwwalter: Why test with the simple treerings as opposed to the model being used in the code?

tests/test_flats.py
```diff
@@ -0,0 +1,50 @@
...
+    def test_make_flat(self):
+        """Test of make_flat function."""
+        gs_det = GsDetector()
+        counts_per_iter = 1000
+        niter = 10
+        rng = galsim.UniformDeviate()
+        logger = desc.imsim.get_logger('WARN')
+        flat = desc.imsim.make_flat(gs_det, counts_per_iter, niter, rng,
+                                    logger=logger)
+        self.assertLess(np.abs(niter*counts_per_iter - flat.array.mean()),
```
cwwalter: I'm not quite sure I understand what is being tested here and how it relates to the tree rings model.  I think this is just checking that the overall level of the flat is correct and the tree rings are irrelevant.  So mostly this is just checking the flat level itself.  I think a comment saying what is being checked would be useful.  The thing I'm not sure about is how the variance caused by the addition of the treerings changes the std().

cwwalter:
BTW, 37 minutes, seems quite a  bit longer than what we calculated when we did our tests before. Any idea why?

tests/test_flats.py
jchiang87: It's generally a good idea to keep these things as decoupled as possible.  If someone does some work on the treering code in this package, any bugs in that new work shouldn't prevent the flat production code from passing its tests.

bin.src/make_flats.py
```diff
@@ -0,0 +1,53 @@
...
+                    help="# electrons/pixel per iteration")
+parser.add_argument('--log_level', type=str,
+                    choices=['DEBUG', 'INFO', 'WARN', 'ERROR', 'CRITICAL'],
+                    default='INFO', help='Logging level. Default: INFO')
+
+args = parser.parse_args()
+
+config = desc.imsim.read_config()
+
+obs_md, phot_params, _ = desc.imsim.parsePhoSimInstanceFile(args.instcat,
```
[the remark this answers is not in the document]
jchiang87: This code is doing essentially the same calculation as `FastSiliconSkyModel` where a specific SED isn't used either.   As far as required instance catalog contents are concerned, this code just needs a complete set of phosim commands so that the output files have the same header information.  This is a requirement to make the `raw` files that would be ingested into the repo using `obs_lsstCam`.   I'd prefer to put all of that sort of documentation in the wiki, since there is a procedure associated with making the flats and other calibration products, so it would be better to have it all in one place.

jchiang87:
What was the original estimate?  One part that is definitely slower is using the LSST  WCS object versus a galsim version (e.g., `FitsWCS`), but it's not that much slower:  for the same parameters, using that galsim WCS, the same simulation takes somewhat less than 30 minutes.

tests/test_flats.py
jchiang87: This is mostly an operational test that the flats had the expected level.  Using stdev isn't very stringent (and not the test to use if one were really concerned about the statistical properties of the pixel values, which we aren't here), but it's convenient since I was playing around with image sizes and total counts per pixel for the test and wanted something that would adjust for all that automatically.  In any case, this test code is not testing the treering implementation; it's testing the  flat generation.

cwwalter:
> What was the original estimate? One part that is definitely slower is using the LSST WCS object versus a galsim version (e.g., FitsWCS), but it's not that much slower: for the same parameters, using that galsim WCS, the same simulation takes somewhat less than 30 minutes.

I was remembering more like 20-30s/amp from the slack discussions when we implemented the static filling. But this is of order 2 min per amp.. So something like 4 to 5x more.  But, maybe that was a  restricted test.

bin.src/make_flats.py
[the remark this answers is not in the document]
rmjarvis: For a flat, I don't think the SED matters at all.  At least if we're ignoring fringing and/or assuming an SED without spectral lines.  Aside from that the only ways that the SED can affect anything are:

1. Setting the overall level from the integral of SED(lambda) * throughput(lambda) dlambda.  But I'm sure we don't actually care about getting that right for some physical SED and exposure time.  Rather, we are just targeting an overall level in the final image.

2. There is a slight effect in terms of the diffusion, since the different wavelengths convert at different levels in the silicon, so they will have different effective diffusion lengths.  However (a) we are already ignoring diffusion entirely in the approximation we are doing to make this fast and (b) diffusion doesn't matter anyway for a uniform input surface brightness, just spreading that around into a new uniform surface brightness with the electrons in different locations.

So I think it's completely correct to ignore the SED here.

rmjarvis:
> BTW, 37 minutes, seems quite a bit longer than what we calculated when we did our tests before. Any idea why?

It's probably worth doing a cProfile analysis of a single CCD run, just to make sure we understand what the tall pole is and there isn't some simple mistake that is making it take more time than it should.  (A good idea for the regular image code too.)

python/desc/imsim/flats.py
```diff
@@ -0,0 +1,90 @@
...
+        Number of iterations. Final flat will have niter*counts_per_iter
+        electrons/pixel.
+    rng: galsim.BaseDeviate
+        Random number generator.
+    buf: int [2]
+        Pixel buffer around each to account for charge redistribution
+        across pixel boundaries.
+    logger: logging.Logger [None]
+        Logging object. If None, then a default logger will be used.
+    wcs: galsim.fitswcs.GSFitsWCS [None]
```
rmjarvis: Any `galsim.WCS` will work here.  The GSFitsWCS is a particular subclass that I used in my test.  But for instance, you use a PixelScale in one of your tests, which is another subclass of WCS.

python/desc/imsim/flats.py
```diff
@@ -0,0 +1,90 @@
...
+    Full LSST CCDs are assembled one amp at a time to limit the memory
+    used by the galsim.SiliconSensor model.
+
+    Parameters
+    ----------
+    gs_det: GalSimDetector
+        The detector in the LSST focalplane to use.  This object
+        contains the CCD pixel geometry, WCS, and treering info.
+    counts_per_iter: int
+        Roughly the number of electrons/pixel to add at each iteration.
```
rmjarvis: "Roughly" reads funny, like there is some uncertainty in how much will be added.  And while that's true for each pixel, the mean value is fairly precisely equal to `counts_per_iter`.  So I'd suggest writing "Mean number of electrons/pixel" instead to be clearer.

python/desc/imsim/flats.py
```diff
@@ -0,0 +1,90 @@
...
+            logger.info("amp: %d, %d", i, j)
+            ymin = j*dy + 1
+            ymax = (j + 1)*dy
+            temp_image = galsim.ImageF(ncol, nrow, wcs=wcs)
+            bounds = (galsim.BoundsI(xmin-buf, xmax+buf, ymin-buf, ymax+buf)
+                      & temp_image.bounds)
+            temp_amp = temp_image[bounds]
+            for it in range(niter):
+                logger.debug("iter %d", it)
+                temp = sensor.calculate_pixel_areas(temp_amp)
```
jchiang87: @rmjarvis From the cProfile output, this pixel area calculation is the main hot spot.  I assume there's no reason not to move this outside of the `for it` loop?

bin.src/make_flats.py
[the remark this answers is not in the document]
cwwalter: > I'd prefer to put all of that sort of documentation in the wiki, since there is a procedure associated with making the flats and other calibration products, so it would be better to have it all in one place.

OK that makes sense.  You mean the DM processing wiki as opposed to imSim documentation here is that right? It will  probably be good for us to also go back and document this in the imSim area too.

python/desc/imsim/flats.py
jchiang87: Moving it outside the for loop, reduces the execution time for the default parameters to ~5 minutes.

rmjarvis: No.  You can't move it outside the for loop.  That's the thing that accounts for the brighter-fatter as the flux is building up.  Moving it out of the for loop would be equivalent to setting niter=1.  (And readjusting counts_per_iter.)

But that's good that this is the hot spot.  I think that's what we expect, since it's doing a fair bit of calculation, figuring out the area of the pixel polynomials.  And I already spent some time optimizing the C++ code that does this.  There might be 10% or so more improvements possible, but I don't think there are large factors of speed up left in that code.

cwwalter: I went back and looked at our original conversation in slack when we first discussed this.. Mike said 

"The script to do this is on the silicon_speedup branch now.  (I just pushed it -- I think it's all working properly now.)  It makes 10 flats.  As written, it takes about 15 minutes on my laptop (for all 10).   It has tree rings turned on.  As well as a non-trivial WCS.  Mostly to make sure these work properly.  So you can either turn them off or take differences."

Then, Jim ran  a sensor in about  ~15 minutes and we decided most of this was due to the LSST WCS.  

For dome flats at sky level where we are below the level where BF kicks in, we should be able to get away with not recalculating those shapes right?  For running in that model can we just use niter=1?

cwwalter: OK, so now I realize the difference in what I was remembering for timing. Those were for actual sky levels where we were below BF and this is what is being used in imSim for the sky.

What level do we need here? 10K?

rmjarvis: 80K is the default in Jim's script, which is pretty typical for real dome flats.  He does 20 iterations, which seems like a good choice to me.  (It matches what I did in my version of this script.)

I don't see any reason to do low-intensity dome flats for imsim, but if you wanted to do something that tries to mimic the behavior of sky flats, then yes setting niter=1 would do it.

cwwalter: I thought we needed both high intensity for BF (80K is good for that) and low intensity for treerings, flat fielding etc. Do you mean since there is no vignetting it isn't needed here?

rmjarvis: I mean low-intensity dome flats aren't normally a thing.  We're just doing that for PhoSim, since it's exorbitant to run flats.  The high intensity ones do everything we need for imsim.

jchiang87: Here's the output of running with cProfile for a 4 iteration run with the pixel area calculation moved back into the loop:
```
In [3]: stats.strip_dirs().sort_stats('tottime').print_stats(10)
Sat Aug 11 20:06:37 2018    make_flats_cProfile.txt

         42367001 function calls (42251154 primitive calls) in 614.059 seconds

   Ordered by: internal time
   List reduced from 10675 to 10 due to restriction <10>

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
       64  340.360    5.318  340.435    5.319 sensor.py:273(calculate_pixel_areas)
      401   30.455    0.076   30.455    0.076 LsstCameraUtils.py:643(<listcomp>)
      401   29.950    0.075   29.950    0.075 LsstCameraUtils.py:647(<listcomp>)
      401   29.363    0.073   29.363    0.073 CameraUtils.py:1031(<listcomp>)
      401   26.412    0.066   26.412    0.066 CameraUtils.py:1029(<listcomp>)
      402   17.127    0.043  253.825    0.631 galSimCameraWrapper.py:616(_raDecFromPixelCoords)
      403   13.241    0.033  197.801    0.491 LsstCameraUtils.py:571(pupilCoordsFromPixelCoordsLSST)
    24665   12.507    0.001   12.508    0.001 {built-in method numpy.core.multiarray.array}
      402   11.135    0.028   72.932    0.181 CameraUtils.py:1001(pupilCoordsFromFocalPlaneCoords)
      402   11.065    0.028   11.065    0.028 {palpy.ampqkVector}
```
All those list comprehensions in the sims code are the expensive part of the sky background calculation, which has only one iteration  Those sims calculations are done by the wcs implementation and only once. As a result, the 20 iteration timing is dominated by the sensor code even if we use the initially slower LSST wcs code.

rmjarvis: Thanks.  I think that all makes sense.  The LSST wcs code could still be sped up a lot I think, but not really a job for today.  I'd keep the production run at around 20 iterations.  Or specifically, I think 4000 counts per iteration is pretty good to make sure the pixel areas don't change much between iterations (and therefore the quantization of that calculation isn't causing problems).

jchiang87:
I'll merge this at 9am PT tomorrow unless I hear otherwise.

PR #136, opened 2018-08-16
Use obs_lsstCam for pixel geometry and readout parameters

jchiang87:


data/default_imsim_configs
```diff
@@ -1,5 +1,9 @@
 [electronics_readout]
 readout_time = 3.
+dark_current = 0.02
+bias_level = 1000.
+pcti = 1e-6
```
cwwalter: Maybe a inline comment `# Parallel CTI level` for here and serial for non-experts who might not be able to figure out the acronym?

python/desc/imsim/camera_readout.py
```diff
@@ -206,43 +237,47 @@ def _make_amp_image(self, amp_name):
...
 
         # Add dark current.
-        full_arr += np.random.poisson(amp_props.dark_current*self.exptime,
+        dark_current = config['electronics_readout']['dark_current']
+        full_arr += np.random.poisson(dark_current*self.exptime,
                                       size=full_arr.shape)
 
         # Add defects.
 
         # Apply CTE.
```
cwwalter: I think a bit more comments/documentation here would be a good idea pointing out that the whole matrix is calculated and then applied column-by-column.  It is not super obvious with this compact code.

It could point out how/why both the efficiency and inefficient are used.

python/desc/imsim/camera_readout.py
```diff
@@ -256,25 +291,27 @@ def _add_read_noise_and_bias(self, amp_name):
...
-        amp_props = self.fp_props.get_amp(amp_name)
+        amp_info = self.camera_info.get_amp_info(amp_name)
         full_arr = self.amp_images[amp_name].getArray()
-        full_arr += np.random.normal(scale=amp_props.read_noise,
+        full_arr += np.random.normal(scale=amp_info.getReadNoise(),
                                      size=full_arr.shape)
-        full_arr += amp_props.bias_level
+        full_arr += config['electronics_readout']['bias_level']
 
     def _apply_crosstalk(self):
```
cwwalter: I think I would put this after _apply_crosstalk so it is in the same order of the operations. That will also make the docstrings a little easier to understand.

python/desc/imsim/camera_readout.py
```diff
@@ -256,25 +291,27 @@ def _add_read_noise_and_bias(self, amp_name):
...
+        amp_names = self.camera_info.get_amp_names(self.sensor_id)
         imarrs = np.array([self.amp_images[amp_name].getArray()
-                           for amp_name in sensor_props.amp_names])
-        for amp_name in sensor_props.amp_names:
-            amp_props = self.fp_props.get_amp(amp_name)
+                           for amp_name in amp_names])
+        for amp_name, xtalk_row in zip(amp_names, xtalk):
             self.amp_images[amp_name].getArray()[:, :] \
-                = sum([x*y for x, y in zip(imarrs, amp_props.crosstalk)])
+                = sum([x*y for x, y in zip(imarrs, xtalk_row)])
```
cwwalter: Wow :)

python/desc/imsim/camera_readout.py
```diff
@@ -292,29 +329,46 @@ def get_amplifier_hdu(self, amp_name):
...
+                except ValueError:
                     # eimages produced by phosim contain non-ASCII or
                     # non-printable characters resulting in a ValueError.
                     self.logger.warn("ValueError raised while attempting to "
-                                     "read {} from eimage header".format(key))
+                                     "read %s from eimage header", key)
+
+        # Transpose the WCS matrix elements to account for the use of the
+        # Camera Coordinate System in the eimage.  These changes
+        # neglect any implied changes in the SIP coefficients.
```
cwwalter: I don't follow this comment about changes in the SIP coefficients.

python/desc/imsim/camera_readout.py
```diff
@@ -407,13 +471,13 @@ def set_itl_bboxes(amp):
         The updated AmpInfoRecord.
     """
     amp.setRawBBox(afwGeom.Box2I(afwGeom.Point2I(0, 0),
-                                 afwGeom.Extent2I(532, 2020)))
+                                 afwGeom.Extent2I(544, 2048)))
```
cwwalter: Not for now: Could these geometries come from the camera object so that we could handle both E2V and ITL here?

bin.src/write_amp_image_file.py
```diff
@@ -11,17 +11,19 @@
...
 parser = argparse.ArgumentParser()
 parser.add_argument("eimage_file", help="eimage file to process")
-parser.add_argument('--log_level', type=str,
+parser.add_argument("--log_level", type=str,
                     choices=['DEBUG', 'INFO', 'WARN', 'ERROR', 'CRITICAL'],
                     default='INFO', help='Logging level. Default: "INFO"')
-parser.add_argument("--seg_file", default=None, type=str,
-                    help="text file describing the layout of the focalplane")
+parser.add_argument("--opsim_db", default=None, type=str,
+                    help="OpSim db file as alternative source of pointing info")
```
cwwalter: I think it is good to specify this is a local file so people know they need to download one.

python/desc/imsim/camera_readout.py
```diff
@@ -206,43 +237,47 @@ def _make_amp_image(self, amp_name):
...
         imaging_segment.getArray()[:] = data
         full_arr = full_segment.getArray()
 
         # Add dark current.
-        full_arr += np.random.poisson(amp_props.dark_current*self.exptime,
+        dark_current = config['electronics_readout']['dark_current']
+        full_arr += np.random.poisson(dark_current*self.exptime,
                                       size=full_arr.shape)
 
         # Add defects.
```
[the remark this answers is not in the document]
jchiang87: Incidently, the dark current is being set incorrectly here.  I've done it correctly in 51b87942 and will apply the same change here.

python/desc/imsim/camera_readout.py
jchiang87: So are you suggesting that the read noise and bias levels should be executed after the crosstalk is applied?  That's consistent with what Robert has said the order should be, but that's not in fact how the correction is applied in `ip_isr` (afaict): https://github.com/lsst/ip_isr/blob/master/python/lsst/ip/isr/isrTask.py#L540
Since the bias levels, we've been using are so low (1000 adu), it's a negligible effect, so we could apply the crosstalk first, but I'd still like some clarity on what's going on in `ip_isr`.

python/desc/imsim/camera_readout.py
jchiang87: If you look at the header of one of our eimage files, you'll see that there are coefficients for the SIP polynomials that have x- and y-indexes.   In going from CCS to `raw` coordinates, one needs to swap x- and y-directions in the `CD_x_y` array, but I'm not applying that swap to the SIP coefficients.  This explanation seems too complicated to put in a comment, but I think anyone with detailed knowledge of the issues would understand the current comment.

python/desc/imsim/camera_readout.py
jchiang87: This function is only used by the `test_camera_readout.py` code, so I will move this function there.  These numbers pretty much have to be hardcoded like this (or at least gotten from some source other than `obs_lsstCam`) otherwise the test wouldn't be checking that the expected values were being used.

python/desc/imsim/camera_readout.py
```diff
@@ -161,40 +195,37 @@ def get_amp_image(self, amp_info_record, imageFactory=afwImage.ImageI):
...
         """
-        return '_'.join((self.sensor_id,
-                         'C%s' % amp_info_record.getName()[::2]))
+        return '_'.join((self.sensor_id, amp_info.getName()))
 
     def _make_amp_images(self):
         """
         Make the amplifier images for all the amps in the sensor.
         """
         self.amp_images = OrderedDict()
```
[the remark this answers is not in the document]
jchiang87: It is being done, i.e., called, in the `__init__`.  I purposely factored it into its own function to make that constructor easier to read.

python/desc/imsim/camera_readout.py
cwwalter: Sorry... I wasn't suggesting changing the order of the operations. I was suggesting change the order of these functions in the file so that they *matched* the order of the operations.  When I was reading it I got confused because of this.

PR #137, opened 2018-08-16
issue/117/persist_atm_psf

jchiang87:
These changes enable the PSF to be saved to and loaded from a pickle file.  It also fixes some handling of the rng seeding: for `imsim.py`, it uses the `seed` value from the instance catalog to seed the rng for the atmospheric PSF; and it passes the seed to the sky models so that simulations with the same seed will agree at the pixel level.

### DIFF
--- a/tests/test_psf.py
+++ b/tests/test_psf.py
@@ -46,7 +46,7 @@ class PsfTestCase(unittest.TestCase):
                                'tiny_instcat.txt')
         obs_md, _, _ = desc.imsim.parsePhoSimInstanceFile(instcat)
         for psf_name in ("DoubleGaussian", "Kolmogorov", "Atmospheric"):
-            psf = desc.imsim.make_psf(psf_name, obs_md)
+            psf = desc.imsim.make_psf(psf_name, obs_md, screen_scale=0.4)
             psf_file = os.path.join(self.test_dir, '{}.pkl'.format(psf_name))
             desc.imsim.save_psf(psf, psf_file)
             psf_retrieved = desc.imsim.load_psf(psf_file)
--- a/tests/test_psf.py
+++ b/tests/test_psf.py
@@ -1,0 +1,57 @@
+"""
+Unit tests for PSF-related code.
+"""
+import os
+import glob
+import unittest
+import desc.imsim
+
+
+def are_psfs_equal(psf1, psf2):
+    """
+    Test that two PSFs are equal.  For PSFs implemented in
+    sims_GalSimInterface, it is sufficient to compare the
+    ._cached_psf attributes.  For the AtmosphericPSF, only certain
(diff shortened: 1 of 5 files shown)


tests/test_psf.py
```diff
@@ -0,0 +1,57 @@
...
+    ._cached_psf attributes.  For the AtmosphericPSF, only certain
+    attributes are meaningful to compare. See issue #117.
+    """
+    if type(psf1) != type(psf2):
+        return False
+    if not isinstance(psf1, desc.imsim.atmPSF.AtmosphericPSF):
+        # Compare cached galsim objects.
+        return psf1._cached_psf == psf2._cached_psf
+    # See issue #117 for an explanation of these comparisons:
+    return (psf1.atm[:-1] == psf2.atm[:-1]) and (psf1.aper == psf2.aper)
```
jmeyers314: Rather than this awkward comparison, I think we should write an `__eq__` for `AtmosphericPSF` (and whatever classes also need an `__eq__` downstream).

tests/test_psf.py
```diff
@@ -0,0 +1,57 @@
...
+    def test_save_and_load_psf(self):
+        """
+        Test that the different imSim PSFs are saved and retrieved
+        correctly.
+        """
+        instcat = os.path.join(os.environ['IMSIM_DIR'], 'tests',
+                               'tiny_instcat.txt')
+        obs_md, _, _ = desc.imsim.parsePhoSimInstanceFile(instcat)
+        for psf_name in ("DoubleGaussian", "Kolmogorov", "Atmospheric"):
+            psf = desc.imsim.make_psf(psf_name, obs_md, screen_scale=0.4)
```
jmeyers314: I don't have a laptop with imSim on it to check the runtime of this at the moment, but this seems potentially overkill.  Setting screen_scale=6.4 or 12.8 or something would be sufficient if that runs faster.

jchiang87: With this value, the test runs in ~10 s on my laptop.  At full resolution, the test was slow, ~200 s, but the real issue was the memory usage:  it would crash on my laptop and in travis with a memory error.  I think the current value is fine, but even larger would be ok too.

tests/test_psf.py
jmeyers314: I pushed an update.  

I think this is ready to go now.  I'll rebase branch r0_500_fsolve once this is merged.

jchiang87: Thanks, Josh.  Now that `AtmosphericPSF` has `__eq__`, I will issue a PR at sims_GalSimInterface to provide that method for the other PSF classes so that we can clean up that test program.

jchiang87:
I'll merge this at 8pm PT tonight unless I hear otherwise.

PR #139, opened 2018-08-17
enable FITS postage stamps to be rendered

jchiang87:
This needs to be run with the [`u/jchiang/render_postage_stamps`](https://github.com/lsst/sims_GalSimInterface/tree/u/jchiang/render_postage_stamps) branch in `sims_GalSimInterface`.   Here is one of the sprinkled strongly lensed AGN disks rendered with this code:
lensed_agn_comparison
The image on the left is the "truth" image, and the image on the right is the imsim output.  I don't know why it's flipped in x.  The FITS files of the statically lensed objects provided by the sprinkler code do not have a WCS; there's just a pixel scale and rotation angle for each in the instance catalog entry.


bin.src/imsim.py
```diff
@@ -37,8 +38,18 @@
...
                     help='number of processes to use in multiprocessing mode')
+parser.add_argument('--image_dir_path', type=str, default=None,
+                    help="search path for FITS postage stamp images."
+                    "This will be prepended to any existing IMAGE_DIR_PATH"
+                    "environment variable, for which $CWD is included by"
+                    "default.")
 args = parser.parse_args()
 
+if args.image_dir_path is not None:
+    os.environ['IMAGE_DIR_PATH'] \
```
cwwalter: IMAGE_DIR_PATH seems a bit generic. Probably we will never have a conflict but to help people who are reseting things etc and looking at printenv maybe we could make it IMSIM_IMAGE_DIR_PATH or FITS_IMAGE_DIR_PATH?

python/desc/imsim/imSim.py
```diff
@@ -232,7 +236,20 @@ def sources_from_list(object_lines, obs_md, phot_params, file_name):
                 internal_rv[i_obj] = float(params[19])
             if params[i_gal_dust_model].lower() != 'none':
                 galactic_av[i_obj] = float(params[i_gal_dust_model+1])
-                galactic_rv[i_obj] =float(params[i_gal_dust_model+2])
+                galactic_rv[i_obj] = float(params[i_gal_dust_model+2])
+        elif (params[12].endswith('.fits') or params[12].endswith('.fits.gz')):
```
cwwalter: This is OK. But I wanted to record the little figure I used to check this in case I want to go back and find it again:

```
blah.fits pixel_size _axis1 rot_angle  DUST_REST dust_pars_1(2) DUST_LAB dust_pars_2(2)
12          13                        14              15                 16/17               18                19/20

blah.fits pixel_size _axis1 rot_angle   None       DUST_LAB dust_pars_2(2)
12           13                        14              15             16              17/18
```

This doesn't exactly line up when rendered by GH for some reason.

python/desc/imsim/imSim.py
```diff
@@ -371,6 +395,11 @@ def sources_from_list(object_lines, obs_md, phot_params, file_name):
 
     return gs_object_arr, out_obj_dict
 
+def get_image_dirs():
+    """
+    Return a list of possible directories for FITS images.
+    """
+    return os.environ.get('IMAGE_DIR_PATH', '.').split(':')
```
cwwalter: This gives you image_dir_path or '.' if it is not set.  Is this consistent with the command string?

"This will be prepended to any existing IMAGE_DIR_PATH environment variable, for which $CWD is included by default." ?

I think the code may leave out CWD if IMAGE_DIR_PATH is set.

bin.src/imsim.py
jchiang87: I worried about this too.  `IMSIM_IMAGE_PATH` seems sufficiently non-generic.

python/desc/imsim/imSim.py
jchiang87: Yes, that's true.  I realized that when I first wrote it, but didn't get around to implementing it properly.  I'll fix it.

jchiang87:
BTW, the flip in the imsim image wrt the truth image may be because RA is increasing to the right in the imsim image.

cwwalter:
Is the direction flip still an issue?

jchiang87:
I think it's ok, i.e., the explanation about the increasing direction of RA seems correct.

PR #140, opened 2018-08-18
Issue/87/cosmic ray rate

jchiang87:
I think this is good to go.  The upshot is that scaling the CTIO rate for LSST CCDs yields ~6 CRs per 30 s exposure.   The CR detection implementation in the notebook analyzing the DECam darks is cruder than the code used to extract the CR footprints from the ITL data, but its detection efficiency is probably good enough.  Here's a 1200s dark for part of one DECam amp with the detected CRs overlaid:

cr_detection_example

### DIFF
--- a/data/default_imsim_configs
+++ b/data/default_imsim_configs
@@ -10,7 +10,8 @@ eimage_compress = True
 centroid_prefix = centroid_
 
 [cosmic_rays]
-#ccd_rate = None   # Use the computed rate for the ITL lab data, ~1.2/sec.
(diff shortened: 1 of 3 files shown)


cwwalter:
That's great!  I talked to Pierre and got some numbers at the meeting (just got home and am now catching up) but this is much better.  This is from a different place than the files that Kevin Real pointed us too?

jchiang87:
Thanks, Chris.   The DECam darks were provided by Brian Yanny back in Dec 2017.

I'll go ahead and merge this tomorrow at 9am PT unless I hear otherwise.

PR #142, opened 2018-08-19
Expose the Silicon Sensor brighter fatter strength parameter to the imSim UI.

cwwalter:
The galsim.SiliconSensor class has a strength parameter that scales the size of the BF effect with 1.0 being the default amount in the model. These changes (and matching changes in sims_GalSimInterface) allow one to adjust the scaling parameter by editing a parameter in an imSim config file like this:

```
[ccd]
full_well = 1e5
bf_strength = 1.0
```

This is generally useful for tuning and may be necessary for run 2.0i depending on the outcome of issue #251.

### DIFF
--- a/data/default_imsim_configs
+++ b/data/default_imsim_configs
@@ -3,6 +3,7 @@ readout_time = 3.
 
 [ccd]
 full_well = 1e5
+bf_strength = 1.0
 
 [persistence]
 eimage_prefix = lsst_e_
--- a/python/desc/imsim/ImageSimulator.py
+++ b/python/desc/imsim/ImageSimulator.py
@@ -114,11 +114,16 @@ class ImageSimulator:
                                       noise_and_background,
                                       epoch=2000.0, seed=seed,
                                       apply_sensor_model=self.apply_sensor_model)
+
             self.gs_interpreters[det_name].sky_bg_per_pixel \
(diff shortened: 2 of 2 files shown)


cwwalter:
Thanks Jim. Both merged.  

@rmandelb: note this will allow us to scale BF if necessary as discussed in Tucson.

jchiang87:
Oops, I just remembered that we should set that parameter in these locations also:

https://github.com/LSSTDESC/imSim/blob/827cbd62f9b02c06a3278bd80e6fe39b9f410d56/python/desc/imsim/skyModel.py#L287

https://github.com/LSSTDESC/imSim/blob/4a552df5133cf32d550824232149ac9fd3cda723/python/desc/imsim/flats.py#L51

cwwalter:
Ugh... should have waited to merge.. 

make_flats has direct access to the config object so we can just pass it I guess.   But I need to look to find a more elegant way of getting it into the derived addNoiseAndBackground short of passing it down 3 to 4 levels.

jchiang87:
In both `addNoiseAndBackground` and `make_flat`, you can just call `config = get_config()`, and access the parameter value directly from the `config` object.  `get_config` essentially returns a singleton, so you can call it in any context you need to.

PR #143, opened 2018-08-20
Pass BF strength parameter to silicon model when making flats or drawing the sky.

cwwalter:
This PR follows on #142 and also passes the BF strength parameter to the sky model and flat production routines.


cwwalter:
I was checking a few more things.  The current test_skyModel.py bypasses this as it isn't really testing that feature and makes its own config file.  OK to leave that as is?

jchiang87:
Yes, I think so.  That test code is exercising the sky level calculation and doesn't call the `galsim.SiliconSensor` code at all.

PR #150, opened 2018-09-03
aggregate checkpoint information in a summary file

jchiang87:
This code maintains an sqlite db of checkpoint progress that's separate from the individual checkpoint pickle files so that the image simulation status of many subprocesses running in parallel on a node can be extracted from a single source.  This uses `multiprocessing.Pipe` connections to send the checkpoint summary info from the single sensor pool workers to the aggregator function that runs asynchronously in its own processing pool.

### DIFF
--- a/data/default_imsim_configs
+++ b/data/default_imsim_configs
@@ -17,7 +17,7 @@ ccd_rate = 0.2    # CTIO rate extracted from DECam darks.
 catalog = default
 
 [checkpointing]
-nobj = 200
+nobj = 500
 cleanup = True
 
 [wl_params]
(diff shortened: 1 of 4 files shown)


villarrealas:
Looks like this above and beyond what I was hoping for! Just to confirm a couple of things to make sure I am reading correctly though...

1. This uses its own processing pool inside the imSim instance. In practice, does this mean the # of hardware threads used is now # sensors + 1 for sqlite syncing?

2. Just confirming that the resulting file outputs to the same directory as the checkpoints with the same naming convention (but with .sqlite format) and contains entries of # objects drawn, # objects total, and the detector name? Is this detector name in the imSim convention that we make calls currently, or is that going to be changing?

PR #164, opened 2018-09-13
Fix optics psf

jmeyers314:
This PR adds a couple of bug fixes and a couple of other changes in an attempt to make the imSim optics PSF correct and more realistic.  See the discussion here: https://github.com/LSSTDESC/DC2-production/issues/259.  

I'm certainly open to the idea of only implementing a subset of these commits.  Given the time constraints, it'd be great to have as many eyes on this as possible.

Some output implementing these changes can be seen here: https://github.com/LSSTDESC/DC2-production/issues/259#issuecomment-421166639

Note that the above plot was created outside of ImSim, though I did use the atmPSF.py and optical_system.py in my external test.

### DIFF
--- a/python/desc/imsim/atmPSF.py
+++ b/python/desc/imsim/atmPSF.py
@@ -14,7 +14,8 @@ from .optical_system import OpticalZernikes, mock_deviations
 class OptWF(object):
     def __init__(self, rng, wavelength, gsparams=None):
         u = galsim.UniformDeviate(rng)
-        self.deviations = mock_deviations(seed=int(u()*2**31))
+        deviationsFudgeFactor = 3.0
+        self.deviations = deviationsFudgeFactor*mock_deviations(seed=int(u()*2**31))
         self.oz = OpticalZernikes(self.deviations)
         self.dynamic = False
         self.reversible = True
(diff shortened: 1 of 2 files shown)


python/desc/imsim/atmPSF.py
```diff
@@ -14,7 +14,8 @@
 class OptWF(object):
     def __init__(self, rng, wavelength, gsparams=None):
         u = galsim.UniformDeviate(rng)
-        self.deviations = mock_deviations(seed=int(u()*2**31))
+        deviationsFudgeFactor = 3.0
```
rmandelb: I think it would be good to put a comment above the definition of `deviationsFudgeFactor`  pointing to the GitHub thread that led to its creation and  briefly saying the nature of this fudge factor (i.e., that it was based on empirical attempts to get reasonable optical PSF shape given that our model was based on a very small number of sims that  don't span the space expected in reality).  This will help with any inevitable future confusion over where this came from etc.

jmeyers314:
Thanks @rmjarvis and @rmandelb.

I'll leave this for whoever's doing the ImSim DC2.2i tagging and packaging to merge when they're ready for that.

jchiang87:
I'm basically happy to merge this into the `dc2_run2.0_rc` branch at any time.   Unless I hear otherwise, I'll do that around Noon tomorrow.

PR #167, opened 2018-09-16
disable checkpoint file deletion for Run2.0i

jchiang87:


cwwalter:
I think we want this not to go into master, so we should remember to make this change if we wind up just merging this release branch with all these changes to master.  I'll make a note on the issue where we are tracking all the PRs that went into the branch.

PR #174, opened 2018-09-25
U/jchiang/trim baseline version

jchiang87:


python/desc/imsim/imSim.py
```diff
@@ -375,9 +382,13 @@ def sources_from_list(object_lines, obs_md, phot_params, file_name):
     assert len(x_pupil) == len(gs_object_arr)
     assert len(y_pupil) == len(gs_object_arr)
 
+    logger.debug('down-selecting by chip')
     out_obj_dict = {}
     for det in lsst_camera():
         chip_name = det.getName()
+        if target_chip is not None and chip_name != target_chip:
```
jchiang87: If there is a specific sensor being considered (`target_chip`), then this if statement will skip the calculations below for the other 188 sensors.

python/desc/imsim/trim.py
```diff
@@ -99,14 +103,24 @@ def get_object_entries(self, chip_name, radius=0.18, sort_magnorm=True):
                     number of sersic objects for minsource application)
 
         """
+        self.trimmer.logger.debug("computing object offsets from %s center",
+                                  chip_name)
         ra0, dec0 = self.compute_chip_center(chip_name)
         seps = degrees_separation(ra0, dec0, self._ra, self._dec)
-        index = np.where(seps < radius)
+        if chip_name in self.trimmer.drawn_objects:
+            not_drawn = [not x in self.trimmer.drawn_objects[chip_name]
```
jchiang87: If this run starts from a checkpoint, this code omits objects that have already been drawn from the objects that are passed to the downstream object processing code.  In addition to speeding up that processing, the per sensor memory usage can be substantially reduced.

jchiang87:
I'm going to merge this tomorrow at 9am PT unless I hear otherwise.

bin.src/make_flats.py
```diff
@@ -54,4 +54,6 @@
     my_flat = desc.imsim.make_flat(gs_det, counts_per_iter, niter, rng,
                                    logger=logger, wcs=wcs)
     ccd_id = "R{}_S{}".format(det_name[2:5:2], det_name[8:11:2])
-    my_flat.write('flat_{}_{}_{}.fits'.format(visit, ccd_id, obs_md.bandpass))
+    prefix = config['persistence']['eimage_prefix']
+    my_flat.write('{}_{}_{}_{}.fits'.format(prefix, visit, ccd_id,
+                                            obs_md.bandpass))
```
cwwalter: It looks like this makes it so Flats are not differently labeled than normal e-image files.  Is that the intention?

python/desc/imsim/imSim.py
```diff
@@ -152,12 +152,16 @@ def metadata_from_file(file_name):
     return commands
 
 
-def sources_from_list(object_lines, obs_md, phot_params, file_name):
+def sources_from_list(object_lines, obs_md, phot_params, file_name,
+                      target_chip=None, log_level='INFO'):
```
cwwalter: This function should have a docstring.

python/desc/imsim/trim.py
cwwalter: Old comment I originally typed:

```
I'm having a hard time understanding the 'chip_name in self.trimmer.drawn_objects' since in the gs_intepreter the list is made out of  uniqueIds here: self.drawn_objects.add(gsObject.uniqueId).  How does the chip_name get in?

I think it is related to the fact that the sensor name is in the checkpoint file?
```

OK, I think I understand this now. The trimmer drawn_objects is different:

self.drawn_objects[detname] = ckpt['drawn_objects']

I think maybe the names of the two variables should be different for clarity.  Perhaps one could be drawn_objects_on_sensor or something?

python/desc/imsim/trim.py
```diff
@@ -99,14 +103,24 @@ def get_object_entries(self, chip_name, radius=0.18, sort_magnorm=True):
...
+            self.trimmer.logger.debug("avoiding drawn objects")
+            self.trimmer.logger.debug(index)
+        else:
+            index = np.where(seps < radius)
 
         # Collect the selected objects.
         selected = [self.object_lines[i] for i in index[0]]
         if sort_magnorm:
             # Sort by magnorm.
+            self.trimmer.logger.debug('sorting by magnorm')
```
cwwalter: Is this going to happen for all runs?  Only ones when multiprocessing is used?

It looks like InstCatTrimmer is always being called from parsePhoSimInstanceFile.

For simple debugging for individuals (rather than big production) having the processing order different than what is in the file will be confusing and will make debugging difficult.  If it is always going to happen maybe we could make the default to be non-sorted and then allow a flag?

cwwalter: Yes, but maybe the last time I did it and really cared about the order was before the sorting started?  If you have a catalog and you want to go through and track what it is doing line by line it  would be confusing if the order wasn't what you input.

Also, when doing testing I often will do (say) the 1st 2000 objects for speed. What happens in that case? Do you get the 2000 brightest objects? This might actually explain some confusing things I have seen recently when I predicted times based on running a fraction of a file.

True about things changing if we break things up.  Maybe we should set up three modes if we do that: "as read", "magnum sorted", and "interleaved"?

bin.src/make_flats.py
cwwalter: OK, git it. I see.  I guess at least the visit number will be different.  Maybe we could have some convention for flat visit # sequences in order to be able to quickly guess from the file name?  AIso, I suppose they typically won't show up in the same directories.

jchiang87: I proposed a convention in the ImageProcessingPipeline wiki in the [Generating Calibration Products](https://github.com/LSSTDESC/ImageProcessingPipelines/wiki/Generating-Calibration-Products) entry: `3xxxxxx`, `4xxxxxx`, `5xxxxxx` for biases, darks, and flats, respectively.

python/desc/imsim/imSim.py
jchiang87: done

python/desc/imsim/trim.py
jchiang87: I added the config parameter.  We can revisit this when we look into breaking up the bright objects.

cwwalter:
OK these all look good. Thanks!

PR #176, opened 2018-10-08
memory fix for chip-level object down-selection

jchiang87:


python/desc/imsim/imSim.py
```diff
@@ -152,6 +152,12 @@ def metadata_from_file(file_name):
     return commands
 
 
+def uss_mem():
+    """Return unique set size memory of current process in GB."""
+    my_process = psutil.Process(os.getpid())
+    return my_process.memory_full_info().uss/1024.**3
```
cwwalter: I did some research and it looks like this is also supposed to work on OSX:

http://www.pybloggers.com/2016/02/psutil-4-0-0-and-how-to-get-real-process-memory-and-environ-in-python/

But it might not be bad to double check this doesn't do something crazy if we aren't on Linux.

python/desc/imsim/imSim.py
```diff
@@ -343,58 +350,77 @@ def sources_from_list(object_lines, obs_md, phot_params, file_name,
         message += "    %d had n_points <= 0 \n" % n_bad_knots
         warnings.warn(message)
 
-    wav_int = None
-    wav_gal = None
+    target_chips = [target_chip] if target_chip is not None else None
```
cwwalter: From the doctoring etc  if target_chip is None then you should use all sensors. This is accomplished here by having the target_chips=None interpreted in `_chip_downselect` as meaning you should use all of them with `[det.getName() for det in lsst_camera()]`.

FWIW I found this confusing when I first read the code.  This is far enough done in the function that a comment here and in the docstring of `_chip_downselect` would probably be useful.

jchiang87:
The code like this
```
on_chip = np.where(np.logical_or(mag_norm<max_mag,
                           np.logical_and(xpix>x_min-pix_tol,
                           np.logical_and(xpix<x_max+pix_tol,
                           np.logical_and(ypix>y_min-pix_tol,
                                          ypix<y_max+pix_tol)))))
```
isn't new of course, and so it isn't really part of this PR.

cwwalter:
> isn't new of course, and so it isn't really part of this PR.

Yes, I know...

I was just saying I struggled a bit with some of the code.  There were enough changes in the diff that I went over it all again from scratch.

python/desc/imsim/imSim.py
rmjarvis: No worries.  It works on OSX.  (My Mac at least.)  And it has both `rss` and `uss` attributes as required.

python/desc/imsim/imSim.py
```diff
@@ -439,11 +451,8 @@ def sources_from_list(object_lines, obs_md, phot_params, file_name,
                            np.logical_and(ypix>y_min-pix_tol,
                                           ypix<y_max+pix_tol)))))
```
rmjarvis: FWIW, this is equivalent to the more legible
```
on_chip = np.where( (mag_norm < max_mag) |
                    ( (xpix > x_min - pix_tol) & (xpix < x_max + pix_tol) &
                      (ypix > y_min - pix_tol) & (ypix < y_max + pix_tol) )
                  )
 ```

cwwalter: OK thank you. I

n anycase, I didn't actually find any problems with the code.  So, it looks good to me.

jchiang87:
Unless I hear otherwise, I'll merge this at 12 Noon PT today.

jchiang87:
@danielsf I had to update the `test_instcat_parser.py` tests to account for the changes to the chip downselection in `source_from_list`.  Could you have a look please?  I'll hold off on merging until you've given your ok.  Thanks!

jchiang87:
Thanks, Scott.  I'll merge at 1pm PT today.

PR #191, opened 2018-11-24
enable eimage and raw files to be persisted independently

jchiang87:
Also,
* allow for existing files to be overwritten
* use compression level 1 for faster gzipping
* switch to new sims interface for extinction calculations

### DIFF
--- a/python/desc/imsim/ImageSimulator.py
+++ b/python/desc/imsim/ImageSimulator.py
@@ -479,7 +479,7 @@ class SimulateSensor:
         if IMAGE_SIMULATOR.config['persistence']['eimage_compress']:
             compress_files(outfiles)
 
-def compress_files(file_list, remove_originals=True):
+def compress_files(file_list, remove_originals=True, compresslevel=1):
     """
     Use gzip to compress a list of files.
 
@@ -489,6 +489,8 @@ def compress_files(file_list, remove_originals=True):
         A list of the names of files to compress.
     remove_originals: bool [True]
         Flag to remove original files.
+    compresslevel: int [1]
+        Compression level for gzip.  1 is fastest, 9 is slowest.
 
     Notes
     -----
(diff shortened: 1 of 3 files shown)


python/desc/imsim/ImageSimulator.py
```diff
@@ -425,27 +442,44 @@ def update_checkpoint_summary(self, gs_interpreter, num_objects):
...
+        """
+        Write the eimage files.
+
+        Parameters
+        ----------
+        gs_interpreter: GalSimInterpreter object
+        """
+        prefix = IMAGE_SIMULATOR.config['persistence']['eimage_prefix']
         obsHistID = str(IMAGE_SIMULATOR.obs_md.OpsimMetaData['obshistID'])
         nameRoot = os.path.join(IMAGE_SIMULATOR.outdir, prefix) + obsHistID
```
cwwalter: This is OK but I see in the write_raw_file function above you are leveraging the IMAGE_SIMULATOR.output_file function to build this path.  Is there are reason it has to be done manually here?

python/desc/imsim/sed_wrapper.py
```diff
@@ -38,10 +38,10 @@ def add_dust(self, sed_obj, Av, Rv, ext_type):
         """
         if (self.wavelen[ext_type] is None or
             not np.array_equal(sed_obj.wavelen, self.wavelen[ext_type])):
-            self.a[ext_type], self.b[ext_type] = sed_obj.setupCCMab()
+            self.a[ext_type], self.b[ext_type] = sed_obj.setupCCM_ab()
```
cwwalter: Are these just a new interface with later versions of the interface code?  Or is it an actual different choice of functions?

python/desc/imsim/ImageSimulator.py
jchiang87: Yes.  The sims_GalSimInterface code composes the output filenames given nameRoot not the actual filename, so output_file can't be used here.

PR #192, opened 2018-11-25
add random seeds to headers; set airmass in the obs_md object so that…

jchiang87:
… it can be set in galSimDetector code

### DIFF
--- a/python/desc/imsim/camera_readout.py
+++ b/python/desc/imsim/camera_readout.py
@@ -86,6 +86,8 @@ class ImageSource(object):
         self.sensor_id = sensor_id
         self.visit = visit
 
+        self._seed = None
+
         self.camera_info = CameraInfo()
 
         self._make_amp_images()
@@ -288,10 +290,7 @@ class ImageSource(object):
         # Add dark current.
         dark_current = config['electronics_readout']['dark_current']
         imaging_arr = imaging_segment.getArray()
-        # Generate a seed from the visit number and sensor_id so that
-        # the dark current noise is deterministic.
-        seed = CosmicRays.generate_seed(self.visit, self.sensor_id)
-        rng = galsim.PoissonDeviate(seed, dark_current*self.exptime)
+        rng = galsim.PoissonDeviate(self.seed, dark_current*self.exptime)
         dc_data = np.zeros(np.prod(imaging_arr.shape))
         rng.generate(dc_data)
         imaging_arr += dc_data.reshape(imaging_arr.shape)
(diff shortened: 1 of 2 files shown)


python/desc/imsim/camera_readout.py
```diff
@@ -477,6 +483,8 @@ def write_fits_file(self, outfile, overwrite=True, run_number=None,
         output[0].header['RATEL'] = self.ratel
         output[0].header['DECTEL'] = self.dectel
         output[0].header['ROTANGLE'] = self.rotangle
+        # Write the seed used by the read noise and dark current.
+        output[0].header['RN_SEED'] = self.seed
```
cwwalter: I wonder if we should somehow label all imSim specific headers with some sort of prefix?  IMSIM_ ?  But, I'm not sure of the length constraints etc.

python/desc/imsim/imSim.py
```diff
@@ -830,14 +831,17 @@ def add_cosmic_rays(gs_interpreter, phot_params):
...
     exptime = phot_params.nexp*phot_params.exptime
     for name, image in gs_interpreter.detectorImages.items():
         imarr = copy.deepcopy(image.array)
         # Set the random number seed for painting the CRs.
-        crs.set_seed(CosmicRays.generate_seed(visit, name))
+        cr_seed = CosmicRays.generate_seed(visit, name)
+        crs.set_seed(cr_seed)
         gs_interpreter.detectorImages[name] = \
             galsim.Image(crs.paint(imarr, exptime=exptime), wcs=image.wcs)
+        image.wcs.fitsHeader.set('CR_SEED', cr_seed)
```
cwwalter: Oops sorry missed this file... Adding as a separate comment.

So one thing I am a bit confused by.  In the seeds above for noise etc they went into the amp file (which makes sense).

By using wcs.fitsHeader.set does that mean it will go into the eimage + amp?

I'm also wondering about RATEL, DECTEL, ROTANGLE above which (I think) only goes to the amp file.  Should those go to both?

cwwalter:
BTW: Should we also add the imSim version # here or in another PR?

python/desc/imsim/imSim.py
cwwalter: OK sorry... I remember/see how this works now. It is grabbing RATEL etc from the header in the GSobject but there is a way to grab it from the opsim database if the info is missing in the old files.

jchiang87:
The version info should be generated automatically (i.e., not updated by hand in some file).  The Stack code can do this, but someone will have to figure out how to get it to work here.

cwwalter:
> The version info should be generated automatically (i.e., not updated by hand in some file). The Stack code can do this, but someone will have to figure out how to get it to work here.

I was thinking maybe the CI system could write a file with the version info or the scons 'make' could do it (that is the sort of way I have seen this done before).  Do we have any templates for that in scons?  That way the output would have the provenance independent of what processing was done on it.

This is the kind of thing of was thinking of:

https://stackoverflow.com/a/12368262/4881970

(but that is for make so, I'm not sure if we could do it our system when we go from bin.src to bin)

PR #193, opened 2018-11-28
Update README following DESC guidelines.

cwwalter:
Hi Rachel,

Please take a look at this new version of the README as you requested.

BTW, I think found an issue in the DESC instructions.  If a code is open source (and as specified by our License file) others are free to use it and must include our copyright notice etc.  There aren't any other requirements.  Yet the file you showed me suggests we include:

"We welcome requests to access the code for non-DESC use; if you wish to use the code outside DESC please contact the developers."

I have left this out as it only makes sense for private not open source repos.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,21 @@
 [![Coverage Status](https://coveralls.io/repos/github/LSSTDESC/imSim/badge.svg?branch=master)](https://coveralls.io/github/LSSTDESC/imSim?branch=master)
 
 # imSim
-GalSim based LSST simulation package
+imSim is a software package that simulates the LSST telescope and survey.
+It produces simulated images from the 3.25 Gigapixel camera which are suitable
+to be processed through the LSST Data Management pipeline. imSim takes as an
+input a catalog of astronomical sources along with information about how the light
+is distorted on the way to Earth including lensing and extinction information.  The
+images which are produced include the the systematic effects of the atmosphere,
+optics and sensors on the observed PSF.
+
+imSim calls the [GalSim](https://github.com/GalSim-developers/GalSim "GalSim GitHub Page") library for astronomical object rendering and is designed to be used inside of the
+LSST and LSST Simulation Group software environment.  It is not a stand alone
+program.  It requires a working LSST software stack to build and run.
+
+## imSim models
+
+Details on the models implemented by imSim including their validations can be found in the [imSim Feature Matrix](https://github.com/LSSTDESC/imSim/wiki/Feature-Matrix).
 
 ## Set up
 Set up the `lsst_sims` distribution, then enter the following from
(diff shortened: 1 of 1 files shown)


rmjarvis:
I think that line you left out is directed at other DESC members.  Internally, our pub guidelines say that DESC members who want to use unpublished DESC resources for non-DESC purposes need to get special clearance.  So it's not a legal requirement (as you point out re the license), but a self-imposed collaboration obligation.

cwwalter:
> (But the current wording doesn't make that clear, so probably some reworking of that line is in order.)

Hmm... Yes... currently it looks like this:

screen shot 2018-11-28 at 11 55 00 am

Maybe it would be better to split it and make it clear when to add that part.

README.md
```diff
@@ -74,6 +92,16 @@ optional arguments:
...
-                        prepended to any existing IMSIM_IMAGE_PATHenvironment
-                        variable, for which $CWD is included bydefault.
+                        prepended to any existing IMSIM_IMAGE_PATH environment
+                        variable, for which $CWD is included by default.
 ```
+
+## Release Status
+The list of released versions of this package can be found [here](https://github.com/LSSTDESC/imSim/releases), with the master branch
+including the most recent (non-released) development.  imSim is currently
+still considered beta code.
```
rmandelb: This paragraph could use some clarifying, I think.  At least in the repo guidelines, we had suggested referring to releases as versions that are intended for general use.  If you're saying it's beta code, then it seems like those are tagged versions but not releases?  If I follow the link, then you appear to have tagged versions and pre-releases (which I think of as another name for a tagged version of beta code) but no releases at all.  So one possible rewording might be "imSim is currently considered beta code until its v1 release, which will be accompanied by a paper that should be cited by those who use the released version in published work.  A list of tagged versions and pre-releases of this package can be found [here](https://github.com/LSSTDESC/imSim/releases), with the master branch including the most recent development beyond the latest pre-release."  - assuming that is indeed what you mean here.

cwwalter: I've made the requested changes here and above.  

For the above point there is a really confusing thing that GH does that is good to appreciate I thought I would write down for people.  You can see their explanation of it here:

https://github.community/t5/How-to-use-Git-and-GitHub/Tag-without-release/td-p/6255

Basically releases happen when you add release notes.  And if you use the API you can get them.  They show up on the "release tab".  *But*, all code tags show up there too.  So if you tag the code without a release it shows up on that page. It has a sort of different style.  At the very bottom of the imSim release page you can see a tag.  The others actually are releases made by Jim but he checked a box "pre-release" when he did it to let people know it is still non-production code.

I made it clearer that this is still considered pre-release code.

rmandelb:
Thanks for this, Chris!

Issue #178, opened 2018-10-10
Determine list of image metadata keywords that should be in the primary HDU


jchiang87:
This is to support jointcal processing. See https://github.com/LSSTDESC/DC2-production/issues/275

cwwalter:
Can we calculate this ourselves from pointing, or must we use the OpSim database? It's nice to not have to depend on having a database for standalone imSim usage.

jchiang87:
We have a [function](https://github.com/LSSTDESC/imSim/blob/master/python/desc/imsim/imSim.py#L826) to calculate it using the altitude info from the instance catalog (which also has an airmass entry that we don't use).

jmeyers314:
I'd add 
• parallactic angle
• temperature
• pressure
• water-vapor partial pressure
to the list if they're available, as these all also affect DCR.

cwwalter:
Thank you Josh.  

Others interested in metadata that they think is necessary please feel free to add to this thread.  I did some research and talked to some of the DM group working on this issue (especially @ktlim).  Here is a bit of what I learned:

- KT and others are working on making such a list of headers that should be included taking inputs from many stake holders.  I have asked for us to also give input.  It will eventually exist in LSE-400.  So this is a good place to gather thoughts you would like people working on this to see.

My current rough understanding of the architecture is as follows and architecture experts may correct or expand this:

- Keywords can be added to files (or to databases that can later be accessed) in more than one way and at different times.  The camera control system (CSS) can write a FITS file with internal camera information and also provide data to the header service (along with other slow-control systems like those for the telescope and environmental conditions). The header service will provide a header object, which will be passed to the archiver service which which can then add them to the files.

- Additionally other pieces of metadata may be added or corrected after the fact.  The resulting data structure (sometimes referred to as "extended raw") will be a  mixture of original headers and updated metadata as stored in a database and retrieved via the butler.  This is described in more detail here: https://confluence.lsstcorp.org/display/DM/Image+Metadata+Flow#ImageMetadataFlow-TelemetryandEventData  .  For example:

```
In the process of generating Calibration Products or preparing for a
Data Release Production, new values for metadata items may be
generated from time to time as our understanding of them improves.
For example, we may update the conversion from voltage to physical
units for a telemetry item, change the method of interpolation for a
slowly-changing item, or change the method of averaging for a
rapidly-changing item.  Some simply incorrect values caused by
hardware or software failure may be replaced.  All changes of this
type will result in new versions of the image metadata in the raw
[...]

PR #194, opened 2018-12-04
add header keywords for jointcal

jchiang87:
These changes add keywords mentioned at https://github.com/LSSTDESC/imSim/issues/178#issuecomment-443814647

@cwwalter It would be good to review this right away so that it can be included in the tag for Run2.1i.

### DIFF
--- a/python/desc/imsim/camera_info.py
+++ b/python/desc/imsim/camera_info.py
@@ -98,5 +98,7 @@ def getHourAngle(observatory, mjd, ra):
                              location=observatory.getLocation())
     # Get the local apparent sidereal time.
     last = time.sidereal_time('apparent').degree
-    ha = last - ra
+    ha = (last - ra) % 360.
+    if ha > 180:
+        ha -= 360.
     return ha
--- a/python/desc/imsim/camera_info.py
+++ b/python/desc/imsim/camera_info.py
@@ -1,10 +1,11 @@
 """
 Class to encapsulate info from lsst.obs.lsst.imsim.ImsimMapper().camera.
 """
+import astropy.time
 import lsst.afw.geom as afw_geom
 from lsst.obs.lsst.imsim import ImsimMapper
 
-__all__ = ['CameraInfo']
+__all__ = ['CameraInfo', 'getHourAngle']
 
 
 class CameraInfo:
@@ -73,3 +74,29 @@ class CameraInfo:
         ymin = 0 if yseg == 1 else height
         return afw_geom.Box2I(afw_geom.Point2I(xmin, ymin),
                               afw_geom.Extent2I(width, height))
+
+
+def getHourAngle(observatory, mjd, ra):
+    """
+    Compute the local hour angle of an object for the specified
+    MJD and RA.
+
+    Parameters
+    ----------
+    observatory: lsst.sims.GalSimInterface.LsstObservatory
+        Extension of lsst.afw.coord.Observatory object.
+    mjd: float
+        Modified Julian Date of the observation.
+    ra: float
+        Right Ascension (in degrees) of the object.
+
+    Returns
+    -------
+    float: hour angle in degrees
+    """
+    time = astropy.time.Time(mjd, format='mjd',
+                             location=observatory.getLocation())
+    # Get the local apparent sidereal time.
+    last = time.sidereal_time('apparent').degree
+    ha = last - ra
+    return ha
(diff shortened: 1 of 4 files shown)


python/desc/imsim/camera_info.py
```diff
@@ -73,3 +74,29 @@ def mosaic_section(amp_info):
...
+
+    Returns
+    -------
+    float: hour angle in degrees
+    """
+    time = astropy.time.Time(mjd, format='mjd',
+                             location=observatory.getLocation())
+    # Get the local apparent sidereal time.
+    last = time.sidereal_time('apparent').degree
+    ha = last - ra
```
cwwalter: Are we allowing negative HA, or are we going with 'If result is negative, add 360 degrees. If result is greater than 360, subtract 360 degrees.'?

python/desc/imsim/camera_readout.py
```diff
@@ -467,11 +469,11 @@ def write_fits_file(self, outfile, overwrite=True, run_number=None,
             run_number = self.visit
         output[0].header['RUNNUM'] = str(run_number)
         output[0].header['DARKTIME'] = output[0].header['EXPTIME']
-        mjd_obs = astropy.time.Time(output[0].header['MJD-OBS'], format='mjd')
-        output[0].header['DATE-OBS'] = mjd_obs.isot
+        output[0].header['TIMESYS'] = 'TAI'
         output[0].header['LSST_NUM'] = lsst_num
         output[0].header['TESTTYPE'] = 'IMSIM'
         output[0].header['IMGTYPE'] = 'SKYEXP'
+        output[0].header['OBSTYPE'] = 'SKYEXP'
```
cwwalter: Do we want to make this one always mirror IMGTYPE? Or allow them to diverge?

python/desc/imsim/camera_readout.py
```diff
@@ -483,6 +485,28 @@ def write_fits_file(self, outfile, overwrite=True, run_number=None,
         output[0].header['RATEL'] = self.ratel
         output[0].header['DECTEL'] = self.dectel
         output[0].header['ROTANGLE'] = self.rotangle
+
+        # Add various keywords needed for jointcal, including  hour angle
+        # and airmass values at the start and end of the observation.
+        mjd_obs = output[0].header['MJD-OBS']
+        mjd_end = astropy.time.Time(output[0].header['DATE-END'], format='isot',
```
cwwalter: Does the form of this format matter to us (isot)?

python/desc/imsim/camera_info.py
rmjarvis: Traditionally HA is given between -180 and +180.  (If given in degrees.  More traditionally, it would be given in hours.)

python/desc/imsim/camera_info.py
```diff
@@ -73,3 +74,29 @@ def mosaic_section(amp_info):
...
+        Right Ascension (in degrees) of the object.
+
+    Returns
+    -------
+    float: hour angle in degrees
+    """
+    time = astropy.time.Time(mjd, format='mjd',
+                             location=observatory.getLocation())
+    # Get the local apparent sidereal time.
+    last = time.sidereal_time('apparent').degree
```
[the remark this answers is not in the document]
jchiang87: I think so.

python/desc/imsim/camera_info.py
jchiang87: This is the code lifted from `sims_GalSimInterface` (we can refactor later).  I originally had that code make sure `ha` was in the range (0, 360), but people complained that that wasn't necessary.  I don't think it's worth revisiting this issue given the previous discussion on this exact point.

python/desc/imsim/camera_readout.py
jchiang87: I think they will always be the same.  I didn't want to just summarily get rid of `IMGTYPE` since I think there is code that still looks for that keyword to be set, e.g., when making calibration products.  The scripts that generate those files will need to add `OBSTYPE`.

python/desc/imsim/camera_info.py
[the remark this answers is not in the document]
rmjarvis: Yes.

python/desc/imsim/camera_readout.py
cwwalter: I was  wondering if (as long as it still exists) to be defensive if we should do 

output[0].header['OBSTYPE'] = output[0].header['IMGTYPE']

to stop someone from accidentally just changing only one in the future.

timj: If we think OBSTYPE is going to be around longer shouldn't we set IMGTYPE from OBSTYPE?

PR #197, opened 2019-01-12
Add gaussian psf to AtmosphericPSF

jmeyers314:
This PR follows from the discussion on slack channel #desc-dm-dc2 around Jan 9-11, 2019.  It adds an option to convolve in an additional Gaussian PSF to the AtmosphericPSF model to account for otherwise unmodeled effects, such as dome seeing, polishing errors, vibration, and other errors listed in document LTS-124.

I also took the opportunity to switch from using `fsolve` to invert the Tokovinin fitting formula for r0_500 to `bisect`, which I've found to be more robust.  I also changed the code to target the seeing at the specific filter effective wavelength being drawn instead of the seeing at 500 nm.  These two changes are mostly aesthetic, yielding delivered FWHM differences around ~0.01 arcseconds.

### DIFF
--- a/python/desc/imsim/atmPSF.py
+++ b/python/desc/imsim/atmPSF.py
@@ -3,7 +3,7 @@ GalSim realistic atmospheric PSF class
 """
 
 import numpy as np
-from scipy.optimize import fsolve
+from scipy.optimize import bisect
 
 import galsim
 from lsst.sims.GalSimInterface import PSFbase
@@ -102,7 +102,11 @@ class AtmosphericPSF(PSFbase):
         r0_500 = self.atm.r0_500_effective
         r0 = r0_500 * (self.wlen_eff/500.0)**(6./5)
         kmax = kcrit / r0
+        if logger:
+            logger.info("Building atmosphere")
         self.atm.instantiate(kmax=kmax, check='phot')
+        if logger:
+            logger.info("Finished building atmosphere")
 
         if doOpt:
             self.atm.append(OptWF(rng, self.wlen_eff))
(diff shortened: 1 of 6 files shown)


python/desc/imsim/atmPSF.py
```diff
@@ -200,4 +228,9 @@ def _getPSF(self, xPupil=None, yPupil=None, gsparams=None):
             t0=self.t0,
             exptime=self.exptime,
             gsparams=gsparams)
+        if self.gaussianFWHM > 0.0:
+            psf = galsim.Convolve(
```
jchiang87: @jmeyers314 @rmjarvis Unfortunately, doing this convolution breaks this code:  https://github.com/lsst/sims_GalSimInterface/blob/master/python/lsst/sims/GalSimInterface/galSimInterpreter.py#L1006
Could you guys have a look please?

rmjarvis: Ah, right.  I forgot we got rid of a feature we used to have where if any argument to a Convolve was itself a Convolution, it would combine the lists.  I guess we could do that manually in `PSFbase.applyPSF`:
```
if isinstance(psf, galsim.Convolution):
    return galsim.Convolution([obj] + psf.obj_list)
else:
    return galsim.Convolution(obj, psf)
```

jchiang87: Thanks, making that change and swapping the order of the items in the `Convolve` in `atmPSF._getPSF` addresses the issue.   I'll go ahead and make the PR in sims_GalSimInterface.

jchiang87:
@jmeyers314 @cwwalter   I think this is ready to merge.  I added some code to read the `gaussianFWHM` value for the additional Gaussian to convolve from the config file.   I added a unit test to ensure that these changes are working as expected, but if you guys want to have a look, please do so.  Unless I hear otherwise, I'll go ahead and merge this into master Monday morning.

PR #198, opened 2019-01-16
minor patches for header keyword handling

jchiang87:


jchiang87:
I'll merge this at 9am PT today unless I hear otherwise.

PR #211, opened 2019-03-04
PSF config handling and checkpoint filename fixes for Run2.1i

jchiang87:


jchiang87:
@villarrealas There are a couple changes to the checkpoint filenames to note:
* Checkpoint files now have names of the form `checkpoint-<file_id>-Rxx_Syy.cpkt`, instead of `checkpoint-<file_id>-R_x_x_S_y_y.ckpt`.
* Checkpoint summary sqlite files have names of the form `ckpt_<file_id>_<node_id>.sqlite3` instead of `ckpt_<visit>_<node_id>.sqlite3`.

python/desc/imsim/camera_readout.py
```diff
@@ -469,8 +469,12 @@ def write_fits_file(self, outfile, overwrite=True, run_number=None,
             Run number.  If None, then the visit number is used.
         compress: bool [True]
             Use RICE_1 compression for each image HDU.
-        image_type str ['SKYEXP']
+        image_type: str ['SKYEXP']
             Image type to write to the `OBSTYPE` and `IMGTYPE` keywords.
+        added_keywords: dict [None]
+            Python dict of key/value pairs to add to the primary HDU.
+            These will be set just before writing the FITS file, so
+            they will override any existing keywords.
```
cwwalter: maybe add:

'they will override any existing keywords or be added to the header list if they don't exist.'

python/desc/imsim/ImageSimulator.py
```diff
@@ -239,9 +240,8 @@ def run(self, processes=1, wait_time=None, node_id=0):
         global CHECKPOINT_SUMMARY
         if (self.file_id is not None and processes > 1 and
             CHECKPOINT_SUMMARY is None):
-            visit = self.obs_md.OpsimMetaData['obshistID']
-            CHECKPOINT_SUMMARY \
-                = CheckpointSummary(db_file='ckpt_{}_{}.sqlite3'.format(visit, node_id))
+            db_file = 'ckpt_{}_{}.sqlite3'.format(self.file_id, node_id)
```
cwwalter: What's the rational for using the file_id command line parameter here instead of the visit_id as before (and same question for the the checkpoint)?

python/desc/imsim/imSim.py
```diff
@@ -763,7 +764,10 @@ def read_config(config_file=None):
     if config_file is None:
         config_file = os.path.join(lsstUtils.getPackageDir('imsim'),
                                    'data', 'default_imsim_configs')
-    cp.read(config_file)
+
+    if not cp.read(config_file):
+        raise FileNotFoundError("Config file {} not found".format(config_file))
+
```
cwwalter: GitHub isn't letting me comment on the line since it hasn't changed, but below here (on line 776) you are still calling read_config (right after the function definition).  But you are now doing this on line 57 in bin/imSim.py.

Will this wipe out any parameters that were changed programmatically since then?

data/default_imsim_configs
```diff
@@ -50,4 +50,4 @@ sort_magnorm = True
 # FWHM in arcsec of the Gaussian to convolve with the baseline
 # AtmosphericPSF + OptWF PSF to account for additional instrumental
 # effects.
-gaussianFWHM = 0.3
+gaussianFWHM = 0.4
```
cwwalter: Do we have the results in a histogram from the multiple visits you ran of ellipticity etc from Javier yet, to know if this is the value we want to use?

python/desc/imsim/ImageSimulator.py
```diff
@@ -257,8 +257,8 @@ def run(self, processes=1, wait_time=None, node_id=0):
         pool = multiprocessing.Pool(processes=processes)
         receivers = []
         if wait_time is not None:
-            results.append(pool.apply_async(process_monitor, (),
-                                            dict(wait_time=wait_time)))
+            results.append(pool.apply_async(TracebackDecorator(process_monitor),
```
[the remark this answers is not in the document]
jchiang87: "Decorator" has an established meaning as a standard design pattern, which this class satisfies.  "Wrapper" also has an established design pattern meaning, which does not fit this class, so I think the current class name is the right one.

python/desc/imsim/ImageSimulator.py
jchiang87: `visit_id` was too restrictive.  If I want to put additional information in the checkpoint summary file name (e.g., the filter), then `file_id` is the better way to do it.

python/desc/imsim/imSim.py
jchiang87: The `read_config` here picks up the defaults so that I can pop into ipython and not have to call `read_config` explicitly if I want to do something simple, like read the observing parameters from an instance catalog or construct an obs_md object from it.   The `read_config` call at https://github.com/LSSTDESC/imSim/blob/u/jchiang/read_config_fix/bin.src/imsim.py#L57 ensures that any user provided config is read in and will be used when the code is executed.

data/default_imsim_configs
jchiang87: We had an extensive discussion on this at XWG computing group session on Wednesday of the DESC meeting, and everyone agreed that this was the value we should use.

python/desc/imsim/camera_readout.py
jchiang87: The name of the option indicates that the keywords will be added and the first sentence of the description also says that they will be added.  I think it's probably ok as-is.

PR #215, opened 2019-04-07
add .fits_atomic_write method to avoid partially written FITS files

jchiang87:


cwwalter:
Look good to me.

Notes:

- It seems like write_amplifier_image() (where a change was made in the PR) may be a vestigial function that could be removed.

- It might be nice to do the same thing for the e-images.  But, that write lives in galSimInterface and that code will probably be subsumed into imSim during the refactoring.  So, we can do it then.

PR #217, opened 2019-05-29
share an LsstObservatory object among all GalSimInterpreter objects

jchiang87:


python/desc/imsim/ImageSimulator.py
```diff
@@ -150,6 +154,7 @@ def _make_gs_interpreters(self, seed, sensor_list, file_id):
                                       apply_sensor_model=self.apply_sensor_model,
                                       bf_strength=self.config['ccd']['bf_strength'])
 
+            self.gs_interpreters[det_name]._observatory = self.lsstObservatory
```
rmjarvis: Rather than this, you might consider using a class attribute instead.

```
GalsimInterpreter._observatory = self.lsstObservatory
```
Then you only have to do it once and all GalSimInterpreter instances will have it.  I don't know whether there are any places where a new GalSimInterpreter instance is made on the fly, where they would miss out on this setup step, but if so, using the class attribute would still work for them.

jchiang87: Unfortunately, the `._observatory` instance attribute is explicitly set to `None` in the `GalSimInterpreter` initializer, so we'd need to update the `sim_GalSimInterface` code to omit that line, among other changes, in order for setting the class attribute like this to work.

PR #219, opened 2019-06-06
re-insert the WCSAXES keyword to precede other WCS keywords

jchiang87:


jchiang87:
I would have hoped that the line
```
output[0].header = self.eimage[0].header
```
would just copy the keywords in the same order as in the eimage, but perhaps because the `.header` acts like an unordered `dict`, the keywords end up in mostly random order.   If I explicitly loop through the keywords in the `self.eimage[0].header`, they are added in the correct order, but astropy can't write the file for some reason.  I didn't want to debug that, so deleting and inserting it at a location that was guaranteed to appear before the other WCS keywords was the most expedient solution to address something we shouldn't have to spend any time on at all anyways.

cwwalter:
Weird.  OK, I understand, thanks.

cwwalter:
I guess it might be good to write something in the comments there that says the reason it is being done is that the 

output[0].header = self.eimage[0].header

is randomizing the order so someone looking at it knows why it is needs to be re-inserted.

PR #227, opened 2019-08-26
Issue 201/use generator for gs objects

jchiang87:
These changes save at least 0.5 GB per sensor.   Here are the memory profile curves for a 2-sensor simulation before these changes:
master_mem_profile
and the same simulation after
issue_201_mem_profile_2
The relevant lines in these plots are the solid blue line, which shows the shared memory among all three processes (parent + 2 subprocess for the individual sensor-visits), and the dotted brown and red lines, which are the unique set size memory for the 2 sensor-visits, which are lower by 0.5GB in the lower plot versus the upper.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,7 @@ dist: trusty
 compiler: gcc
 
 env:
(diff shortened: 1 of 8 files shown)


jchiang87:
Running on Cori-KNL with different numbers of sensors in parallel:
23_40_46_50_sensors
The 46 and 50 sensor cases died with out-of-memory errors.

PR #228, opened 2019-09-18
Add instructions to the example directory for running examples.

cwwalter:
Add a readme to the example directory with simple example instructions.


cwwalter:
I think we can just merge it. I think we were waiting because @jmeyers314 said he was going to take a peek.

PR #230, opened 2019-09-25
Issue 229/parallelize atm screens

jchiang87:
@jmeyers314 I haven't tested this yet on a full resolution set of screens to see how well it works, but it seems to work in the unit tests.  Could you have a quick look to see if everything make sense?

### DIFF
--- a/python/desc/imsim/atmPSF.py
+++ b/python/desc/imsim/atmPSF.py
@@ -105,7 +105,7 @@ class AtmosphericPSF(PSFbase):
         self.logger = logger
 
         ctx = multiprocessing.get_context('spawn')
-        self.atm = galsim.Atmosphere(**self._getAtmKwargs(), mp_context=ctx)
+        self.atm = galsim.Atmosphere(mp_context=ctx, **self._getAtmKwargs())
         self.aper = galsim.Aperture(diam=8.36, obscuration=0.61,
                                     lam=self.wlen_eff, screen_list=self.atm)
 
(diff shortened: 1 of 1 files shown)


jchiang87:
Memory profiles for 1, 4, and 8 processes run in parallel to generate the screens, run on Cori-Haswell:
atm_psf_mp_timing

python/desc/imsim/atmPSF.py
```diff
@@ -101,7 +104,8 @@ def __init__(self, airmass, rawSeeing, band, rng,
         self.screen_scale = screen_scale
         self.logger = logger
 
-        self.atm = galsim.Atmosphere(**self._getAtmKwargs())
+        ctx = multiprocessing.get_context('fork')
```
jmeyers314: I *think* 'spawn' here is safer than 'fork'.  See this [blog post](https://codewithoutrules.com/2018/09/04/python-multiprocessing/), for example.  That said, I don't have any actual experience with problems using either launch context, so if this works in tests, it's probably fine.

jchiang87: It does sound safer based on that blog post, but when I used `spawn`, the python environment was getting imported from scratch multiple times--in fact many more times than just the number of processes specified.  This was slowing down the test code past my limited patience and was also outputting a lot of DM info messages.  However, when I increased the number of processes to 16, the calculation died in a way I didn't understand, so maybe `fork` is at issue here.

jchiang87: fwiw, I did a run last night to check that the full RandomKnots instance catalogs would work for CCDs that succeeded and failed in Antonio's Theta testing, and there was no problem using nproc=16.  But since there are only 6 screens being computed, I assume this is unnecessary?  @jmeyers314 How many processes would you recommend for the `self.atm.instantiate(...)` call?

jmeyers314: >the python environment was getting imported from scratch multiple times--in fact many more times than just the number of processes specified.

I see, so maybe safer but slower.  It's certainly weird that the imports are more frequent than the number of processes though.

>How many processes would you recommend for the self.atm.instantiate(...) call?

There's no advantage to using more than the number of screens that I can see.

jchiang87: I updated the code to set the number of processes to be the number of atmospheric phase screens if `nproc` is not otherwise set.

> It's certainly weird that the imports are more numerous than the number of processes though.

I was mistaken about this.  I didn't realize `test_atmPSF.py` was creating 100 different `AtmosphericPSF` objects.

jchiang87:
Unless I hear otherwise, I will merge this ~5pm PT today.

PR #232, opened 2019-10-01
write selected LSST Stack package versions and tags to FITS primary header

jchiang87:


jchiang87:
@cwwalter This is ready to go.

jchiang87:
Unfortunately, getting this info into the eimage is difficult since the FITS file writing is done by the galsim `Image.write` function and there's no direct interface to manipulating the header keywords before writing.  If we really want this, we'd need to update the file after it is first written.

cwwalter:
> Unfortunately, getting this info into the eimage is difficult since the FITS file writing is done by the galsim `Image.write` function and there's no direct interface to manipulating the header keywords before writing. If we really want this, we'd need to update the file after it is first written.

Ah, OK; too bad. Actually I think you already told me this in the past.

rmjarvis:
Actually, anything you put into `image.header` should show up in the written FITS header.

rmjarvis:
I don't think there's particularly good documentation for this feature, but this unit test should show you the syntax if you want to try it.

https://github.com/GalSim-developers/GalSim/blob/releases/2.2/tests/test_image.py#L2412

jchiang87:
Thanks, Mike.   I've updated the code to add the version keywords to the eimage files following the galsim example, and I refactored the version keyword handling so that we can test some of that functionality directly.

Unless I hear otherwise, I will merge this into master at 5pm PT today, and will tag the imSim release for Run2.2i.

Issue #234, opened 2019-10-31
loading an `AtmosphericPSF` from a pickle file no longer works


jchiang87:
@jmeyers314 Using galsim v2.2.1, the mechanism we had for saving an `AtmosphericPSF` to a pickle file then reading it in for a later `imsim.py` run no longer works.   Here's a self-contained script that shows the problem:
```
import subprocess
import desc.imsim

commands = {'rightascension': 61.99398718973142,
 'declination': -32.83695608408115,
 'mjd': 60278.28041661111,
 'altitude': 58.81635120206924,
 'azimuth': 255.71142857124866,
 'filter': 3,
 'rotskypos': 265.809115420881,
 'camconfig': 1,
 'dist2moon': 74.5977298,
 'moonalt': 30.9167693,
 'moondec': 28.6386564,
 'moonphase': 90.8750305,
 'moonra': 104.4448764,
 'nsnap': 1,
 'obshistid': 479028,
 'rottelpos': 1.1543954,
 'seed': 479028,
 'seeing': 0.512611,
 'sunalt': -29.2091162,
 'vistime': 30,
 'minsource': 100,
 'bandpass': 'i'}
obs_md = desc.imsim.phosim_obs_metadata(commands)

psf = desc.imsim.make_psf('Atmospheric', obs_md, #log_level='DEBUG',
                          screen_scale=6.4)

psf_file = 'atm_psf.pkl'
desc.imsim.save_psf(psf, psf_file)

# This works:
desc.imsim.load_psf(psf_file)

# but loading in a new python session results in a KeyError:
subprocess.check_call(f'''python -c "import desc.imsim; desc.imsim.load_psf('{psf_file}')"''', shell=True)
```
Here's the output I get:
```
CameraMapper INFO: Loading Posix exposure registry from .
CameraMapper INFO: Loading Posix exposure registry from .
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/global/u1/j/jchiang8/dev/imSim/python/desc/imsim/imSim.py", line 873, in load_psf
    psf = pickle.load(fd)
  File "/opt/lsst/software/stack/stack/miniconda3-4.5.12-1172c30/Linux64/galsim/2.2.1.lsst/lib/python/galsim/phase_screens.py", line 298, in __setstate__
    self._objDict = _GSScreenShare[self._shareKey]
KeyError: 46914331211872
Traceback (most recent call last):
  File "make_atm_psf.py", line 38, in <module>
[...]

PR #235, opened 2019-11-01
Fix Atmospheric PSF pickling

jmeyers314:
Addresses #234 .

Serializing AtmosphericScreens now requires dumping the shared memory and then separately pickling.  I added a command line argument to imsim.py to handle the shared memory piece when the PSF type is Atmospheric.  Depends on https://github.com/GalSim-developers/GalSim/pull/1057

### DIFF
--- a/bin.src/imsim.py
+++ b/bin.src/imsim.py
@@ -45,6 +45,10 @@ parser.add_argument('--psf_file', type=str, default=None,
                     "If the file exists, the psf will be loaded from that "
                     "file, ignoring the --psf option; "
                     "if not, a PSF will be created and saved to that filename.")
+parser.add_argument('--psf_screen_file', type=str, default=None,
+                    help="File containing shared memory portion of "
+                    "AtmosphericPSF.  Required if reading/writing psf_file and "
+                    "the psf type is Atmospheric.")
 parser.add_argument('--image_path', type=str, default=None,
                     help="search path for FITS postage stamp images."
                     "This will be prepended to any existing IMSIM_IMAGE_PATH "
@@ -72,9 +76,10 @@ obs_md = desc.imsim.phosim_obs_metadata(commands)
 if args.psf_file is None or not os.path.isfile(args.psf_file):
     psf = desc.imsim.make_psf(args.psf, obs_md, log_level=args.log_level)
     if args.psf_file is not None:
-        desc.imsim.save_psf(psf, args.psf_file)
+        desc.imsim.save_psf(psf, args.psf_file, args.psf_screen_file)
 else:
-    psf = desc.imsim.load_psf(args.psf_file, log_level=args.log_level)
+    psf = desc.imsim.load_psf(args.psf_file, args.psf_screen_file,
+                              log_level=args.log_level)
 
 sensor_list = args.sensors.split('^') if args.sensors is not None \
     else args.sensors
--- a/python/desc/imsim/imSim.py
+++ b/python/desc/imsim/imSim.py
@@ -886,7 +886,7 @@ def make_psf(psf_name, obs_md, log_level='WARN', rng=None, **kwds):
                              logger=logger, **kwds)
     return psf
 
-def save_psf(psf, outfile):
+def save_psf(psf, outfile, screen_file=None):
     """
     Save the psf as a pickle file.
     """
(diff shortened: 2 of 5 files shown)


cwwalter:
We have to make sure we don't merge this until we are using a (not yet released GalSim 2.2) is that correct?

[Which makes me realize; I think Scott has been the one pushing those releases to the stack.  We need to make sure someone else has both the permissions and knowledge to do that.]

bin.src/imsim.py
```diff
@@ -72,9 +76,10 @@
 if args.psf_file is None or not os.path.isfile(args.psf_file):
     psf = desc.imsim.make_psf(args.psf, obs_md, log_level=args.log_level)
     if args.psf_file is not None:
-        desc.imsim.save_psf(psf, args.psf_file)
+        desc.imsim.save_psf(psf, args.psf_file, args.psf_screen_file)
 else:
-    psf = desc.imsim.load_psf(args.psf_file, log_level=args.log_level)
+    psf = desc.imsim.load_psf(args.psf_file, args.psf_screen_file,
```
rmjarvis: I'd recommend explicitly naming the kwarg here: `screen_file=args.psf_screen_file`.
IMO, it is good practice to use the kwarg name when an argument is optional, and limit positional arguments to required arguments with no default.  This tends to lead to less breakage if someday additional optional arguments are added in the future, which might mess up the ordering.

python/desc/imsim/imSim.py
```diff
@@ -6,7 +6,10 @@
 import sys
 import warnings
 from collections import namedtuple, defaultdict
-import pickle
+try:
+    import cPickle as pickle
+except:
+    import pickle
```
rmjarvis: FYI, I recommended Josh switch to this pattern.  AFAIU, cPickle tends to be equivalent functionality, but much faster.  So if it is available, it is better to use it.  (And it's basically always available unless you're using pypy I think.)
cf. https://pymotw.com/2/pickle/

python/desc/imsim/imSim.py
```diff
@@ -883,7 +886,7 @@ def make_psf(psf_name, obs_md, log_level='WARN', rng=None, **kwds):
                              logger=logger, **kwds)
     return psf
 
-def save_psf(psf, outfile):
+def save_psf(psf, outfile, screen_file=None):
```
jchiang87: I was kind of hoping that we could get a picklable object from galsim containing the phase screen data and retain the `save_psf` and `load_psf` interface so the user doesn't have to keep track of two files for the atmospheric PSF, i.e., do something along the lines of 
```
def save_psf(psf, outfile, screen_file=None):
    """
    Save the psf as a pickle file.
    """
    # Set any logger attribute to None since loggers cannot be persisted.
    if hasattr(psf, 'logger'):
        psf.logger = None

    with open(outfile, 'wb') as output:
        if isinstance(psf, AtmosphericPSF):
            phase_screens = galsim.phase_screens.get_picklable()
            pickle.dump((psf, phase_screens), output)
        else:
            pickle.dump(psf, output)


def load_psf(psf_file, log_level='INFO'):
    """
    Load a psf from a pickle file.
    """
    with open(psf_file, 'rb') as fd:
        psf_data = pickle.load(fd)

    if len(psf_data) == 2:
        psf = psf_data[0]
        galsim.phase_screens.set_data(psf_data[1])
    else:
        psf = psf_data

    # Since save_psf sets any logger attribute to None, restore
    # it here.
    if hasattr(psf, 'logger'):
        psf.logger = get_logger(log_level, 'psf')

    return psf
```

jmeyers314:
Thanks @jchiang87 and @rmjarvis .  I re-attacked this one, please look again.  Much simpler solution now.

rmjarvis:
GalSim 2.2.2 is now released with the functionality necessary for this to work.

jchiang87:
@jmeyers314 @rmjarvis Thanks for making these changes to preserve the interface.  Unfortunately, I'm having trouble trying this out since `pip install galsim` isn't working in the docker image I use to run the lsst_sims code.  It's complaining about not finding the pybind11 header files.  Since we probably need galsim v2.2.2 included in lsst_sims anyways, we may need to wait until that happens before merging.

rmjarvis:
Maybe try `pip install pybind11` before installing galsim?  I've had trouble sometimes when pip is trying to install things for you.  Sometimes just explicitly installing the dependencies first helps.

cwwalter:
Just noticed this was still around and was curious about the status of it. We are just waiting until galsim 2.2.2 is released right?

( The latest version of the stack still have 2.2.1 but everything else is bumped up. So, it's possible it could be tested now without the pybind11 problems. But probably best just to wait for 2.2.2 if this is still the desired code)

rmjarvis:
The latest version is 2.2.3.  And it has been released since November...

cwwalter:
Yes... sorry... it was late last night.  Jim and I saw it was 2.2.3 when we discussed this last week but as far as we could tell it was only this issue that was relevant for us.  So, we decided we didn't have to upgrade right away and could wait on bumping the version in the stack.

Here I was basically asking if this code is still all just fine and we just need for the GaLsim version to bump before we merge it.

rmjarvis:
Yes.  I think this should work fine.  But requires >= v2.2.2.

PR #240, opened 2020-03-30
Modify imSim to be compatible with 2020 versions of the stack.

cwwalter:


cwwalter:
Argh! Hit return by mistake after only title!  

I don't see how to unpublished this.

Deleting this one.

rmjarvis:
FYI, you can edit the description, rather than delete and make a new one.  For next time...

PR #241, opened 2020-03-30
Modify imSim to be compatible with 2020 versions of the stack.

cwwalter:
These changes are required to make imSim compatible with recent versions of the stack.  Recent versions of the API have made it so that imSim no longer runs.  After this work is merged, imSim will no longer run with the older versions of the stack used for DC2 production.  To use those docker images etc, one should use the tagged versions of imSim created for the DC2 production. the changes to the AFW API are outlined in the this DM RFC: https://jira.lsstcorp.org/browse/RFC-585.

This work is a necessary prerequisite for new imSim development.

Trivial changes in this PR include:

- Removing various pieces of Python 2 Compatibility
- Some PEP8 spacing changes

The non trivial changes here are related to changes in the AFW API. There are two main relevant differences.  

- The way different detector sensors are specified is now changes.  Where before one would refer to (e.g.) WAVEFRONT. Now you should use DetectorType.WAVEFRONT.

- Rather than accessing AmpInfoRecord from the amplifier objects.  We can get the Amplifier objects themselves through iteration.

There are currently 3 unit tests that fail that we should discuss how and if we should fix:

1) test_skyModel.py

The test fails because it is comparing a hardcoded expected zeropoint with one that is no longer the same:

```
>       self.assertAlmostEqual(skycounts_persec_u.value, self.zp_u)
E       AssertionError: 0.5276285809019426 != 0.36626526294988776 within 7 places (0.16136331795205489 difference)
```

Experimentation has shown that this is due to an updating of the throughputs data (as copied from the PST) and not the sims_skymodel_data.  By manually checking out, declaring, and changing to the DC2Production tag of `throughputs`. I can make this test pass by producing the old number. [ This is, IMO,  an example where we don't have a good handle of these throughput constants.  (What version are we on now compared to before and what changed etc etc)]
 
I believe we have two options:

- Modify the test to work with the new value. We don't have a way to *verify* the value, but we can say the old inputs give the previous result.

- Modify this test to (somehow?) solely check the code itself and make it no longer dependent on its inputs.

2) test_get_stack_products.py

This test only fails if you declare your own EUPS products from local directories (as I needed to do to get the old results as described in the section above.  It does not seem robust to user formatted tags.

```
                self.assertEqual(products[target].version, version)
>               self.assertEqual(set(products[target].tags), tags)
E               AssertionError: Items in the first set but not the second:  'user:walter'
E               Items in the second set but not the first:   'walter'
```

We might decide not to change this one.

3) test_camera_readout.py

Fixing this unit test requires a more extensive change to code, and it is also due to the AFW API change.   The test_get_amp_image function calls: `desc.imsim.set_itl_bboxes(det['C03'])`

which in term calls `set_itl_bboxes` in camera_readout.py.  In those functions ` amp.setRawBBox` is called.   However, the new Amplifier object is now immutable. It is not possible to modify it. Rather we should tear it down and make a new one. 

So, to make this test pass we should rewrite the function to destroy and create new amplifiers.  At the same time I would change the variable nomenclature and documentation (including return types which are no longer the same etc)  for other functions in that file and camera_info.py.

However, it looks to me that the set_itl_bboxes is currently only called by the test and used to make amplifiers to test.  So, before doing this work I am interested in hearing from @jchiang87 whether this unitest (and the code it calls) should continue to exist, or if we should test things another way.

### DIFF
--- a/python/desc/imsim/ImageSimulator.py
+++ b/python/desc/imsim/ImageSimulator.py
@@ -13,7 +13,7 @@ import sqlite3
 import numpy as np
 from astropy._erfa import ErfaWarning
 import galsim
-from lsst.afw.cameraGeom import WAVEFRONT, GUIDER
+from lsst.afw.cameraGeom import DetectorType
 from lsst.sims.photUtils import BandpassDict
 from lsst.sims.GalSimInterface import make_galsim_detector
 from lsst.sims.GalSimInterface import make_gs_interpreter
@@ -154,7 +154,7 @@ class ImageSimulator:
             det_name = det.getName()
             if sensor_list is not None and det_name not in sensor_list:
                 continue
-            if det_type in (WAVEFRONT, GUIDER):
+            if det_type in (DetectorType.WAVEFRONT, DetectorType.GUIDER):
                 continue
             gs_det = make_galsim_detector(self.camera_wrapper, det_name,
                                           self.phot_params, self.obs_md)
@@ -190,7 +190,7 @@ class ImageSimulator:
     def _get_all_sensors(self):
         """Get a list of all of the science sensors."""
         return [det.getName() for det in self.camera_wrapper.camera
-                if det.getType() not in (WAVEFRONT, GUIDER)]
+                if det.getType() not in (DetectorType.WAVEFRONT, DetectorType.GUIDER)]
 
     @staticmethod
     def checkpoint_file(file_id, det_name):
--- a/python/desc/imsim/camera_info.py
+++ b/python/desc/imsim/camera_info.py
@@ -36,7 +36,7 @@ class CameraInfo:
 
     def get_amp_info(self, amp_name):
         """
-        Get the AmpInfoRecord object for the desired amplifier.
+        Get the Amplifier object for the desired amplifier.
 
         Parameters
         ----------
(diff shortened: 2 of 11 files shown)


rmjarvis:
Is it possible in the imsim redesign to extract ourselves from using the stack?  It's absolutely ridiculous that they *STILL* don't have a stable API.

jchiang87:
Here are my suggestions for handling the three unit test failures:

1. Looking at the code, we can address this by adding another `SkyModel` class that provides a standard SED and a bandpass object that provides a throughput curve that doesn't depend on external packages.  Then in the test code we can just use objects from those alternative implementations in the `SkyCountsPerSec` initializer.

2. In the near term, I don't think we need to fix this.

3. The `set_*_bboxes` functions were needed before `obs_lsst` existed and were used to define the pixel geometries of the amplifiers for the different vendors.  Since we now get that info from `obs_lsst`, we can just delete those functions.

So to resolve this PR, I suggest we just disable the `test_skycounts_function` (and do the development I describe above in a new branch), and remove the `set_*_bboxes` functions and the `test_get_amp_image` function.

cwwalter:
Hi @jchiang87, I think this is ready for review. 

As you suggested I:

- Put in a decorator to skip the sky test until it is refactored.
- Removed the get_amp_image test and accompanying helper functions that no longer work.

I also tried to update the docstrings and variable names to use new AFW API Amplifier class name.

PR #245, opened 2020-04-19
write FITS image when starting from a checkpoint and when there are n…

jchiang87:
…o objects left to render at restart

### DIFF
--- a/python/desc/imsim/ImageSimulator.py
+++ b/python/desc/imsim/ImageSimulator.py
@@ -276,9 +276,10 @@ class ImageSimulator:
         if wait_time is not None:
             results.append(pool.apply_async(TracebackDecorator(process_monitor),
                                             (), dict(wait_time=wait_time)))
-        for det_name in self.gs_interpreters:
+        for det_name, gs_interpreter in self.gs_interpreters.items():
             gs_objects = self.gs_obj_dict[det_name]
-            if self._outfiles_exist(det_name) or not gs_objects:
+            if (self._outfiles_exist(det_name) or
+                (not gs_objects and not gs_interpreter.drawn_objects)):
                 continue
 
             # If we are checkpointing, create the connections between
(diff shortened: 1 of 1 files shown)


jchiang87:
I'm not sure I understand the concern.  In the default configurations, the checkpoint file is only deleted (or moved) immediately **after** writing the FITS and centroid files for the associated sensor-visit.  So in normal operations, we can restart from the most recent checkpoint an arbitrary number of times without any  FITS file for that sensor-visit getting in the way.

For cases like Run3.0i, we have to set the config to move the final checkpoint to a separate directory, so any restart from those checkpoint files would require special handling, which I think would be the desired way of handling those sorts of cases.

cwwalter:
Thanks.  Sorry for the delay

I was wondering about 

https://github.com/LSSTDESC/imSim/blob/52fd8f21d54dbdd742fb556fefab51824cbb40fc/python/desc/imsim/ImageSimulator.py#L281

I think as long as the FITS file was moved away after a job stopped and before restarting again from a checkpoint it would work.  But, if the FITS file was still there, when it restarted wouldn't it see the file and decide that the sensor was already complete and skip simulating any new objects due to the 

if (self._outfiles_exist(det_name)

line?

jchiang87:
If the FITS file is there, then the checkpoint file would either be deleted or moved to a different location.  In either case, the fact that the FITS file is there indicates that that sensor visit finished so there would be no need to restart from a checkpoint.   If the job was stopped purposely somehow in the very short interval between writing the FITS file and deleting or moving aside the checkpoint file, then there'd still be no need to start from a checkpoint file since the sensor visit had finished.  The only case where we'd restart from a checkpoint after having written the FITS file would be cases where we've updated the instance catalogs as we're doing for Run3.0i, purposely redirected the checkpoint file. In that case, being forced to deal with the relocated checkpoint file and pointing the code to use the updated/altered instance catalogs is the right thing since otherwise there's a greater risk of mistakenly overwriting the FITS file from the original run, which is exactly what that line is meant to avoid.

cwwalter:
Ah.. OK I see what I was thinking.  I was looking at 
https://github.com/LSSTDESC/DESC_DC2_imSim_Workflow/pull/44

and then was reading the title "write FITS image when starting from a checkpoint".   After I went through all the code I was thinking through various scenarios and - from the title - imagined a case where a fits file was written to disk as soon as you started from the checkpoint.  So, in that case you could wind up with a checkpoint file and FITS file  around that might not get moved/removed.

But, of course as you say, they are only actually written at the end.  Thanks.

PR #248, opened 2020-08-24
Add configuration-file based implementation

rmjarvis:
This isn't completely done, but there is enough here that I think it's worth getting a code review of what I have so far.

So far this includes the following modules:
1. LSST_CCD output type.  This is the main driver of the imsim run.  It mostly just sets things up appropriately in terms of the detector/sensor names.
2. LSST_Silicon stamp type.  This is where the objects get drawn.  It's the main work horse of the simulation run.  It has most of the stuff that was in `galSimInterpreter.py` in the sims_GalSimInterface repo.
3. instance_catalog input type.  This reads in the instance catalog.
4. InstCatObj object type.  This makes a galaxy or star according to the specifications from the instance catalog.
5. InstCatSED SED type.  This makes an SED according to the instance catalog info.
6. InstCatWorldPos value type.  This returns the (RA,Dec) position according to the instance catalog info.
7. opsim_meta_dict input type.  This reads in just the meta-data part of an instance catalog.  (We need this first before reading in all the object information, so it was easier to just make it a separate input type.)
8. OpsimMeta value type.  This gets an item of meta information from the instance catalog.
9. OpsimMetaBandpass Bandpass type.  This reads in the right bandpass file according to the instance catalog meta information.
10. atm_psf input type.  This "input" type really just sets up the atmospheric screens to be used for the PSF across the focal plane.  It doesn't read in any input, but functionally it sits in the same role as an input object in terms of where it needs to do work in the simulation run.
11. AtmosphericPSF object type.  This gets a PSF object at any location using the atmospheric screens.
12. tree_rings input type.  This reads in the tree ring files we're using.
13. TreeRingFunc value type.  This gets the tree ring function from the input data.
14. TreeRingCenter value type.  This gets the tree ring center position from the input data.
15. readout extra output type.  This is minimally functional and mostly wrong probably.  But it has the hooks set up for how the readout code can be implemented as an extra output file.
16. Dict WCS type.  This is a place-holder for now.  It gets the WCS from a dictionary based on the DC2 WCS for the particular exposure I've been testing on.  I can probably work this up into something that recenters these WCS functions according to the boresight and focal plane rotation.  But for now it only works on this one exposure.  Longer term, we'll want a Batoid-based WCS type I think.

There are still quite a few TODOs and XXXs marking places where something is either not yet implemented or likely wrong.  I doubt this listing is complete, but I tried to mark what I knew about and noticed.


config/desc/imsim/instcat.py
```diff
@@ -0,0 +1,627 @@
...
+    # so we don't require the user to specify that by hand.  We can do it for them.
+    galsim.config.SetDefaultIndex(config, inst.getNObjects())
+
+    req = { 'index' : int }
+    opt = { 'num' : int }
+    kwargs, safe = galsim.config.GetAllParams(config, base, req=req, opt=opt)
+    index = kwargs['index']
+
+    rng = galsim.config.GetRNG(config, base, logger, 'InstCatObj')
+    bp = base['bandpass']
```
jchiang87, line 526: This line seems to be looking for a `bandpass` entry in the header part of the instance catalogs, but the current instance catalogs use `filter` (as an integer) to specify the band.   If I add a `bandpass u` entry to my instcat, I get a `ValueError` from the attempt to convert to `float` in line 427.

Mike: How does this work when you run?   I assume you've modified your instcat files, but how?

config/desc/imsim/instcat.py
```diff
@@ -0,0 +1,627 @@
...
+
+        # Note: params here starts at 12, so all indices are 12 less than in Jim's code.
+        if params[0].lower() == 'point':
+            obj = galsim.DeltaFunction(gsparams=gsparams)
+
+        elif params[0].lower() == 'sersic2d':
+            a = float(params[1])
+            b = float(params[2])
+            if b > a:
+                # Invalid, but Jim just let's it pass.
```
jchiang87, line 288: I think most of the comments in this file attributed to my work was actually done by Scott.  In any case, I suggest changing to something like `# Invalid, but the original imSim version just let it pass`

jchiang87 (changes requested):
@rmjarvis Looks good to me except for the comments I noted and the issue with the expectation for a `bandpass` entry in the instcat.

config/desc/imsim/instcat.py
jchiang87: Never mind about this one.  I was using an old version of the `imsim.yaml` file.  When I update to the current version, I don't encounter this problem.

rmjarvis: To clarify, it's looking for the 'bandpass' key in the base config dict, which is a fully constructed `galsim.Bandpass` object.

The path is:
1. This line converts "filter" in the instance catalog into a "band", which is just a single letter that things in the config file can use.
```
self.meta['band'] = 'ugrizy'[self.meta['filter']]
```
2. The config specifies the bandpass of the image as
```
bandpass: { type: OpsimMetaBandpass }
```
3. The OpsimMetaBandpass class builds the galsim.Bandpass as:
```
band = meta.get('band')
bandpass = galsim.Bandpass('LSST_%s.dat'%band, wave_type='nm')
bandpass = bandpass.withZeropoint('AB')
```
4. GalSim saves the current bandpass in use at the top level so anyone can easily access it.  So somewhere in the galsim code it saves the return value from OpsimMetaBandpass as `base['bandpass']`, which this line is then able to use.

config/desc/imsim/instcat.py
rmjarvis: Done.  Reworded all of these to refer to "previous code".

jchiang87 (approved):
lgmt

PR #250, opened 2021-01-04
update readout code so that the LSST Stack can process the raw file output

jchiang87:


config/imsim.yaml
```diff
@@ -115,8 +119,8 @@ image:
                 # the main process for all of the (many) vert faint sources, which only shoot
                 # a small numer of photons.
 
-    xsize: 4096
-    ysize: 4096
+    xsize: 4072   # These are the numbers for ITL sensors, but they should be
+    ysize: 4000   # found from the camera geometry individually for each CCD.
```
rmjarvis, line 123: We can definitely get these from the camera_geometry. 
I think a good fairly general solution is probably to add something like the `OpsimMeta` function that can grab an arbitrary attribute from the current detector's geometry.  Something like this I think (untested):
```
def CameraValue(config, base, value_type):
    """Return some information from the camera geometry for the current detector
    """
    camera = galsim.config.GetInputObj('camera_geometry', config, base, 'CameraValue')
    det_name = base['det_name'].replace('-', '_')
    ccd = camera[self.det_name]

    req = { 'field' : str }
    opt = { 'num' : int }  # num, if present, was used by GetInputObj
    kwargs, safe = galsim.config.GetAllParams(config, base, req=req, opt=opt)
    field = kwargs['field']

    val = value_type(getattr(ccd, field))
    return val, safe

RegisterValueType('CameraValue', CameraValue, [float, int, str], input_type='camera_geometry')
```
Then in the config, it would be
```
xsize: { type: CameraValue, field: xsize }
ysize: { type: CameraValue, field: ysize }
```
And just add these attributes to the CCD class.  Or make them properties to derive from the bounds attribute, which already has this information (but in a slightly less accessible form).

config/desc/imsim/readout.py
```diff
@@ -26,64 +223,61 @@ def finalize(self, config, base, main_data, logger):
...
+            logger.warning("Amp %s has bounds %s.", amp_name,
+                           hdu.header['DETSEC'])
+        return hdus
 
     def writeFile(self, file_name, config, base, logger):
         """Write this output object to a file.
 
-        The base class implementation is appropriate for the cas that the result of finalize
-        is a list of images to be written to a FITS file.
+        The base class implementation is appropriate for the cas that
```
rmjarvis, line 265: Typo (mine I suspect) cas -> case.
But in fact, this comment is not really relevant anymore, since the output from your `finalize` is now an `HDUList`.  So probably best to just remove this paragraph.

config/desc/imsim/readout.py
```diff
@@ -12,9 +208,10 @@ def finalize(self, config, base, main_data, logger):
 
         This function will be called after all images have been built.
 
-        It returns some sort of final version of the object.  In the base class, it just returns
-        self.data, but depending on the meaning of the output object, something else might be
-        more appropriate.
+        It returns some sort of final version of the object.  In the
+        base class, it just returns self.data, but depending on the
+        meaning of the output object, something else might be more
+        appropriate.
```
rmjarvis, line 214: Could mention that your final version is an HDUList rather than a list of images.  This doc string was for the base class implementation.

rmjarvis (changes requested):
Hi Jim.  Most of my comments are about doc strings.  The main substantive suggestion is about the CTE implementation, which I don't think is correct.  But of course, it's possible that I'm just misunderstanding how that function is supposed to work.  So feel free to push back on that if I'm totally off base.

rmjarvis:
FYI, I merged in the current main and resolved the conflict.  Really just that the new camera.py is now in imsim/, rather than config/desc/imsim/.

rmjarvis:
LGTM

jchiang87:
@cwwalter Shall we go ahead and merge this or do you have comments?

PR #253, opened 2021-05-26
Fix logic of how we decide image_size for very bright objects

rmjarvis:
Fix a bug Chris ran into with the very bright stars.


cwwalter (approved):
Looks good.

Thanks Mike.

PR #254, opened 2021-08-30
Install new config modules with setup.py

rmjarvis:
This PR moves the new GalSim config-based stuff to a primary `imsim` directory that we can install with `python setup.py install`.

I tested that
```
python setup.py install
cd config
galsim imsim.yaml
```
ran as expected.  However, I only tested it with stars so as not to take forever.  But it completed and produced output files, so I think it's working.


imsim/meta_data.py
```diff
@@ -0,0 +1,10 @@
+
+import os
+
+imsim_dir = os.path.split(os.path.realpath(__file__))[0]
+install_dir = os.path.split(imsim_dir)[0]
+
+if 'IMSIM_DATA_DIR' in os.environ: # pragma: no cover
+    data_dir = os.environ['IMSIM_DATA_DIR']
+else:
+    data_dir = os.path.join(install_dir, 'imsim', 'data')
```
cwwalter, line 10: Could we leave out the symlink for the data directory in the new imsim directory , not make install_dir variable here and then just:

```
else:
     data_dir = os.path.join(imsim_dir, 'data')
```

here?

Or is this tied to either the install system, or trying to be flexible about where we might put the data directory?

setup.py
```diff
@@ -0,0 +1,73 @@
...
+verstrline = open(version_file, "rt").read()
+VSRE = r"^__version__ = ['\"]([^'\"]*)['\"]"
+mo = re.search(VSRE, verstrline, re.M)
+if mo:
+    imsim_version = mo.group(1)
+else:
+    raise RuntimeError("Unable to find version string in %s." % (version_file,))
+print('ImSim version is %s'%(imsim_version))
+
+dist = setup(name="ImSim",
```
cwwalter, line 61: imSim (first capital letter)

setup.py
```diff
@@ -0,0 +1,73 @@
...
+    version=imsim_version,
+    author="ImSim Developers (point of contact: Chris Walter)",
```
cwwalter, line 63: imSim

cwwalter (approved):
Thanks Mike.  

I'm not a setup.py expert so didn't really check that part. But it looks good. I mostly just had a question on the data_directory.

Also a minor capitalization fix.

cwwalter:
I think I might have seen this elsewhere, but can't find it now.. did you want to move the old code into a new directory after this merge?

imsim/meta_data.py
rmjarvis: We need the symlink for it to get installed correctly.  But I think you're right that breaking off the `imsim` part from `imsim_dir` and then adding it back here is unnecessary.  (I just copied what we did for GalSim, but I can't think why that would be necessary.)

rmjarvis:
> did you want to move the old code into a new directory after this merge?

I originally said that, but since the new version is in a different directory name anyway, there's no conflict with the old code staying where it is for now.   

We'll need to do a little work to get the test suite working with the new installation, since not all the code has been ported over yet.  I started working on that, but that can be a different PR.

imsim/meta_data.py
```diff
@@ -7,4 +7,4 @@
 if 'IMSIM_DATA_DIR' in os.environ: # pragma: no cover
     data_dir = os.environ['IMSIM_DATA_DIR']
 else:
-    data_dir = os.path.join(install_dir, 'imsim', 'data')
+    data_dir = os.path.join(imsim_dir, 'data')
```
cwwalter, line 10: LGTM, Can you actually delete the line that sets install_dir since you aren't using it anymore here?

Or, is it used other places for the installation?

cwwalter (approved):

imsim/meta_data.py
rmjarvis: OK.  Removed.

PR #255, opened 2021-09-17
Batoid wcs

jmeyers314:


jmeyers314:
I think this is close.  The only question I have is what time scale to use for the input to the WCS factory.  From dissecting the old WCS code, I'm pretty sure that the input time was given in the TAI scale.  I can't remember if the time was usually given for the beginning of the exposure or the middle.  I propose we figure out what OpSim uses natively for these (scale and beginning/middle of exposure convention) and just copy that.

rmjarvis (approved):
I've already gone through it (and the config edits are mine), so I'm happy with this, modulo the right time to use for obstime.

PR #256, opened 2021-10-13
add bleed trail and cosmic ray code

jchiang87:
This is the code from the old imSim.


jchiang87:
@rmjarvis I tried to implement most of your suggestions, but my understanding of the galsim config stuff is pretty limited, so I expect these changes may not be the preferred way.

config/imsim.yaml
```diff
@@ -242,6 +235,9 @@ output:
 
     exp_time: 30
 
+    cosmic_ray_rate: 0.2
+    cosmic_ray_catalog:
```
rmjarvis, line 239: I think this is a syntax error.  To leave it empty, I'm pretty sure you need it to be `{}`.

rmjarvis: Actually, this is apparently fine. I just tried it, and it worked. Huh. Yaml is cleverer than I thought... :)

cwwalter: Hmm.. that's surprising... I had to use the {} in the camera geometry case or it didn't work.

rmjarvis: I think with nothing there, YAML sets it to None, not an empty dict, which in this case is fine, but wouldn't be for camera_geometry.  In any case, I think it's probably cleaner to remove this line, since most users will be fine with the default CR file I suspect.

imsim/bleed_trails.py
```diff
@@ -0,0 +1,132 @@
...
+        xpix: int
+            Index of the pixel to which charge will be redistributed.
+            If it is already at full_well, do nothing.
+
+        Returns
+        -------
+        bool: True if all excess charge has been redistributed.
+        """
+        try:
+            bled_charge = min(self.full_well - self.imarr[xpix],
```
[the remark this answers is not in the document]
rmjarvis: I'll make a unit test that checks the thing I'm concerned about on my unit test branch and fix it there if the test confirms my thought about this.

config/imsim.yaml
cwwalter: I see.

So, if you used the {} would it overwrite the default so things wouldn't work?

In the old config we had 

```
catalog = default
```

which is at least good in letting people know the option is there.

rmjarvis: Right, but that nominally means to use the catalog file named `default`.  So IMO is more confusing than just leaving it out, which means to use the default value, which is some catalog we ship with imsim.

rmjarvis (approved):

rmjarvis:
I made one change to how some of the config things are parsed, which I think gives better error messaging when the user does something wrong.
I also fixed the det_name split to split on '_', rather than '-'.  I think Chris noticed this on main, and it tripped when I tested my own changes for this branch.
And I rebased to master, so it should merge cleanly now.

I think this is all ready to go.

config/imsim.yaml
cwwalter: Right.. I actually meant just to leave it blank with a comment as a reminder it is settable.  But, anyway I agree that almost no one will ever touch it.

jchiang87:
Thanks @rmjarvis!  I'll go ahead and merge ~COB PT today.

cwwalter:
Great!

Jim, I just merged the split config files. Could you update the new imsim-config.yaml file with the new yaml for these changes too?

(At some point I want to  delete the old monolithic imsim.yaml file too)

jchiang87:
> Jim, I just merged the split config files. Could you update the new imsim-config.yaml file with the new yaml for these changes too?

Would you mind doing it?  I haven't been following the details of your changes, so I think it would be easier if you did that. Thanks.

cwwalter:
Sure. 

I think I can wait until you merge and then just copy your changes from that section of the monolithic file into my yaml file.

jchiang87:
@cwwalter I rebased to main and made the change to imsim-config.yaml.   If everything looks ok to you, please go ahead and merge.

cwwalter:
Thanks! Sorry I was out before.

So only change to the yaml files is the two cosmic ray lines right?

PR #257, opened 2021-10-27
Make split config files

cwwalter:
This PR fixes a small detector name parsing option in the output code to make things run (and be consistent with gen3 conventions) and also uses the galsim template keyword to split the config into a 

config-user.yaml and config-imsim.yaml.

Users can just modify (or copy and modify) the user version.  Currently it relies on two environment variables so (e.g.)

```
export IMSIM_DIR=/Users/walter/LSST/imSim

export SIMS_SED_LIBRARY_DIR=/cvmfs/sw.lsst.eu/darwin-x86_64/lsst_sims/sims_w_2020_15/stack/current/DarwinX86/sims_sed_library/2017.01.24

PYTHONPATH=$IMSIM_DIR:$PYTHONPATH
```

and one can run imSim from anywhere.


config/imsim-config.yaml
```diff
@@ -0,0 +1,293 @@
...
+
+# Any input data is set here.  These are read in at the start of the program and potentially
+# updated for each output file.
+# Also includes things that need some set up at the start of an exposure, like the atmospheric PSF.
+input:
+    camera_geometry: {}
+
+    instance_catalog:
+        # This enables InstCat types
+        file_name: {} # No default instance catalog.
```
rmjarvis, line 32: This strikes me as slightly confusing, since the intended value here is a string, not a dict.  If you plan to always give this on the command line, I'd suggest putting something that looks like a file_name (e.g. `instance_catalog.txt`) and a comment saying that this is normally overridden on the command line using `galsim imsim-config.yaml input.instance_catalog.file_name=/path/to/some/instance_catalog.txt`.

rmjarvis (approved):
Just one comment, that I don't like the {} as the default instance catalog filename.  Mostly because I think it will give a confusing error message if the user doesn't override it.  But I'll leave that up to you whether you want to switch it.

config/imsim-config.yaml
cwwalter: OK. Good idea I changed this to make the error message more comprehensible.

PR #258, opened 2021-11-04
importing cameras from obs lsst

jchiang87:


config/imsim-config.yaml
```diff
@@ -25,7 +25,8 @@ eval_variables:
 # updated for each output file.
 # Also includes things that need some set up at the start of an exposure, like the atmospheric PSF.
 input:
-    camera_geometry: {}
+    camera_geometry:
+        camera_class: lsst.obs.lsst.LsstCam
```
rmjarvis, line 29: I don't love having the full specification of the lsst stack namespace here.  Can we make this just be `LsstCam`?  That seems like the extent of what the user should have to specify.

jchiang87: That's fine, but then we'd have to prepend the `lsst.obs.lsst` part in a few different places.

config/imsim-config.yaml
```diff
@@ -230,6 +232,8 @@ output:
     nproc: 1    # Change this to work on multiple CCDs at once.
     nfiles: 1   # Default is all 189 CCDs.  Set to 1 while testing.
 
+    camera_class: lsst.obs.lsst.LsstCam
```
rmjarvis, line 235: I'd suggest probably have this be just
```
camera: LsstCam
```
here, and remove it entirely from input.  I think having it load at input time made sense when we were reading a pickle file, but probably doesn't anymore.  This seems like it should just be a value specification, not an input type now.

config/imsim-config.yaml
rmjarvis: Right.  I guess my point is that that should be our job, not the user's job.

jchiang87: @rmjarvis I think I've addressed your concerns.  Please have another look.

config/imsim-config.yaml
```diff
@@ -230,6 +229,8 @@ output:
     nproc: 1    # Change this to work on multiple CCDs at once.
     nfiles: 1   # Default is all 189 CCDs.  Set to 1 while testing.
 
+    camera_class: LsstCam
```
rmjarvis, line 232: I would prefer this to be just `camera`, rather than `camera_class`, but I'll defer to others' opinions if people disagree.  From a user's perspective, it feels like they should be identifying what camera they want to simulate, not some python class in the lsst stack they want to use.  

In fact, if we start having people use DECam or other non Rubin camera's, I'd want us to special case them and prepend the right modules, not make them know to write `lsst.obs.decam.DarkEnergyCamera`.  Rather, I'd want us to list somewhere the values that we support and internally prepend whatever the right thing is in terms of the lsst naming details.

jchiang87: `camera` is fine as far as I'm concerned.  In principle, people can write new obs packages or add cameras to existing obs packages, so asking users to give the fully qualified class for these cases doesn't seem like much of an ask, but I guess we can just support the obs.lsst cameras for now.

cwwalter: Not having the users need to know about the camera class names (or that there are different classes) also feels correct to me.

I think there will be a group of people using the program unfamiliar with DM who will just want to choose from a list of possible camera names.

cwwalter:
BTW, I'm not positive is this is the best place to address this, but since the PR touches camera.py:

Do you think the information in 

_ts3_xtalk = np.array(

is best kept here?  I know it isn't in the camera object now (and I didn't really understand the answers from Chris Waters about where it might go when I asked), but it seems to me it should be stored somewhere different than in the code itself.

jchiang87:
>Do you think the information in
>
> _ts3_xtalk = np.array(
>
> is best kept here? 

Currently, it appears in places like this:  https://github.com/lsst/obs_lsst/blob/master/policy/imsim/ccdData.yaml
However, there isn't a corresponding file in the lsstCam folder, and I don't think we want to use `LsstCamImSim` do we?  I have no idea if things will change though given that Chris couldn't just point you at this file.

cwwalter:
What does  "policy" mean in this context?  Is that info retrievable via the API of package?

jchiang87:
> What does "policy" mean in this context?
I don't really know.  It is a confusing name here.

> Is that info retrievable via the API of the package?
Yes.

cwwalter:
OK based on slack discussions, the camera calibration info is a more complicated question than can be addressed here.  So, don't let that issue affect this PR.

jchiang87:
@rmjarvis please have another look.

rmjarvis (approved):
Thanks Jim.  I'm happy with this UI.

Issue #213, opened 2019-03-26
Allow galaxies with no internal or MW extinction or model. 


cwwalter:
Simple objects/stars like this:

```
object 0 0.0 0.0 20 flatSED/sed_flat.txt.gz 0 0 0 0 0 0 point none none
```

Are sometimes used for tests and studies.  As it is now these objects are filtered out because 

```
Av == Rv == 0
```

and sims_photUtils divides by Rv.  Since the amplitude of dust extinction added to the SED goes roughly as Av/Rv, the case where the dust_model is ‘none’  should probably be interpreted as something like Av==0, Rv==1.

Since we will be re-writing the instance catalog and parser soon this should be kept in mind when that new work is done.  Thanks to @danielsf and @jchiang87 for insight into this issue.

PR #259, opened 2021-11-05
Update unit tests for the new code installed via setup.py

rmjarvis:
This updates some of the unit tests to work with the new code structure.  Specifically, the following files now all work with imsim installed with setup.py:

test_FWHMgeom.py
test_atmPSF.py
test_batoid_wcs.py  (already was working prior to this PR)
test_cosmic_rays.py
test_fopen.py
test_instcat_parser.py
test_logging.py
test_optical_zernikes.py
test_psf.py
test_tree_rings.py
test_trimmer.py

I've marked some places of concern with XXX comments.  The most notable of these, I think, is about the WCS in test_instcat_parser.py.  The Batoid WCS for the test instance catalog seems to be about 10 arcsec off.  (Not uniformly, but that's ~ the maximum error.)  That seems like more than we should have expected, so maybe @jmeyers314 could take a look at this to see if this is expected because you are doing something right now that used to be wrong, or if there is something here to be fixed.

There are more test files still to convert over, but this seems like enough of a start that it's worth people looking over the updates so far.


rmjarvis:
Update: I rebased onto main to use Jim's new camera code.
Also, I managed to get github actions working with stackvana (installed with mamba), running the tests that are working so far, and the CI agrees that they are all working.  :)

cwwalter:
Awesome!

tests/test_instcat_parser.py
```diff
@@ -147,99 +150,145 @@ def test_object_extraction_stars(self):
...
+            wcs = all_wcs[det_name]
+            cat = all_cats[det_name]
+            for i in range(cat.nobjects):
+                image_pos = cat.image_pos[i]
+                world_pos = cat.world_pos[i]
+                self.assertLess(world_pos.distanceTo(wcs.toWorld(image_pos)), 0.0005*galsim.arcsec)
+
+        ra_arr = np.array([pos.ra.rad for cat in all_cats.values() for pos in cat.world_pos])
+        dec_arr = np.array([pos.dec.rad for cat in all_cats.values() for pos in cat.world_pos])
+        # XXX: These are only within 1.e-4 radians, which is kind of a lot.
```
jchiang87, line 209: Almost all of the tests in this file were written by Scott, so I think we should just update things in a way that makes sense to us.

tests/test_instcat_parser.py
```diff
@@ -147,99 +150,145 @@ def test_object_extraction_stars(self):
...
-                self.assertAlmostEqual(flux/gs_obj.flux(bp), 1.0, 10)
+        bp = galsim.Bandpass('LSST_%s.dat'%md['band'], wave_type='nm').withZeropoint('AB')
+
+        for det_name in all_wcs:
+            wcs = all_wcs[det_name]
+            cat = all_cats[det_name]
+            for i in range(cat.nobjects):
+                obj = cat.getObj(i, bandpass=bp)
+                if 0:
+                    # XXX: The old test used the sims Sed class.  Circumventing this now,
```
jchiang87, line 228: ok, with me.

tests/test_instcat_parser.py
```diff
@@ -147,99 +150,145 @@ def test_object_extraction_stars(self):
...
-            chip_name = chipNameFromPupilCoordsLSST(xpup, ypup)
-            if chip_name is not None:
-               self.assertIn(unq, unique_id_dict[chip_name])
-               valid_chip_names.add(chip_name)
-               valid += 1
-
-        self.assertGreater(valid, 10)
-        self.assertGreater(len(valid_chip_names), 5)
+        if 0:
+            # XXX: This doesn't seem relevant anymore.  But leaving this here in case we want
```
jchiang87, line 259: Ok.

tests/test_instcat_parser.py
```diff
@@ -253,153 +302,193 @@ def test_object_extraction_galaxies(self):
...
+        for det_name in all_wcs:
+            wcs = all_wcs[det_name]
+            cat = all_cats[det_name]
+            for i in range(cat.nobjects):
+                obj = cat.getObj(i, bandpass=bp)
+                i_obj = np.where(truth_data['uniqueId'] == cat.id[i])[0]
+                if len(i_obj) == 0: continue
+                i_obj = i_obj[0]
+                if 0:
+                    # XXX: The old test using the sims Sed class.
```
jchiang87, line 389: Ok.

tests/test_instcat_parser.py
```diff
@@ -253,153 +302,193 @@ def test_object_extraction_galaxies(self):
...
-        self.assertEqual(phot_params.bandpass, 'r')
-        self.assertEqual(phot_params.nexp, 2)
-        self.assertAlmostEqual(phot_params.exptime, 15.)
-        self.assertEqual(phot_params.readnoise, 0)
-        self.assertEqual(phot_params.darkcurrent, 0)
+        meta = imsim.OpsimMetaDict(self.phosim_file)
+
+        self.assertEqual(meta['gain'], 1)
+        self.assertEqual(meta['band'], 'r')
+        # XXX: These used to be nexp=2 ?? and exptime=15.  I guess per snap, but we're not doing
```
jchiang87, line 457: ok, with me.

jchiang87 (changes requested):
I added comments at most of your XXX questions, but didn't go through the changes in detail otherwise.  I'll have a closer look, but wanted to address those questions right away.

imsim/instcat.py
```diff
@@ -152,18 +157,45 @@ def __init__(self, file_name, wcs, sed_dir=None, edge_pix=100, sort_mag=True, fl
...
-                    objinfo_list.append(tokens[12:dust_index])
-                    dust_list.append(tokens[dust_index:])
+                    objinfo = tokens[12:dust_index]
+                    dust = tokens[dust_index:]
+
+                    if skip_invalid:
+                        # Check for some reasons to skip this object.
+                        # XXX: Is this right?  We require dust?
+                        #      The previous code had that galactic_av = galactic_rv = 0 is invalid.
+                        #      That doesn't seem right to me...
```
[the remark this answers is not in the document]
rmjarvis: But if no dust is given in the input, can't we just not apply any extinction from dust?  I.e. Av=0, Rv=1.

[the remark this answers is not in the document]
jchiang87: sure.  I'm not advocating we keep things as they are.  I was just giving the reason why I think the test is written this way.  Same applies to my other comments.

tests/test_trimmer.py
```diff
@@ -16,24 +18,49 @@ def setUp(self):
...
+                                    skip_invalid=False)
+        self.assertEqual(instcat.nobjects, 0)
 
         # Check various values of chunk_size.
-        for chunk_size in (5, 10, 100):
-            objs = desc.imsim.InstCatTrimmer(instcat, [sensor], minsource=None,
-                                             chunk_size=chunk_size)
-            self.assertEqual(len(objs[sensor]), 24)
+        # XXX: We don't use chunking anymore.  If we revisit this and put it back in,
+        #      then we can re-enable this test.
```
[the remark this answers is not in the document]
rmjarvis: Right.  I meant if we decided that the new code should be chunking like you did in the trimmer, then we might want a version of this test, modified to test the new code. I guess I'm also kind of prodding that we recall why we did chunking in the old case and think about whether it would make sense to add something similar now.

tests/test_instcat_parser.py
rmjarvis: I'm hoping @jmeyers314 can take a look at this test to see whether the previous version did make sense, and the failures are a sign of a bug or if the test isn't quite relevant anymore given the new wcs handling.

imsim/instcat.py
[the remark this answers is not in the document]
cwwalter: We ran into this before:

https://github.com/LSSTDESC/imSim/issues/213

tests/test_trimmer.py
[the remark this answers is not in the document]
jchiang87: IIRC, the original reason was to avoid using up a lot of memory when reading in the instance catalogs (which covered the full focal plane) since most of the entries would be omitted once the objects that landed on the CCD were identified.

tests/test_instcat_parser.py
```diff
@@ -54,15 +45,41 @@ def tearDownClass(cls):
             shutil.rmtree(cls.scratch_dir)
 
     def setUp(self):
-        self.phosim_file = os.path.join(self.data_dir,
-                                         'phosim_stars.txt')
+        self.phosim_file = os.path.join(self.data_dir, 'phosim_stars.txt')
+        # XXX: What is the point of the extra_commands?  Seems like we could remove this bit.
```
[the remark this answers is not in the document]
jchiang87: Now that I think about it, this is a phosim thing: it allowed for extra commands to be added via this mechanism, so I suspect it was added here simply to be consistent with phosim. So I think we can safely remove it.

tests/test_instcat_parser.py
jmeyers314: I'll take a closer look.  Browsing `main`, it looks like these positions weren't previously tested at all?  Is that right?  (So maybe they were always 1e-4 rad off?)

rmjarvis: They weren't tested on main (since we weren't running these tests), but I'm pretty sure they were tested in v1.0.  That was with the older wcs code, not the BatoidWCS.

tests/test_logging.py
```diff
@@ -2,8 +2,9 @@
 Test the setting of the logging configuration.
 """
 import unittest
-import desc.imsim
+import imsim
 
+# XXX: I think this is probably not needed anymore.  Now use GalSim logger.
```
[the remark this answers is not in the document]
rmjarvis: Good.  Removed this.  Also test_TracebackDecorator and test_config_reader.

imsim/instcat.py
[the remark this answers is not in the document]
rmjarvis: I changed this so these are not skipped, but are treated as Av=0, Rv=1.

tests/test_trimmer.py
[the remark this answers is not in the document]
rmjarvis: OK, I've removed this.  In the new code, we already have the WCS as we're reading in the file, so there's no overhead of keeping any objects in memory that won't end up being kept.

tests/test_instcat_parser.py
cwwalter: I think that is right, it used the sims code with the distortion model from for the WCS as we discussed before.

imsim/camera.py
```diff
@@ -202,15 +202,16 @@ def __getattr__(self, attr):
...
+def get_camera(camera='LsstCam'):
     """
     Return an lsst camera object.
 
     Parameters
     ----------
     camera : str
        The class name of the LSST camera object. Valid names
-       are 'LsstCam', 'LsstComCam', 'Latiss'.
+       are 'LsstCam', 'LsstComCam', 'Latiss'. [default: 'LsstCam']
```
jchiang87, line 214: Looking at the check for `telescope` above,
```
if telescope != 'LSST':
    raise NotImplementedError("Batoid WCS only valid for telescope='LSST' currently")
```
it occurs to me that we should omit `'Latiss'` from the camera options here and below.

jchiang87 (changes requested):
Except for the one comment about removing `'Latiss'` as a camera option, looks good to me.

imsim/camera.py
rmjarvis: OK.  I don't actually know what Latiss is.  I guess that's AuxTel maybe?
I'm fine with removing that for now, but I guess people may want to make simulations of that eventually.  So we should add an issue to add that back in.

rmjarvis: Ah, nvm I guess.  That's a slitless spectrograph.  Yeah, we're never going to simulate that with ImSim (at least until GalSim Issue [#1018](https://github.com/GalSim-developers/GalSim/issues/1018) is done).  So just removing this is probably right.

cwwalter: Right: LSST Atmospheric Transmission Imager and Slitless Spectrograph (LATISS) on the Auxtel.

For the future I wonder if the way to do something like this would be to have dispersive elements in Batoid?

jchiang87 (approved):

rmjarvis:
I opened Issue #262 for addressing the WCS regression tolerance being a bit large.  I think everything else has been resolved, so this can probably be merged, but @cwwalter if you want more time to look it over, just say so.

cwwalter:
OK, LGTM and we will track #262 separately.

PR #261, opened 2021-11-15
opsim db interface

jchiang87:


imsim/instcat.py
```diff
@@ -399,67 +400,65 @@ def getHourAngle(self, mjd, ra):
         return ha
 
 class OpsimMetaDict(object):
-    """This just handles the meta information at the start of the instance catalog file.
+    """"Read the exposure information from the opsim db file.
```
rmjarvis, line 403: Extra quote here

imsim/instcat.py
```diff
@@ -399,67 +400,65 @@ def getHourAngle(self, mjd, ra):
...
 
     The objects are handled by InstCatalog.
     """
-    _req_params = { 'file_name' : str, }
+    _req_params = { 'file_name' : str,
+                    'visit' : int,
+                    'snap' : int}
```
rmjarvis, line 409: Maybe make snap optional, with default 1?  (Or 0?)  Not sure how often users will actually care to distinguish the snap number.

imsim/instcat.py
```diff
@@ -399,67 +400,65 @@ def getHourAngle(self, mjd, ra):
...
 
         # Add a couple derived quantities to meta values
         # Note a semantic distinction we make here:
-        # "filter" is the number 0,1,2,3,4,5 from the input instance catalog.
-        # "band" is the character u,g,r,i,z,y.
+        # "filter" or "band" is a character u,g,r,i,z,y.
         # "bandpass" will be the real constructed galsim.Bandpass object.
-        self.meta['band'] = 'ugrizy'[self.meta['filter']]
-        self.meta['HA'] = self.getHourAngle(self.meta['mjd'], self.meta['rightascension'])
-        self.meta['airmass'] = self.getAirmass(self.meta['altitude'])
```
rmjarvis, line 442: I'm assuming (but wanted to confirm) that this line is removed because 'airmass' is already in the db?

rmjarvis (approved):
This looks good to me.  However, I'd like to wait to merge it until the unit tests PR is merged.  We have tests there that check most of this functionality, so it would be nice to verify that this didn't break anything.

imsim/instcat.py
jchiang87: The current cadences have visits with either 1 or 2 exposures per visit, as well as different exposure times depending on the band.   Putting the snap here is a near term way to address this.  Ultimately, I would think we'd want the code to generate as many exposures as specified for the requested visit, but I wasn't sure what changes that imply for the current code, so I thought we could address that in a separate PR.

imsim/instcat.py
jchiang87: Yes, that's right.

cwwalter:
Do you want to also handle allowing the text formatted input files by adding back in the code here (as an option?)?  Or do it separately?

jchiang87:
I was thinking to add back the code from before and set things up to check if the `file_name` was an instance catalog or an opsim db file and proceed accordingly.

jchiang87:
@cwwalter @rmjarvis I restored the functionality to read the opsim meta data from an instance catalog if one of those is given instead of an opsim db file, and I rebased to main.  Please have a look.

imsim/instcat.py
```diff
@@ -508,14 +584,92 @@ def __init__(self, file_name, logger=None):
...
+            The LSST ugrizy band.
+            [default: self.get('band')]
+        altitude: float
+            The altitude in degrees of the pointing.
+            [default: self.get('altitude')]
+        Returns
+        -------
+        float: FWHM of the combined PSF in arcsec.
+        """
+        return 0.822*self.FWHMeff(rawSeeing, band, altitude) + 0.052
```
rmjarvis, line 672: These 3 functions (getAirmass, FWHMeff, FWHMgeom) already exist below in this file, so don't need to add them here.

rmjarvis:
I think it would be worth adding some tests of the new db functionality.  I think right now, all the unit tests are taking the instcat path through the code, rather than the db path.

rmjarvis:
LGTM.  👍

PR #263, opened 2021-12-01
Update test_camera_readout.py and test_bleed_trails.py

rmjarvis:
This PR updates two more test files to work with the new code, with one exception.  

One of the tests in test_camera_readout is currently failing, and I don't know what to do to fix it, so @jchiang87 could you please take a look.  It's `test_get_amplifier_hdu`.  It seems that either the amps are in a different order than the test expects, or maybe there is a bug in the new code related to how the amp bounds are computed.  

Here is the pytest output for it:
```
test_camera_readout.py .F..                                                                  [100%]

============================================= FAILURES =============================================
____________________________ ImageSourceTestCase.test_get_amplifier_hdu ____________________________

self = <test_camera_readout.ImageSourceTestCase testMethod=test_get_amplifier_hdu>

    def test_get_amplifier_hdu(self):
        "Test the .get_amplifier_hdu method."
        hdus = self.readout.final_data
        hdu = hdus[1]
        self.assertEqual(hdu.header['EXTNAME'], "Segment10")
        self.assertEqual(hdu.header['DATASEC'], "[4:512,1:2000]")
        # XXX: This test is currently failing.  This DETSEC now seems to correspond to
        #      hdu 17, which is C00, not C10.  C10 has DETSEC = [509:1,4000:2001]
        #      I don't know what should be changed to fix this.  Is it an error in the code?
        #      Or is the test wrong now after some change to the camera layout?
        #      Or did I screw something up in trying to adapt the test from the old code?
>       self.assertEqual(hdu.header['DETSEC'], "[509:1,1:2000]")
E       AssertionError: '[509:1,4000:2001]' != '[509:1,1:2000]'
E       - [509:1,4000:2001]
E       ?        ^^^^    ^
E       + [509:1,1:2000]
E       ?        ^    ^

test_camera_readout.py:80: AssertionError
```

I also added a test that confirmed my worry about the bleed not working the way I think it should work.  cf. https://github.com/LSSTDESC/imSim/pull/256#discussion_r729393366

Specifically, a bleed at start of a channel (small y values) shouldn't depend on the pixel values way at the other end of the channel (large y values).  There was already a test that checked aspects of this, but the rest of the channel were all 0, so it didn't notice the bug.  

I added a test where the other end has full_well/2 values and check that the bleed at the start is the same regardless of what's going on at the end.  This test failed with the old code, but works with my change.


jchiang87:
The `test_camera_readout.py` failure is an ITL vs e2V mismatch:  The test code assumes the CCD is ITL, but it's actually an e2V sensor given the actual LSSTCam configuration.  Switching to raft R20 from R22 should make things work for that test code.  We can add similar tests for e2V.

rmjarvis:
I already switched it to R20.  Otherwise the sizes are even more different.  This is the right overall shape, 1/16 of an ITL sensor.  Just it seems to be the wrong amp.

rmjarvis:
Update: I added test_raw_file_writing.py to this PR too, since it's pretty related to other things on this branch.

tests/test_camera_readout.py
```diff
@@ -5,57 +5,91 @@
...
+        hdus = self.readout.final_data
+        hdu = hdus[1]
+        self.assertEqual(hdu.header['EXTNAME'], "Segment10")
         self.assertEqual(hdu.header['DATASEC'], "[4:512,1:2000]")
+        # XXX: This test is currently failing.  This DETSEC now seems to correspond to
+        #      hdu 17, which is C00, not C10.  C10 has DETSEC = [509:1,4000:2001]
+        #      I don't know what should be changed to fix this.  Is it an error in the code?
+        #      Or is the test wrong now after some change to the camera layout?
+        #      Or did I screw something up in trying to adapt the test from the old code?
         self.assertEqual(hdu.header['DETSEC'], "[509:1,1:2000]")
```
jchiang87, line 80: There's a bit of history regarding this test, but the bottom line is that I think the `DETSEC` value should be changed to `"[509:1,4000:2001]"` here for C10 and `"[3563:3055,4000:2001]"` for C17 below.  If we were to test e2V CCDs, those would be `"[512:1,4004:2003]"` and `"[3584:3073,4004:2003]"`, respectively.

The back story on this test is that the current `DETSEC` values are based on Camera document [LCA-10140](https://ls.st/lca-10140), but more recently, the Camera team decided to adopt the DM conventions for displaying mosaics of the focal plane.   Regarding amp orientation and location, the DM conventions amount to rotating images like those seen on sheets 1 or 5 of [LCA-13381](https://docushare.lsst.org/docushare/dsweb/Get/LCA-13381) 90 degrees clockwise.

rmjarvis: OK.  Thanks for tracking it down.  I fixed the test to match the values the code is giving.

cwwalter: Jim, does 

"more recently, the Camera team decided to adopt the DM conventions for displaying mosaics of the focal plane. "

this only refer to this situation? Or will the camera team move towards DM conventions more generally?

jchiang87: The Camera team is also moving towards using the Gen3 infrastructure and pipe_task framework for implementing new electro-optical analysis code, but that a much more involved transition and we may always rely to some extent on non-DM compliant scripts or code. For straight-forward things like the focal plane layout conventions, we've generally agreed that using DM conventions is best path forward, but I can't say definitively for any specific item.

jchiang87 (changes requested):
Sorry for the long delay in looking at this.  There are just a few requested changes, mostly needed to get the code to run with skyCatalogs.

imsim/readout.py
```diff
@@ -63,64 +268,80 @@ def cte_matrix(npix, cti, ntransfers=20):
...
     """Create a primary HDU for the output raw file with the keywords
     needed to process with the LSST Stack."""
-    opsim_md = galsim.config.GetInputObj('opsim_meta_dict', config, base,
-                                         'OpsimMeta')
-    md = opsim_md.meta
     phdu = fits.PrimaryHDU()
-    phdu.header['RUNNUM'] = md.get('obshistid')
-    phdu.header['OBSID'] = md.get('obshistid')
-    exp_time = base['exp_time']
+    phdu.header['RUNNUM'] = opsim_md.get('obshistid')
```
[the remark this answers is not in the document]
rmjarvis: Is this a change?  The `tiny_instcat.txt` instance catalog we've been using for tests has `obshistid`, not `observationId'.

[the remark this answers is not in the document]
jchiang87: The most recent opsim db files have `observationId`, and for the old instance catalogs, `obshistid` is mapped to that here: https://github.com/LSSTDESC/imSim/blob/main/imsim/instcat.py#L590

rmjarvis:
Thanks Jim!  I think I addressed everything.

jchiang87 (approved):
Thanks, Mike!  Everything looks good to me.

rmjarvis:
@cwwalter Do you want to look too, or should I merge?

cwwalter:
Let me take a quick look at the new changes in a bit.

cwwalter:
OK LGTM to merge thanks!

PR #264, opened 2021-12-04
Add a module for building Flat-field images

rmjarvis:
This PR adds an image builder for building flat-field images.  There is an example config file in the config directory, which makes a realistic flat-field image using the LSST_CCD output type with the LSST_Flat image type.

Screen Shot 2021-12-04 at 12 46 23 AM

I also updated the unit tests in test_flats.py to test the code with (a) no sensor, (b) Silicon sensor, but no tree rings, and (c) Silicon with tree rings and verified that the mean, variance and covariance all make sense in each case.  So it's quite a bit more extensive than what used to be there.

Finally, I removed the `transpose=True` specification that we had been using for the Silicon sensor, since the consensus seems to be that this is wrong now that we are no longer using phosim-style e-images.


rmjarvis:
This and #263 have been sitting here since before Christmas.  Anyone want to review them?  @jchiang87 @cwwalter @jmeyers314

config/flat.yaml
```diff
@@ -0,0 +1,80 @@
...
+
+    dir: flats
+    file_name:
+        type: FormattedStr
+        format : flat_%05d-%s-%s-det%03d.fits
+        items:
+            - 1  # When making multiple flats, probably should increment this for each file.
+            - "@image.wcs.band"
+            - "$det_name"   # A value stored in the dict by LSST_CCD
+            - "@output.det_num"
```
jchiang87, line 80: To make use of the flats by the LSST code, they'll need to be formatted into FITS files with per-amp HDUs, just like the on-sky images.   So we should include a `readout` section in this example.  However, that requires information from the `OpsimMetaDict` class.  This aligns with the needs of DM code to be able to ingest calibration data.  One way to handle this would be to have an opsim db specifically for making calibration frames.

config/flat.yaml
```diff
@@ -0,0 +1,80 @@
...
+modules:
+    - imsim
+
+input:
+
+    tree_rings:
+        # This enables TreeRingCenter and TreeRungFunc, which are stored as a dict here based
+        # on the detector name, so the right value can be accessed for each object.
+        # This file lives in the imsim repo.
+        file_name: "tree_ring_data/tree_ring_parameters_2018-04-26.txt"
```
jchiang87, line 14: When I run using this config, I need to omit the `tree_ring_data/` folder from the `file_name`.

jchiang87 (changes requested):
@rmjarvis I had a quick look at this, and I still need to go through the code in more detail, but I had a high-level comment/change request that is probably worth bringing up now:  The flats will need to be formatted the same as the on-sky images in order to use them with the DM code.  I thought it might be just a matter of adding a `readout` section to `flat.yaml`, but it looks like some code to use the `OpsimMeta` info is needed.

config/flat.yaml
rmjarvis: Yep.  Fixed.  (This changed at some point on another branch that has since been merged.)

config/flat.yaml
jchiang87: I just tried to ingest the per-amp file into a test repo using the LSST code.  Unfortunately, it complains that it can't process the file, but it doesn't say what's missing that it expects.   The on-sky per-amp images can be ingested, so there's some subset of header keywords in those files, e.g., `RUNNUM`, that are needed for ingesting the flats.  I think we should go ahead and merge this branch and sort out with the DM folks later what keywords are needed for these files.

jchiang87 (approved):
We can sort out the header keywords for ingesting the flats later.

rmjarvis:
Sounds good.  Thanks Jim!

jchiang87:
Using this code, I found from trial and error that the following keywords need to have the following data types or values in order for the ingest to work:

| keyword |  datatype |
| ------- | ------ |
| OBSID |   int |
| MJD-OBS  | float |
| RATEL   |  float |
| DECTEL   |  float |
| ROTANGLE   |  float |
| FILTER | one of 'u', 'g', 'r', 'i', 'z', 'y' |

The others having 'N/A' as the default seems ok.   I haven't tried running the superflat generation code, so that code may have other constraints on the keywords in order for things to work, e.g., `OBSID` or `RUNNUM` probably need to be unique for a given CCD.

rmjarvis:
Is `-999` (or `-999.`) reasonable for the ones that need to be numeric?

jchiang87:
Unfortunately, using a default of `MJD-OBS = 99999` yields
```
ValueError: Timespan.end may not be later than 2100-01-01 00:00:00.000.
```
This is MJD `88069`.

jchiang87:
Also, when I run this code, the `FILTER` keyword gets written out as
```
FILTER  = '@image.wcs.band'  
```
which causes an error on ingest.

rmjarvis:
Bleh.  Too much range checking...  :)
OK.  I switched to Jan 1, 2000, which is MJD 51544.  The telescopes were all closed that day (to check for Y2K errors), so that seems like a good not real date.

cwwalter:
Wow.  I didn't know that!

rmjarvis:
I was in Chile that day...  My run started Jan 2.  Only found one Y2K bug that they missed.  The observer logs had the date as Jan 2, 19100.  🤣

imsim/readout.py
```diff
@@ -264,7 +262,23 @@ def finalize(self, config, base, main_data, logger):
...
-        hdus = fits.HDUList(get_primary_hdu(opsim_md, det_name))
+
+        # If we don't have an OpsimMeta, then skip some header items.
+        # E.g. when reading out flat field images, most of these don't apply.
+        # The only exception is filter, which we look for in config and use that if present.
+        try:
+            opsim_md = galsim.config.GetInputObj('opsim_meta_dict', config, base,
+                                                 'CameraReadout')
+        except galsim.GalSimConfigError:
+            filt = config.get('filter', 'N/A')
```
jchiang87, line 273: Should this line be something like
```
filt = base['image']['wcs'].get('band', 'N/A')
```
instead?  The current code sets `filt` to `'@image.wcs.band'`.   This is the last change needed for the ingest to work.

rmjarvis: No.  That was me being sloppy.  Sorry!  It should use ParseValue, rather than get.  Fixed I think.

jchiang87:
Thanks for the changes Mike.  The flats from this code can now be ingested, so I think we can go ahead and merge.

PR #265, opened 2021-12-17
use lowercased token so that 'sersic2D' yields the right starting index for the dust entries

jchiang87:


cwwalter:
Because GalSim uses sersic2d ?

jchiang87:
The skyCatalog instcat entries and existing phosim instcats use 'sersic2D', so this is to avoid a problem given how the dict is defined [here](https://github.com/LSSTDESC/imSim/blob/5f5290abac48d8e333941d586605b0e8d6bb7f92/imsim/instcat.py#L116).  This change is also consistent with usage [elsewhere](https://github.com/LSSTDESC/imSim/blob/5f5290abac48d8e333941d586605b0e8d6bb7f92/imsim/instcat.py#L211).

cwwalter:
OK. Got it. Change LGTM.

rmjarvis (approved):
Good.  That was my oversight.  Thanks for fixing.

PR #267, opened 2022-02-01
Use SkyCatalogs for accessing object catalogs

jchiang87:
@rmjarvis @cwwalter I think this is ready to be looked at.  To run with a recent `lsst_distrib`, one also needs ~~CCL, gcr, and gcr-catalogs~~ the [dust_extinction](https://dust-extinction.readthedocs.io/en/stable/) package installed in addition to [skyCatalogs](https://github.com/LSSTDESC/skyCatalogs).  I've included an example config file in `examples/imsim-skycat.yaml`


rmjarvis:
CCL is a pretty heavy dependency.  Are you sure we need that?  Is it just for the tests?  Or for any use of the sky catalog?

jchiang87:
@joannebogart should comment, but it seems to be needed by skyCatalogs.   It's possible that what it's used for can be replaced by something much lighter weight.

jchiang87:
It's just used [for calculating luminosity distance].(https://github.com/JoanneBogart/skyCatalogs/blob/8338571b697c760640534272757e104d18e428eb/python/desc/skycatalogs/utils/sed_utils.py#L164)

rmjarvis:
I recommend switching that over to astropy if that's all it's doing.  That's already a dependency (implicitly via GalSim if nothing else) and won't pull in Camb along with it.

jchiang87:
I was looking into using astropy last night.  I'll make a PR over in the skyCatalogs repo, but it would be good if you could have a look to make sure the cosmological parameters are incorporated correctly.  (I'll @- you.)

jchiang87:
@rmjarvis The dependence on CCL is now removed from skyCatalogs.  In the end, we just ended up doing the same thing as gcr-catalogs, which uses astropy for the cosmology.  In case you're interested, here's [the PR](https://github.com/LSSTDESC/skyCatalogs/pull/13).

BTW, I was also going to look into adding Milky Way extinction, but probably not in this PR.   How did you want to handle that?

cwwalter:
I understand why we would need gcrcatalogs to build the sky catalogs.  But, I'm not sure I understand why imSim needs it to use them?

jchiang87:
I suppose it doesn't, but it's imported by the skyCatalogs code that imSim calls.

imsim/skycat.py
```diff
@@ -0,0 +1,374 @@
...
+        self.objects = sky_cat.get_objects_by_region(region,
+                                                     obj_type_set=obj_types)
+        self._index_subcomponents()
+
+    def _index_subcomponents(self):
+        """
+        Create a list of all object subcompoents so that galsim.config
+        can use indexes corresponding the GSObjects, in contrast to
+        skyCatalog objects, which can correpond to composites of
+        GSObjects.
```
rmjarvis, line 60: Why do we want to do this?  I thought we were going to keep the subcomponents together as a single `galsim.Sum` object.  (Or `galsim.ChromaticSum` when chromatic=True.)

imsim/skycat.py
```diff
@@ -0,0 +1,374 @@
...
+                my_component = 'disk'
+            a = skycat_obj.get_native_attribute(f'size_{my_component}_true')
+            b = skycat_obj.get_native_attribute(f'size_minor_{my_component}_true')
+            assert a >= b
+            pa = skycat_obj.get_native_attribute('position_angle_unlensed')
+            if self.flip_g2:
+                beta = float(90 - pa)*galsim.degrees
+            else:
+                beta = float(90 + pa)*galsim.degrees
+            # TODO: check if should be hlr = a. See similar note in instcat.py.
```
rmjarvis, line 231: I think for the skycat stuff, we should be able to figure this out, since Joanne's numbers are closer to the original simulation code.  We just need to figure out what definition that catalog uses for these values.

imsim/skycat.py
```diff
@@ -0,0 +1,374 @@
...
+        wcs : galsim.WCS
+            WCS of the image to render.
+        obj_types : list-like [None]
+            List or tuple of object types to render, e.g., ('star', 'galaxy').
+            If None, then consider all object types.
+        edge_pix : float [100]
+            Size in pixels of the buffer region around nominal image
+            to consider objects.
+        flip_g2 : bool [True]
+            Flag to flip sign of g2.
```
rmjarvis, line 39: Please tell me we don't still have the g2 values flipped!  I would hope that this should default to False, and even better that we could remove it entirely here.

imsim/stamp.py
```diff
@@ -99,6 +99,7 @@ def setup(self, config, base, xsize, ysize, ignore, logger):
             base['current_noise_image'] = base['current_image']
             noise_var = galsim.config.CalculateNoiseVariance(base)
             keep_sb_level = np.sqrt(noise_var)/3.
+            self._large_object_sb_level = 3*keep_sb_level
```
rmjarvis, line 102: I guess this implies that this code wasn't unit tested.... Should probably add something for this.

rmjarvis (changes requested):
I didn't try running anything, so these comments are just from reading the code.

The other comment I have is that you should probably add some unit tests.  At least something that touches most of the new code.  Doesn't have to bee very comprehensive in terms of accuracy checking yet.

rmjarvis:
> I suppose it doesn't, but it's imported by the skyCatalogs code that imSim calls.

If we find it burdensome, we could probably refactor the dependencies of skyCatalogs a bit to only require gcrcatalog for the operations the build the catalog, and not the operations that read it.  I don't know how heavy a dependency that package is.

rmjarvis:
> BTW, I was also going to look into adding Milky Way extinction, but probably not in this PR. How did you want to handle that?

I have basically zero expertise with dust (of either variety), so I don't know how best to handle this.  I guess there must be some MW dust maps that we could read from and apply based on ra, dec.  So that would likely involve an input object to load in the map, and then a value type that would apply it based on the current world_pos.

examples/imsim-skycat.yaml
```diff
@@ -0,0 +1,47 @@
+# Use baseline configuration
+template: /global/homes/j/jchiang8/dev/imSim/config/imsim-config.yaml
+
+# Disable inputs from instance catalogs.
+input.instance_catalog: ""
+
+# Use skyCatalogs for obtaining the objects to render.
+input.sky_catalog:
+    file_name: /global/homes/j/jchiang8/dev/imSim/tests/data/sky_cat_9683.yaml
```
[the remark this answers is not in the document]
jchiang87: This is a local file in the repo.  I just couldn't workout the right way to specify the path to the file in the config.  How does one do that?

imsim/instcat.py
```diff
@@ -411,34 +421,6 @@ def getObj(self, index, gsparams=None, rng=None, bandpass=None, chromatic=False,
             flux = sed.calculateFlux(bandpass) * fAt
             return obj.withFlux(flux)
 
-    def getHourAngle(self, mjd, ra):
```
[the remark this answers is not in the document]
jchiang87: This method was a copy of a function in the OpsimMetaDict class, which is used.  This one, in the InstCatalog class, wasn't used.

imsim/skycat.py
jchiang87: Simply ignorance on my part.  I had no idea that `galsim.Sum` is an option.

jchiang87:
> I have basically zero expertise with dust (of either variety), so I don't know how best to handle this. I guess there must be some MW dust maps that we could read from and apply based on ra, dec. So that would likely involve an input object to load in the map, and then a value type that would apply it based on the current world_pos.

Getting the info from the dustmaps is already done in the skyCatalog code.  What we need here is something that does [this sort of calculation](https://github.com/lsst/sims_photUtils/blob/main/python/lsst/sims/photUtils/Sed.py#L921).

imsim/skycat.py
jchiang87: I would like to remove it, but it seems to be in both the shear convention and in the convention for the PA of the galaxy.  So I kept it in here for now since I wanted to compare directly to runs using instance catalogs so I could have the flexibility to toggle it and then figure out where it can be removed.

examples/imsim-skycat.yaml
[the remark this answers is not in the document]
rmjarvis: `../tests/data/sky_cat_9683.yaml` would work for the current location.  Or copy it to the config directory, and just use `sky_cat_9683.yaml`.

imsim/instcat.py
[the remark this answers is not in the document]
rmjarvis: OK.  That was probably my mistake then.  Thanks.

imsim/skycat.py
rmjarvis: OK.  Hopefully we can fix the convention upstream in the skycats themselves and remove it here.

cwwalter:
> If we find it burdensome, we could probably refactor the dependencies of skyCatalogs a bit to only require gcrcatalog for the operations the build the catalog, and not the operations that read it. I don't know how heavy a dependency that package is.

I think it would be good if we could do that.  It not a package I normally have installed for example.  Sticking to the stack packages would be good if we can.

jchiang87:
@rmjarvis I think this is ready to be looked at again.

.github/workflows/ci.yml
```diff
@@ -58,13 +58,19 @@ jobs:
...
+                cd ..
+                git clone https://github.com/LSSTDESC/skyCatalogs.git
+
             - name: Install test deps
               run:
                 conda install -y pytest nose
 
             - name: Run tests
               run: |
+                setup -r ../skyCatalogs -j
```
rmjarvis, line 72: Any chance we can get skyCatalogs pip installable rather than using eups?  This is likely to be a blocker for some users down the line.  We don't want to force everyone to clone the repo to access it.

config/imsim-config.yaml
```diff
@@ -291,5 +285,5 @@ output:
             dec: "$sky_pos.dec.deg"
             x: "$image_pos.x"
             y: "$image_pos.y"
-            flux: "$(@current_prof).flux"
+            flux: "$(@current_prof).calculateFlux(bandpass)"
```
rmjarvis, line 288: We probably don't want this I suspect.  This does a non-trivial calculation for every object, so it's going to slow things down a lot.  Probably better to just remove the flux column in the output.  Not sure if there is something appropriate we can replace it with that would just be a look-up, not a calculation.

examples/imsim-skycat.yaml
```diff
@@ -0,0 +1,47 @@
...
+image.random_seed: 42
+
+image.wcs.boresight:
+    type: RADec
+    ra:
+        type: Degrees
+        theta: { type: OpsimMeta, field: fieldRA }
+    dec:
+        type: Degrees
+        theta: { type: OpsimMeta, field: fieldDec }
```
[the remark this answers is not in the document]
rmjarvis: I still think this should be swapped in the original rather than replaced here.  I.e. Have `image.wcs.boresight` in imsim-config.yaml use this specification.  And then have `input.atm_psf.boresight` be `'@image.wcs.boresight'`.  Then, this override in the user config could be removed.

imsim/instcat.py
```diff
@@ -859,4 +825,3 @@ def getKwargs(self, config, base, logger):
 RegisterValueType('InstCatWorldPos', InstCatWorldPos, [CelestialCoord],
                   input_type='instance_catalog')
 RegisterObjectType('InstCatObj', InstCatObj, input_type='instance_catalog')
-RegisterSEDType('InstCatSED', InstCatSEDBuilder(), input_type='instance_catalog')
```
rmjarvis, line 862: I think we probably want to keep this for now, don't we?  I thought we wanted to keep the functionality of using instance catalogs, and this is the way you get an SED with one of those.

imsim/stamp.py
```diff
@@ -407,20 +424,15 @@ def draw(self, prof, image, method, offset, config, base, logger):
...
         # XXX: Make this a parameter?  Or ok to leave like this?
         # Note: if it's just 1, it would be simpler to just remove it in this function.
         gain = 1.
 
-        if faint or 'sed' not in config:
-            sed = self._trivial_sed
-        else:
-            sed = galsim.config.BuildSED(config, 'sed', base, logger=logger)[0]
-        base['current_sed'] = sed
-
```
rmjarvis, line 423: Did you preserve the essence of this functionality somewhere?  I don't see it.  We want to make sure the faint objects don't bother with all the chromatic stuff.

.github/workflows/ci.yml
jchiang87: I'll check with Joanne.  Should be doable.

config/imsim-config.yaml
jchiang87: The revised code _is_ noticeably slower than before.  I think we can just omit it.

imsim/instcat.py
jchiang87: Yes, of course.  I'll add it back in.

imsim/stamp.py
jchiang87: > Did you preserve the essence of this functionality somewhere?

I guess I don't know off hand how to do that.

config/imsim-config.yaml
rmjarvis: That may also be because you aren't doing the fast thing for the faint objects.  (Depending on the input catalog how many faint objects you are using.)

imsim/skycat.py
```diff
@@ -0,0 +1,355 @@
...
+        ----------
+        skycat_obj : skyCatalogs.BaseObject
+            A skyCatalogs object, e.g., a star or galaxy.
+        component : str
+            The name of the sub-component of the skyCatalogs object
+            to consider.
+        """
+        sed, magnorm = self.getSED_info(skycat_obj, component)
+        if sed is None or magnorm >= 50:
+            return None
```
rmjarvis, line 199: This is probably where we want to do the faint check.  We may not be able to use realized_flux anymore, since that's currently a non-trivial calculation using the (multiple) SEDs, so we may need to base it on magnorm I guess?  But with some definition of "faint" here, we can use the trivial sed for any faint component.

imsim/skycat.py
```diff
@@ -0,0 +1,355 @@
...
+                               component)
+
+        # The seds are normalized to correspond to magnorm=0.
+        # The flux for the given magnorm is 10**(-0.4*magnorm)
+        # The constant here, 0.9210340371976184 = 0.4 * log(10)
+        flux = math.exp(-0.9210340371976184 * magnorm)
+
+        # This gives the normalization in photons/cm^2/sec.
+        # Multiply by area and exptime to get photons.
+        fAt = flux * self._rubin_area * exp_time
```
rmjarvis, line 246: Actually, here is the right place I think.  This is the relevant flux.  We can then even do the poisson realization of it here and do `withFlux` with that value.  And if the realized flux is what below we call faint, use the trivial sed rather than making the right one.

But that means that we need to split up the get_SEDinfo method to have a fast way to get magnorm first that doesn't do any complex calculations (hopefully none are done internally in skyCatalog to get magnorm), and then only if we care about the real SED do the full calculation with redshift and extinction and all that.

imsim/skycat.py
```diff
@@ -0,0 +1,355 @@
...
+
+        # Apply Milky Way extinction
+        extinction = F19(Rv=mwRv)
+        wl = np.linspace(300, 1200, 901)
+        ext = extinction.extinguish(wl*u.nm, Av=mwAv)
+        spec = galsim.LookupTable(wl, ext)
+        mw_ext = galsim.SED(spec, wave_type='nm', flux_type='1')
+        sed = sed*mw_ext
+
+        sed = sed.thin()
```
rmjarvis, line 114: I had a comment before about removing this line, which seems to be gone now.  This is a very non-trivial calculation.  If we do it on every object, it will be faster to not do it. It's only useful if you do it once on and SED and then reuse that SED repeatedly.

I think what you want here is to use sed.wave_list rather than a new `np.linspace`, which you then have to thin.

cwwalter:
Hi Jim, sorry.. some of the things before have now been changed so I think I can't easily reply inline to them. 

But, I was wondering. If we there is code living in skycat.py for dust extinction, check from simple SED etc: can we arrange it to also use those functions in the instCat code so that we don't have duplicates?  Or maybe use class inheritance so that we only override what is needed in one of them?

jchiang87:
> But, I was wondering. If we there is code living in skycat.py for dust extinction, check from simple SED etc: can we arrange it to also use those functions in the instCat code so that we don't have duplicates? Or maybe use class inheritance so that we only override what is needed in one of them?

We do need to put the common functionality in skycat.py and instcat.py in one place and avoid duplication, but I think it would be easier to do that as a separate refactoring PR after we've sorted out the desired functionality for the skycat.py module.  There are significant differences between how the skycat and instcat modules handle galaxies, i.e., as composite objects in the skycat case, that I'd rather defer that refactoring until after we've decided that we have everything we need for skycat.py

cwwalter:
Sounds good. That makes sense, thanks.

.github/workflows/ci.yml
rmjarvis: This line doesn't seem to have worked on the CI.  It's not able to import skyCatalogs currently.

imsim/instcat.py
```diff
@@ -274,10 +286,23 @@ def getSED(self, index):
...
+        sed = sed.atRedshift(redshift)
+
+        # Apply Milky Way extinction.
+        extinction = F19(Rv=mwRv)
+        wl = np.linspace(300, 1200, 901)
+        ext = extinction.extinguish(wl*u.nm, Av=mwAv)
+        spec = galsim.LookupTable(wl, ext)
+        mw_ext = galsim.SED(spec, wave_type='nm', flux_type='1')
+        sed = sed*mw_ext
+        sed = sed.thin()
```
rmjarvis, line 303: I mentioned this before.  Don't do this.  It's super slow.  Instead, use `wl=sed.wave_list` rather than make a new linspace.

jchiang87: right, sorry, I fixed it in the skycat.py module, but neglected to fix it here.

.github/workflows/ci.yml
jchiang87: That's because skyCatalogs is trying to `import GCRCatalogs` at the top of the module.  I've already asked Joanne to move that import into the function that uses it so that it doesn't affect things here.

imsim/stamp.py
```diff
@@ -44,9 +44,16 @@ def setup(self, config, base, xsize, ysize, ignore, logger):
...
 
-        # Check if the realized flux is 0.
+        # Compute or retrieve the realzed flux.
         self.rng = galsim.config.GetRNG(config, base, logger, "LSST_Silicon")
-        self.realized_flux = galsim.PoissonDeviate(self.rng, mean=gal.flux)()
+        bandpass = base['bandpass']
+        if hasattr(gal, 'flux'):
+            flux = gal.flux
+        else:
+            flux = gal.calculateFlux(bandpass)
```
rmjarvis, line 53: This is the slow thing that we want to cache in skyCatalog.  I guess that will be a different PR, but for now maybe add a TODO comment to remind us that we know this is going to dominate the running time for faint objects.

imsim/stamp.py
```diff
@@ -89,7 +96,10 @@ def setup(self, config, base, xsize, ysize, ignore, logger):
             # For extended objects, recreate the object to draw, but
             # convolved with the faster DoubleGaussian PSF.
             psf = self.DoubleGaussian()
-            obj = galsim.Convolve(gal, psf).withFlux(self.realized_flux)
+            # For Chromatic objects, need to evaluate at the
+            # effective wavelength of the bandpass.
+            obj = galsim.Convolve(gal, psf).withFlux(self.realized_flux, bandpass)
+            obj = obj.evaluateAtWavelength(bandpass.effective_wavelength)
```
rmjarvis, line 102: The other way I think.
```
gal_achrom = gal.evaluateAtWavelength(bandpass.effective_wavelength)
obj = galsim.Convolve(gal_achrom, psf).withFlux(self.realized_flux)
```

.github/workflows/ci.yml
JoanneBogart: That change was merged to master earlier this week.

JoanneBogart: There is a file scripts/write_sed_files.py which still has an import of GCRCatalogs at the top level.
Could this be the problem?  I can similarly reorganize so that the import will be buried in a function if that would help.

rmjarvis: Yep.  Looks like it works now.  I restarted the tests and they passed.  Thanks Joanne!

rmjarvis: I don't think the imsim usage cares about what's in the scripts.

JoanneBogart: Actually I think I can get rid of the import altogether.

JoanneBogart: I've committed changes to skyCatalogs which eliminate that import

jchiang87:
@rmjarvis I think I've addressed your recent comments.  Do you have any for https://github.com/LSSTDESC/imSim/blob/u/jchiang/skyCatalogs_interface/imsim/skycat_object_wrapper.py?  This is the code we are thinking to move to skyCatalogs so that it can do the flux calculations and caching there.

rmjarvis (approved):

rmjarvis:
This looks good to me.  I'm sure we'll run across issues when we start using it in anger, but I don't see anything that should prevent us from merging.

Issue #10, opened 2016-10-18
Add optical raytracing


cwwalter:
Some people (@dkirkby?, etc)   have early codes to do this based on open raytracers.  

If this is possible where would it live?  In GalSim proper? As a standalone package that could be imported?

This would make matching a ray traced atmosphere and sensor model much easier and would also allow for simulating things like the CBP.

cwwalter:
Noting that @jmeyers314 now has started external python controlled code that does this.

cwwalter:
Added list of optical effects we hope to simulate with Batoid to the main issue post.  For now these don't have their own issues, but let's track desired functionality here.

cwwalter:
We are discussing various approaches to the spider and diffraction spikes.  There are two main possible approaches: extra kicks in batoid and really calculating the diffractive PSF and using that same PSF contribution over the entire sensor (as is done in Roman).  There was some discussion of the diffractive calculation previously in the GalSim repo that I am tagging here for reference.  See the discussions with @rmandelb near the end of.

https://github.com/GalSim-developers/GalSim/issues/556
 
I will separately share some literature on the kick approach.

cwwalter:
@jmeyers314 and @g-braeunlich , with this merge, should vignetting now be showing up emergently?

jmeyers314:
Vignetting of sources should show up I think.  Vignetting of the background won't though.

cwwalter:
Right.. That will be a good check.

cwwalter:
Update:  In addition to vignetting, we now have diffraction spikes including accounting for camera rotations relative to the spider (Both here with batoid and also parametrically for the the bright FFT case).

Ghosts and scattered light (which will be particularly important for simulating CBP calibration) will wait until the redesign of the parallelization strategy which will allow us to track light which spans multiple sensors via a 2nd pass.

PR #268, opened 2022-04-20
batoid raytracing

g-braeunlich:
This is related to  #10


tests/test_batoid_wcs.py
```diff
@@ -597,8 +598,46 @@ def test_config():
...
+            tests = super().loadTestsFromModule(module, *args, pattern=pattern, **kws)
+            test_functions = (
+                f
+                for f_name, f in globals().items()
+                if f_name.startswith("test_") and callable(f)
+            )
+            tests.addTests([unittest.FunctionTestCase(f) for f in test_functions])
+            return tests
+
+    unittest.main(testLoader=TestLoader())
```
rmjarvis, line 643: Why all this unittest boilerplate?  I believe this is functionally equivalent to:
```
test_functions = (
    f 
    for f_name, f in globals().items() 
    if f_name.startswith("test_") and callable(f))
for test_fn in test_functions:
    test_fn()
```
which is simpler and also doesn't activate my distaste for the unittest module...  :)

config/imsim-config.yaml
```diff
@@ -203,6 +203,10 @@ stamp:
             HA:
                 type: Degrees
                 theta: { type: OpsimMeta, field: HA }
+        -
+            type: lsst_photon_op
```
rmjarvis, line 207: lsst_optics

imsim/batoid_wcs.py
```diff
@@ -31,6 +32,45 @@
...
+    -------
+    x, y : array
+        Pixel coordinates.
+    """
+    tx = det.getTransform(cameraGeom.FOCAL_PLANE, cameraGeom.PIXELS)
+    x, y = tx.getMapping().applyForward(np.vstack((fpx, fpy)))
+    return x.ravel(), y.ravel()
+
+
+def pixel_to_focal(x, y, det):
```
rmjarvis, line 55: These two functions probably ought to move to your new batoid_utils file.

tests/test_batoid_wcs.py
```diff
@@ -597,8 +598,46 @@ def test_config():
...
+    x = y = np.linspace(start=-100.0, stop=100.0, num=4)
+    x_t, y_t = batoid_wcs.pixel_to_focal(
+        np.tile(x, len(y)), np.repeat(y, len(x)), det
+    )
+
+    x_expected = np.linspace(start=-190.615, stop=-188.615, num=4)
+    y_expected = np.linspace(start=-317.255, stop=-315.255, num=4)
+    np.testing.assert_almost_equal(x_t, np.tile(x_expected, len(y)))
+    np.testing.assert_almost_equal(y_t, np.repeat(y_expected, len(x)))
+
```
rmjarvis, line 625: Need some kind of test of the new photon operator.

Also, you seem to have added an `__init__.py` file in the tests directory, which doesn't belong here.

tests/test_batoid_wcs.py
g-braeunlich: I just wanted to benefit from `unittest` (it prints what failed, what it runs, ...). And all the rest of the test files in imsim also are using the module.
But of course, we also can use a different test loader. I dont care

config/imsim-config.yaml
g-braeunlich: Thx

tests/test_batoid_wcs.py
g-braeunlich: The `__init__.py` file turns the `tests` dir into a python module. This simplifies test discovery and python module discovery.
For example, a `python -m unittest tests/test_batoid_wcs.py` would only work with an `__init__.py` in the `tests` dir, as otherwise, you would get `ModuleNotFoundError: No module named 'tests.test_batoid_wcs'`

imsim/batoid_wcs.py
g-braeunlich: I did not move them, because they are not batoid specific and I did not want to initiate a big module messed up with tons of unrelated things in it. I like the idea of grouping things together which also change together (it is a coordinate transform which I think fits well into "wcs").
Of course we also could argue about the new `batoid_utils` module. I am also open here.

rmjarvis: OK.  It seems to me that these are both utility functions used by multiple batoid-related modules, so would make sense to be in `batoid_utils`.  Or possibly in a non-batoid-specific `utils`.  Seems cleaner to me than having `photon_ops` load some things from `batoid_utils` and some from `batoid_wcs`.  
But I don't care that much if you prefer to keep them here.

tests/test_batoid_wcs.py
g-braeunlich: The boilerplate code is necessary to let the `unittest` module discover functions.
I really understand your distaste for `unittest` just because of the fact that this is not the default of `unittest`.
But this test file is the only one in imsim which does not pack the tests in `unittest.TestSuite` classes. We have to do this manually to mimic the behaviour of pytest

tests/test_batoid_wcs.py
g-braeunlich: OK. Just thought that adding a `__init__.py` would not hurt and enable `unittest` support. Note, that this practice is also described here: https://docs.pytest.org/en/6.2.x/goodpractices.html
If you want me to remove it again, I ll of course do that.

g-braeunlich: Also note: https://docs.python.org/3/library/unittest.html#test-discovery
(for the case we want to stick with `unittest`)
And I also saw, that my boilerplate code is not sufficient for test discovery. I would need to do something like:
https://docs.python.org/3/library/unittest.html#load-tests-protocol

rmjarvis: If it were up to me, I'd rip out all the unittest stuff, and just have test functions, calling them with either pytest or nosetests, not the unittest module.  That's by far my preferred structure for test suites.
But as you noted, that's not what we seem to be mostly doing in imsim so far.  So I'll defer to @jchiang87's opinion on this, since I think he made the initial structure here.

tests/test_batoid_wcs.py
g-braeunlich: For now, I moved as much boilerplate code as I could to `tests.__init__` and use it in `test_batoid_wcs` and `test_utils`.

imsim/batoid_wcs.py
g-braeunlich: created a new `utils` module.

tests/test_batoid_wcs.py
g-braeunlich: Do you have a suggestion for a test for the photon op?
Test some very simple batoid telescope?

g-braeunlich: `__init__.py` and other unittest boilerplate removed in the tests for `batoid_wcs` and the new photon op

imsim/photon_ops.py
```diff
@@ -0,0 +1,140 @@
...
+    """Converts a batoid.RayVector to a galsim.PhotonArray
+
+    Stores into an already existing galsim.PhotonArray out"""
+
+    assert all(np.abs(ray_vector.z) < 1.0e-15)
+    out.x, out.y = focal_to_pixel(ray_vector.x * 1e3, ray_vector.y * 1e3, detector)
+
+    out.dxdz = ray_vector.vx / ray_vector.vz
+    out.dydz = ray_vector.vy / ray_vector.vz
+    return out
```
jmeyers314, line 115: I think we need to add something like `out.flux *= (1-ray_vector.vignetted)` here.  (My fault for forgetting this before).

g-braeunlich: Should be done now

jmeyers314:
Still looks like we need a unit test that uses the new photon op.  It's probably sufficient to just make sure that it runs, and not worry too much here if it produces the correct output.

imsim/photon_ops.py
rmjarvis: Is that right?  I would have thought probabilistically remove photons (or set to flux=0) rather than change their fluxes to fractional values.

rmjarvis: Or is ray_vector.vignetted just a 0 or 1?  If that's the case, I think this is fine.

jmeyers314: `ray_vector.vignetted` is a bool, so this is just zeroing the flux for geometrically obstructed paths.  Maybe there's a np.where() clause that would make this more obvious?

jmeyers314: `out.flux = np.where(ray_vector.vignetted, 0, out.flux)` may be clearer, but looks like it allocates instead of modifies in place.

rmjarvis: Maybe `out.flux[ray_vector.vignetted] = 0.`?

rmjarvis: But the existing one is fine too, now that I know it's a bool.

g-braeunlich: I changed it to `out.flux[ray_vector.vignetted] = 0.` which I think is clearer.

g-braeunlich:
@jmeyers314 I added a unittest.
For now, I just feed in data, I collected during a run of `galsim imsim-user.yml`.
Do you think this is fine, or should we setup the environment differently?

g-braeunlich:
Thanks @jmeyers314 for the fixes.
I squashed (parts of)  your commits together.
I also could get rid of
```py
if isinstance(wcs, tuple):
   wcs = wcs[0]
```
in [tests/test_instcat_parser.py](tests/test_instcat_parser.py)

jchiang87:
@rmjarvis @jmeyers314 @g-braeunlich Is this PR ready to merge?  I've got some development that relies on the changes to the batoid_wcs.py code in order for the DM code to be able to process imSim images.

rmjarvis (approved):
Seems good to me.

PR #269, opened 2022-04-22
Add pyproject.toml with pytest config

g-braeunlich:
* Make tests independent of working dir
* Tests can be run by just invoking "pytest" in ci and dev environment
* blacklist skipped tests instead of whitelisting tests (new tests are run by default)


rmjarvis:
> Make tests independent of working dir

Why?  We only need tests to run from the tests directory.  There's no need to make them runnable from arbitrary working directories.

> Tests can be run by just invoking "pytest" in ci and dev environment

Again, why?  `cd test; pytest` is our standard workflow now.  I don't see why we should change that.
In particular, the downside of running pytest from the main directory is that it looks for test files in all subdirectories, including any development sandbox space.  We don't have that currently in imsim, but in my other projects, I typically have a `devel` directory where people can write more exploratory tests to try out new ideas.  Some of these take a long time to run, which is fine for exploratory development, but not for CI.

> blacklist skipped tests instead of whitelisting tests (new tests are run by default)

We are working towards having all tests working, in which case we don't need to whitelist or blacklist.  So this isn't a longterm problem.  Just temporary until we get the sims stuff from LSST working, which I think is the blocker for the only ones we haven't enabled yet.  (Some are enabled in other PRs, btw.)

So TL;DR, I guess I'm not particularly in favor of these changes.

g-braeunlich:
I just did this because:
`pytest` is simpler than `cd test; pytest`, it integrates better into an IDE and it simplifies things for newcomers like me.

Note that I defined
```toml
[tool.pytest.ini_options]
testpaths = [
    "tests",
]
```
in `pyproject.toml`. Therefore only tests in this directory are discovered.

Regarding blacklisting instead of whitelisting: even though this is temporary, it is useful just now, and will be if we want to add a new test file for `photon_ops.py`.

Again: you are the boss. If you dont like it, I ll remove it again.

rmjarvis:
I'm not the boss.  Just giving my opinion.  The main thing I don't like is all the extra work to avoid using straightforward directory names in the test functions. (e.g. `'data'` -> `os.path.join(os.path.dirname(__file__), 'data')`)

g-braeunlich:
Would you be happy with
```py
from pathlib import Path

IMSIM_DIR = Path(__file__).parent

and then `'data'` -> `IMSIM_DIR / 'data'`.
```

This also would make life easier if you would decide to move the data dir around.

g-braeunlich:
Or put differently: a variable is much easier to refactor than a string

rmjarvis:
I've stated my opinions.  I feel like I should let someone else weigh in on this at this point.

cwwalter:
Hi All,

One of our guiding design principles for imSim is that it should live in the Rubin and DESC software environments and, to the extent possible, follow the same conventions as used there.  This way it will work (and the design patterns will be like) lik e the other software packages people are using.

Not only does this make it easier for people to understand what is going on, but it makes development tools also work without having to make the setup for imSim "Special". I of course realize for someone working only on imSim this is not a driving concern, but we are striving to make the program as easy to use and understand across the entire Rubin community as possible.

So: I was also under the impression that `cd` into the test directory and running the tests there was the standard pattern (for me it also has the advantage that any output that gets created is nicely contained there and I can find and delete it).  But, things have also been changing and I might not be right.

So, @jchiang87 can you comment on how this is typically setup in both DESC and Rubin code.  Is there a mostly standard approach being used in each?

jchiang87:
For Rubin DM code, the tests are typically run from the top-level directory of the package.  This is what I'm accustomed to, and it's how I've set things up for the Camera team code as well as for other packages I've worked on for DESC including imSim v1.0.0 and earlier.   That said, I don't think there is an established convention in DESC.   I don't personally have a strong preference, but then I don't use an IDE where it might make a difference.

g-braeunlich:
From my side: I dont want to be the "arrogant newcomer" dictating new rules. It is actually a detail and I wont be mad if you tell me to roll back.
As I just did not see this before in projects I contributed so far (and also [astropy](https://docs.astropy.org/en/latest/development/testguide.html) seems to run tests from the toplevel dir), I thought I ll do a PR.

cwwalter:
Hi Jim,

I think we would like this to operate just as other Rubin packages since we hope we will be able to use this in the with the commissioning pipelines and possibly be deployed. And, it sounds like it is how you had imSim v1.0 setup before.

But, I'm still a bit unclear on how different that is than we have now, or if you need to make a change just like this one to make it happen.

Should we back this change out to make that happen and have you do it another way? Or is this actually  how you had it setup?

jchiang87:
Geri's initial comment pretty much summarizes the differences:  Using this branch (or using imSim v1.0.0), one can run `pytest` from the top level directory and the tests run without errors. Explicitly, this is what I see using this branch:
```
[w_2022_17] pytest
=============================================== test session starts ================================================
platform linux -- Python 3.8.13, pytest-7.1.1, pluggy-1.0.0
rootdir: /home/DESC/tmp/imSim, configfile: pyproject.toml, testpaths: tests
plugins: typeguard-2.13.3, arraydiff-0.3, filter-subpackage-0.1.1, openfiles-0.5.0, session2file-0.1.11, flake8-1.1.0, cov-3.0.0, hypothesis-6.40.0, forked-1.4.0, remotedata-0.3.3, subtests-0.7.0, doctestplus-0.12.0, asdf-2.10.1, mock-3.7.0, astropy-header-0.1.2, xdist-2.5.0
collected 30 items                                                                                                 

tests/test_FWHMgeom.py ...                                                                                   [ 10%]
tests/test_atmPSF.py .                                                                                       [ 13%]
tests/test_batoid_wcs.py ....                                                                                [ 26%]
tests/test_cosmic_rays.py ...                                                                                [ 36%]
tests/test_fopen.py .                                                                                        [ 40%]
tests/test_instcat_parser.py .......                                                                         [ 63%]
tests/test_optical_zernikes.py ......                                                                        [ 83%]
tests/test_psf.py ..                                                                                         [ 90%]
tests/test_tree_rings.py .                                                                                   [ 93%]
tests/test_trimmer.py ..                                                                                     [100%]

================================================= warnings summary =================================================
tests/test_instcat_parser.py::InstanceCatalogParserTestCase::test_metadata_from_opsim_db
  /opt/lsst/software/stack/conda/miniconda3-py38_4.9.2/envs/lsst-scipipe-3.0.0/lib/python3.8/site-packages/astropy/coordinates/builtin_frames/utils.py:68: AstropyWarning: Tried to get polar motions for times after IERS data is valid. Defaulting to polar motion from the 50-yr mean for those. This may affect precision at the arcsec level. Please check your astropy.utils.iers.conf.iers_auto_url and point it to a newer version if necessary.
    warnings.warn(wmsg.format('after'), AstropyWarning)

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
==================================== 30 passed, 1 warning in 109.29s (0:01:49) =====================================
[w_2022_17] 
```
By contrast, running `pytest` using current main from the top level directory, the old, un-converted tests fail because they're still trying to `import desc.imsim`, and if `pytest` tried to run the other, updated tests, some of them would fail because they can't find some input data.  If I cd to the `tests` folder, the updated tests would find the input data, but the older tests would still fail.

I don't think going back to what's in imSim v1.0.0 makes sense since that setup relies on the directory structure that we had back then to some extent, as well as on environment variables that are available when a package is set up with eups.   Anyway, from what I'm seeing, Geri's branch would get us closer to what other Rubin packages are doing.

cwwalter:
OK I see. Thank you.

So, Jim could you review the PR and see if there is anything to change to make it more consistent with the Rubin setup?

(especially related to the CI setup, the DATA_DIR or path joins etc.  I noticed some inconsistencies with the use or not of DATA_DIR in the tests BTW).

After that, and if we make any changes because of it, I propose we use it.

I'm happy to hear more opinions about this too.

g-braeunlich:
@cwwalter
* Do you want me to move `shear_test_instcat.txt` and `tiny_instcat.txt` into `data`?
* Do you prefer pathlib over `os.path.join` and co?

cwwalter:
I'd like to hear from @jchiang87 and have him give guidance as to how to make it most like the camera/DM package tests.

jchiang87 (approved):
Thanks, Geri.  LGTM.

jchiang87:
I resolved the conflict.  If it passes ci, then I think we can go ahead and merge.

jchiang87:
Yes, I see it. Probably easiest to skip that test and fix it later.

cwwalter:
So, go ahead and merge this first?

pyproject.toml
```diff
@@ -0,0 +1,19 @@
...
+# We're working towards getting all the tests working, but so far these are
+# the ones that dont work with the pip installation:
+addopts = """
+--ignore=tests/test_bleed_trails.py
+--ignore=tests/test_skyModel.py
+--ignore=tests/test_camera_readout.py
+--ignore=tests/test_get_stack_products.py
+--ignore=tests/test_flats.py
+--ignore=tests/test_imageSimulator.py
+--ignore=tests/test_raw_file_writing.py
```
jchiang87, line 18: Hi @g-braeunlich,
Could you add
```
--ignore=tests/test_skycat.py
```
here please?   Fixing this test to use the `pathlib` stuff is a bit more complicated because there is a path in `tests/data/sky_cat_9683.yaml` that the skyCatalogs code uses that can't make use of this mechanism, so I think some changes in skyCatalogs need to be made also.

g-braeunlich: If some function is not `pathlib.Path` aware, it should be in principle as simple as converting the `Path` obj to a `str`.
I tried that and pushed. If this does not work, I ll have another look tomorrow. Sorry, that I forgot to watch out for new commits!

g-braeunlich: Also had to change the path
`root_directory: ../tests/data`
to
`root_directory: ./tests/data`

in [tests/data/sky_cat_9683.yaml](tests/data/sky_cat_9683.yaml)

jchiang87: Thanks, that will work of course for the GHA testing.  I've been discussing with Joanne changes to skyCatalogs itself regarding how that `root_directory` entry would be handled so that one could  run `pytest` either from the top-level `imSim` folder or from the `tests` subfolder.

PR #272, opened 2022-05-27
check if original object is a DeltaFunction for stamp size calculation

jchiang87:
I neglected to update this line when we switched to `ChromaticObject`s.  Without this fix, all stars with `flux >= 10` end up with `image_size=16`.


imsim/stamp.py
```diff
@@ -68,7 +68,8 @@ def setup(self, config, base, xsize, ysize, ignore, logger):
             # For really faint things, don't try too hard.  Just use 32x32.
             image_size = 32
 
-        elif isinstance(gal, galsim.DeltaFunction):
+        elif (not isinstance(gal, galsim.ChromaticSum)
```
rmjarvis, line 71: I assume this part is just to make sure `gal._original` doesn't give an AttributeError.  So probably better to write this as:
```
elif (hasattr(gal, 'original') and isinstance(gal.original, galsim.DeltaFunction)):
```
Also I removed the leading underscore.  The intended API is for users to use `gal.original`, which is a read-only property returning what is currently implemented as a private attribute, `gal._original`.

rmjarvis (changes requested):
Looks good.  Glad to see that skyCatalogs doesn't need eups anymore.  :)
I just have a slight edit suggestion, which I think should be more robust to possible future updates.

imsim/stamp.py
jchiang87: Thanks Mike!  I've implemented this change.

rmjarvis (approved):

PR #276, opened 2022-07-15
Add statistical diffraction

g-braeunlich:


cwwalter:
Associated with #10.

imsim/diffraction.py
```diff
@@ -0,0 +1,166 @@
...
+LSST_SPIDER_GEOMETRY = Geometry(
+    thick_lines=np.array(
+        [
+            [1 / np.sqrt(2.0), 1 / np.sqrt(2.0), -0.4, 0.025],
+            [-1 / np.sqrt(2.0), 1 / np.sqrt(2.0), -0.4, 0.025],
+            [1 / np.sqrt(2.0), 1 / np.sqrt(2.0), 0.4, 0.025],
+            [-1 / np.sqrt(2.0), 1 / np.sqrt(2.0), 0.4, 0.025],
+        ]
+    ),
+    circles=np.array([[0.0, 0.0, 2.3905], [0.0, 0.0, 4.35], [0.0, 0.0, 2.4005]]),
```
jmeyers314, line 41: I think we just want radii of 4.18 and 2.558 (the inner/outer radius of M1) for now.  We'll need to think harder about this to get off-axis right, but we can postpone that for now.

imsim/diffraction.py
```diff
@@ -0,0 +1,166 @@
...
+        random values for phi
+    """
+
+    d, n = directed_dist(geometry, pos)
+    d_tan_phi = distribution(phi_star(d, wavelength))
+    return d_tan_phi[:, None] * v_z[:, None] * n
+
+
+def phi_star(delta: np.ndarray, wavelength: np.ndarray) -> np.ndarray:
+    """According to https://ntrs.nasa.gov/citations/19990094899
```
jmeyers314, line 75: Add something to the doc string stating this is the standard deviation in radians for a  deflection.

imsim/photon_ops.py
```diff
@@ -64,6 +65,14 @@ def __init__(
         self.sky_pos = sky_pos
         self.image_pos = image_pos
         self.icrf_to_field = icrf_to_field
+        deviate = GaussianDeviate()
+
+        def diffraction_rng(phi):
+            sigma = phi.copy()
+            deviate.generate_from_variance(sigma)
```
jmeyers314, line 72: variance is `sigma**2` I believe.

tools/LSST_spider_2d_geometry.py
```diff
@@ -0,0 +1,141 @@
...
+    rectangles = [
+        component for component in components if isinstance(component, ObscRectangle)
+    ]
+    rectangles = [
+        rect
+        for rect in rectangles
+        if rect.width < strut_thickness or rect.width < strut_thickness
+    ]
+
+    annuli = [
```
jmeyers314, line 44: I think this is probably missing the `ClearAnnulus` obscuration type (which, to be fair, is not well-documented in batoid).

jmeyers314 (approved):
This looks good to me Geri.  I think you just missed a stddev -> variance calculation, which is why so much energy ended up in the spikes initially.  If you make this fix, then only ~1.5% ends up in the spikes which is much more in line with my intuition for spikes.

I think we might also want to add a keyword to `LsstOptics` making the diffraction optional.

jmeyers314:
I made one tiny addition to ignore any sanity checks for vignetted rays.  I think this is ready to merge @g-braeunlich .

PR #278, opened 2022-07-17
add sky background calculation using rubin_sim.skybrightness code

jchiang87:


imsim/sky_model.py
```diff
@@ -0,0 +1,113 @@
...
+
+    wcs = galsim.config.BuildWCS(base['image'], 'wcs', base)
+    value = sky_model.get_sky_level(wcs.center.ra/galsim.degrees,
+                                    wcs.center.dec/galsim.degrees)
+
+    return value, safe
+
+
+RegisterInputType('sky_model', SkyModelLoader(SkyModel, takes_logger=True))
+RegisterValueType('sky_level', sky_level, [float], input_type='sky_model')
```
rmjarvis, line 113: All the existing types so far are capitalized (CamelCase), so I would call this SkyLevel to match current conventions.

imsim/sky_model.py
```diff
@@ -0,0 +1,113 @@
...
+def sky_level(config, base, value_type):
+    """
+    Use the rubin_sim skybrightness model to return the sky level in
+    photons/arcsec^2 at the center of the current WCS.
+    """
+    sky_model = galsim.config.GetInputObj('sky_model', config, base, 'sky_level')
+
+    kwargs, safe = galsim.config.GetAllParams(config, base)
+
+    wcs = galsim.config.BuildWCS(base['image'], 'wcs', base)
```
rmjarvis, line 105: I think `wcs = base['wcs']` should work here.  At the time when we calculate the sky level, the wcs should already be built and saved.  No need to remake it here.
Edit: given the below comment, I don't think you even need the wcs here.

imsim/sky_model.py
```diff
@@ -0,0 +1,113 @@
...
+    value = sky_model.get_sky_level(wcs.center.ra/galsim.degrees,
+                                    wcs.center.dec/galsim.degrees)
```
rmjarvis, line 107: A few things here.
First, `wcs.center` is not necessarily anything we want to use here.  It's a arbitrary location that doesn't have to have any connection to the image.  I think what we want is `base['world_center']`.
Second, I would suggest passing this as a CelestialCoordinate, rather than convert to ra, dec in degrees and then in the other function have LSST sims use degrees.  That way the two uses of degrees, which need to match, are in the same function.
Finally, a slightly simpler syntax to get the angle in degrees is just `ra.deg`.

tests/test_sky_model.py
```diff
@@ -0,0 +1,34 @@
+import numpy as np
+import galsim
+from imsim import SkyModel
+
+
+def test_sky_model():
```
rmjarvis, line 6: This test looks fine.  But I'll note that we already had a file test_skyModel.py.  Is anything in there reusable, modifying to the new syntax?  If not we should either remove it or mark which ones are still aspirational.

rmjarvis (changes requested):
Thanks for getting this working Jim.  I do have a few suggestions, but it basically looks good.

tests/test_sky_model.py
jchiang87: I was thinking of just removing test_skyModel.py since almost all of the tests are specific to how that old skyModel.py code was written, i.e., using that `NoiseAndBackgroundBase` class from sims_GalSimInterface.  So, I'll just go ahead and remove it.

rmjarvis (approved):

jchiang87:
@rmjarvis Before merging, I'm thinking it might be good to move the `from rubin_sim import sky_brightness` line from the top of sky_model.py to inside the `SkyModel.__init__` function.  This way, `rubin_sim` doesn't need to be installed, and one can just set
```
input.sky_model: ""
image.sky_level:  <some value>
```
and still run the code.  What do you think?

rmjarvis:
Yeah, that's probably a good idea.  At least until rubin_sim gets a bit easier to install.

jchiang87:
Unless I hear otherwise, I'll merge this after 5pm PT today.

PR #279, opened 2022-07-17
update sdet_name when considering a different det_num

jchiang87:
Otherwise, `det_name` doesn't change when simulating multiple detectors.


rmjarvis (approved):
LGTM

PR #280, opened 2022-07-28
Don't accept folding_threshold=0 if sky_level=0.

rmjarvis:
Josh ran into a runtime error trying to do sky_level=0 with our default config.  This led to a calculation that folding_threshold=0 (since it tries to match the limiting surface brightness to the sky noise), which in turn caused stepk=0, and then a divide by zero error.

I think the right thing to do in this case is to reject that whole line of optimization and go back to the default gsparams, as we also do if the calculated folding threshold is more than the normal default.


cwwalter (approved):
LGTM.

PR #281, opened 2022-07-29
Move boresight spec to eval_variables

rmjarvis:
This shows what I meant about moving the boresight specification to eval_variables.

We had originally had the boresight specified in atm_psf, but then the wcs broke when someone chose to skip the atmosphere.  Now the situation is reversed -- if someone wants to use a simplified WCS, but keep the atmosphere, it will break.  As will the batoid optics.

Putting boresight in eval_variables is safe.  This way the 3 places that use this variable aren't mutually dependent on which other ones are kept by the user.  The user can turn off the atmospheric PSF or simplify the WCS or replace the batoid optics, each without messing up the others.

### DIFF
--- a/config/imsim-config.yaml
+++ b/config/imsim-config.yaml
@@ -21,6 +21,17 @@ eval_variables:
     # So to use this in an Eval string, you would write just run, not srun.
     srun: '0001'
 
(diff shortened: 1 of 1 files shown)


jchiang87 (approved):
LGTM.

PR #282, opened 2022-07-29
Config fixes with default nobj

rmjarvis:
The current user config doesn't work with the default nobjects.  I.e. `galsim imsim-user.yaml image.nobjects="''"` fails in several ways.

1. First, sky_model and skycat input types both assume that bandpass is already set.  This can be the case, but isn't actually required.  And letting the nobject be automatic changed the order of some things, so this wasn't true.  So I switched them to build the bandpass from image.bandpass.  This isn't inefficient, since galsim will notice if it's already been made and still "current" and just use the existing one.
2. Astropy doesn't like the default dust values I picked when we don't want any dust: Av=0, Rv=1.  Since Av=0, Rv is irrelevant, but astropy needs it to be between 2 and 6, so I switched it to the MW standard value of 3.1.
3. GSObjects made from image files (e.g. AGN) currently used some old code calling a now non-existent function find_file_path.  So I switched it to assume the file name is relative to the directory where the instcat is.  We can make this more general if we want, but I guess this is probably ok for now.


jmeyers314 (approved):
LGTM.  Thanks!

PR #283, opened 2022-08-01
get image xsize and ysize from detector object

jchiang87:


imsim/ccd.py
```diff
@@ -32,14 +32,31 @@ def setup(self, config, base, file_num, logger):
         if 'eval_variables' not in base:
             base['eval_variables'] = {}
         base['eval_variables']['sdet_name'] = det_name
+        # Get detector size in pixels.
+        det_bbox = camera[detnum].getBBox()
+        base['xsize'] = det_bbox.width
+        base['eval_variables']['ixsize'] = det_bbox.width
+        base['ysize'] = det_bbox.height
+        base['eval_variables']['iysize'] = det_bbox.height
```
rmjarvis, line 40: Another option here (not sure if it's better or not) is to add xsize and ysize to the list of eval_base_variables.
cf. http://galsim-developers.github.io/GalSim/_build/html/config_values.html#module-defined-variables

rmjarvis (approved):
LGTM.  I just have one possible design change.  It's pretty cosmetic though, so I'll let you decide which version you like better.  (Adding things to base['eval_variables'] vs telling GalSim additional base keys to allow for use in eval statements.)

imsim/ccd.py
jchiang87: Adding to the list of `eval_base_variables` seems a bit cleaner, so I've done that.

PR #284, opened 2022-08-05
Add an option to apply a specific SED in FlatBuilder

rmjarvis:
@Alex-Broughton wanted to make flats with imsim for a narrow band of wavelengths to study the wavelength dependence of the brighter-fatter effect, but that wasn't possible with the current FlatBuilder.  In fact, it doesn't even try to do anything correct about the wavelengths for a full bandpass.

So this PR adds that capability, along with an example config called flat_with_sed.yaml.  If an sed item is given in the FlatBuilder field, then it will make photons drawn with that SED and run accumulate to get the right absorption depth for each photon.  It's quite a bit slower than the regular method we use, which implicitly converts everything right at the surface.  But it does the wavelength dependence correctly.

### DIFF
--- a/tests/test_flats.py
+++ b/tests/test_flats.py
@@ -5,6 +5,7 @@ from collections import namedtuple
 import unittest
 import numpy as np
 import logging
+import sys
 import galsim
 import imsim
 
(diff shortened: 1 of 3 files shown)


tests/test_flats.py
```diff
@@ -153,6 +164,57 @@ def test_treerings_flat(self):
         assert cov01 > 0.5*tot_counts
         assert cov11 > 0.5*tot_counts
 
+    def test_sed_flat(self):
+        """Test LSST_Flat with an sed item"""
+
+        counts_per_iter = 100
+        niter = 5
+        tot_counts = counts_per_iter * niter
+        # Use an SED with a tight wavelength range in the z band
```
jchiang87, line 173: Do you mean g band here?

jchiang87 (approved):
LGTM

tests/test_flats.py
rmjarvis: Thanks.  I originally did something in z band I think, but I seem to have ended up in g band.  I think because even in z, there were a fair number of photons falling through the bottom, so here that's not an issue.

Issue #285, opened 2022-08-18
Errors should be raised if no objects are found for a CCD


jchiang87:
If there are no objects from the input catalogs that are potentially on a CCD for rendering, then error messages should be printed (or an exception raised?) to anticipate downstream failures.

For instance catalogs, this is [the relevant line of code](https://github.com/LSSTDESC/imSim/blob/3edf18c23048ec68911e88c655177b4560a52c7b/imsim/instcat.py#L212), and for skyCatalogs, it's [this line](https://github.com/LSSTDESC/imSim/blob/36b730f7bfab0eaf450c7ce6678042d2a5cb5b65/imsim/skycat.py#L53).

g-braeunlich:
Should we actually raise an `galsim.config.SkipThisObject` exception instead of a `RuntimeError`?
Or can this somehow be handled in galsim?
If yes: how could galsim figure out (or could be notified about) the number of visible objects?

PR #287, opened 2022-08-19
Raise exception when no objects are on image

g-braeunlich:
Closes: #285

### DIFF
--- a/imsim/skycat.py
+++ b/imsim/skycat.py
@@ -60,6 +60,9 @@ class SkyCatalogInterface:
         region = skyCatalogs.Box(*get_radec_limits(wcs, xsize, ysize, logger, edge_pix)[:4])
         self.objects = sky_cat.get_objects_by_region(region,
                                                      obj_type_set=obj_types)
+        if not self.objects and logger is not None:
+            logger.warning("No objects found on image")
+
 
     def getNObjects(self):
         """
@@ -103,6 +106,9 @@ class SkyCatalogInterface:
         -------
         galsim.GSObject
         """
+        if not self.objects:
+            raise RuntimeError("Trying to get an object from an empty sky catalog")
+
         skycat_obj = self.objects[index]
         gsobjs = skycat_obj.get_gsobject_components(gsparams, rng)
         seds = skycat_obj.get_observer_sed_components()
--- a/tests/test_skycat.py
+++ b/tests/test_skycat.py
@@ -3,11 +3,39 @@ import unittest
 import astropy.time
 import numpy as np
 import pandas as pd
-import yaml
 import galsim
 import imsim
 
 DATA_DIR = Path(__file__).parent / 'data'
+
+def load_test_skycat():
+    """Read in the sky catalog data used by the various tests."""
+    opsim_db_file = str(DATA_DIR / "small_opsim_9683.db")
+    visit = 449053
+    det_name = "R22_S11"  # detector 94
+
+    # Make the WCS object.
+    obs_md = imsim.OpsimMetaDict(opsim_db_file, visit=visit)
+    boresight = galsim.CelestialCoord(
+        ra=obs_md["fieldRA"] * galsim.degrees,
+        dec=obs_md["fieldDec"] * galsim.degrees,
+    )
+    rottelpos = obs_md["rotTelPos"] * galsim.degrees
+    obstime = astropy.time.Time(obs_md["mjd"], format="mjd", scale="tai")
+    band = obs_md["band"]
+    bandpass = galsim.Bandpass(
+        f"LSST_{band}.dat", wave_type="nm"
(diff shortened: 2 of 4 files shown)


rmjarvis:
I don't think this is the right place to raise an exception.  It can be completely fine to have no objects land on a particular sensor.  The problem in Virginia's case was that there were no objects in the input catalog, but then she explicitly said to draw 500 objects.  If she let `nobjects` be computed automatically (i.e. use all the input objects), it would have come out as zero and it would have happily drawn all (i.e. none) of them and moved on.
So I think we want
a) A higher verbosity message that there are no objects found (nuse==0).  Either logger.error or logger.warning.
b) An error raised in getObj if there are no objects available, but the user nonetheless requests one.

g-braeunlich:
> So I think we want a) A higher verbosity message that there are no objects found (nuse==0). Either logger.error or logger.warning. b) An error raised in getObj if there are no objects available, but the user nonetheless requests one.

This should be done now

imsim/skycat.py
```diff
@@ -52,6 +52,9 @@ def __init__(self, file_name, wcs, bandpass, obj_types=None,
         region = skyCatalogs.Box(*get_radec_limits(wcs, logger, edge_pix)[:4])
         self.objects = sky_cat.get_objects_by_region(region,
                                                      obj_type_set=obj_types)
+        if not self.objects:
+            raise RuntimeError("No skyCatalogs objects on image")
```
rmjarvis, line 56: I don't think this should be an error.  I believe this is the correlate to the place in instcat where you warn about "No objects found on image".  Thus, I think this should also be just a logged warning.

rmjarvis (changes requested):
This still needs two things:
1. Rebase to the current main to get the tests working.
2. Add unit tests that check the new functionality.  At least checking that the new error is raised in the appropriate situation.

imsim/skycat.py
g-braeunlich: OK. But what should I return if the we really hit the invalid access of item 0 in the empty list?
Or how would you prevent reaching this piece of code?

g-braeunlich: What I also thought:
Cannot we interpret `image.nobjects` as an upper limit? Or is there a usecase where a user really wants to have exact that many objects?

rmjarvis: It's the same thing as you did in instcat, isn't it?  It should only be an error if the user explicitly asks for more objects than exist.  But at this point, it should be allowed for there to be 0 objects on the image.

g-braeunlich: OK. Rebased, added a quick test and turned the above exception into a warning

rmjarvis:
Why isn't there a corresponding error raised in the skycat `getObj`?  Is that possibility already handled gracefully somehow?  I would have thought it would have the same issue.

Also, maybe this is just a philosophical difference on my part, but since this issue stemmed directly from user error, I would have liked to see a test that recapitulated what Virginia did to show that it now produces better messaging than previously.

g-braeunlich:
> Why isn't there a corresponding error raised in the skycat `getObj`? Is that possibility already handled gracefully somehow? I would have thought it would have the same issue.

Actually we also have `skycat_obj = self.objects[index]` there. Therefore I also added a check there.

> Also, maybe this is just a philosophical difference on my part, but since this issue stemmed directly from user error, I would have liked to see a test that recapitulated what Virginia did to show that it now produces better messaging than previously.

Are you thinking of an integration test? I personally would try to keep tests simple (the more test code the more maintenance)  while of course covering the relevant piece of code.

tests/test_skycat.py
```diff
@@ -69,15 +76,17 @@ def test_get_wl_params(self):
...
-            self.assertAlmostEqual(g2, gamma2/(1. - kappa))
-            self.assertAlmostEqual(mu, 1./((1. - kappa)**2 - (gamma1**2 + gamma2**2)))
+            gamma1 = row["shear_1"]
+            gamma2 = row["shear_2"]
+            kappa = row["convergence"]
+            self.assertAlmostEqual(g1, gamma1 / (1.0 - kappa))
+            self.assertAlmostEqual(g2, gamma2 / (1.0 - kappa))
+            self.assertAlmostEqual(
+                mu, 1.0 / ((1.0 - kappa) ** 2 - (gamma1**2 + gamma2**2))
+            )
```
rmjarvis, line 89: You ran black on this file I guess...?  There are a lot of content-free changes here.

rmjarvis (approved):

tests/test_skycat.py
g-braeunlich: Yes, sorry. Forgot to turn my auto formatting off. I will remove the reformatting again

g-braeunlich: Done

PR #288, opened 2022-08-22
Refactor CcdReadout so that it can be used outside of CameraReadout class

jchiang87:


rmjarvis (approved):
This seems fine to me.  I didn't try to check in excruciating detail, since I think this is really just moving functionality around, not really changing functionality.  So I just trusted that big blocks of red matched other big blocks of green in their basic structure.  :)
But those high livel structural changes seem good to me.

jchiang87:
Thanks, Mike.  This was mostly just moving things around.  However, there are a few substantive code changes that entailed writing info to the header of the e-image file in order to make it self-contained, so that the `CcdReadout` class didn't have to go back to the opsim db file for the information on each visit.    I'll make some comments where I've made those changes, in case you or anyone else wants to have a look.

imsim/ccd.py
```diff
@@ -122,6 +123,35 @@ def buildImages(self, config, base, file_num, image_num, obj_num, ignore, logger
             rng = galsim.config.GetRNG(config, base)
             cosmic_rays.paint(image.array, rng, exptime=exp_time)
 
+        # Add header keywords for various values written to the primary
```
jchiang87, line 126: This block is new functionality, i.e., writing info from the opsim meta dict object to the PHDU of the eimage file.  I think it's pretty straight-forward, but thought I should point it out anyway.

rmjarvis: The new bit just being that these values are now in the eimage header as well, rather than just the final final's headers, right?  That seems like a good change to me.

jchiang87:
Unless I hear otherwise, I'll go ahead and merge this at ~9am PT tomorrow (Oct 11).

PR #290, opened 2022-09-02
Fix test_psf

rmjarvis:
This PR fixes the error that was causing test_psf to hang recently.  It's due to a [bug in the gomp implementation of OpenMP](https://patchwork.ozlabs.org/project/gcc/patch/CAPJVwBkOF5GnrMr=4d1ehEKRGkY0tHzJzfXAYBguawu9y5wxXw@mail.gmail.com/) when used in forks.  I actually ran into it in the [GalSim unit tests too](https://github.com/GalSim-developers/GalSim/pull/1177/commits/0c3ed82f13a5e6321852c00bf30e68ee03287284), and I forgot that I would have to make the same change in imSim.

More references: [Here](http://galsim-developers.github.io/GalSim/_build/html/phase_psf.html#galsim.AtmosphericScreen) is the doc string where we officially recommend the specific pattern that I adopted in imsim.  And [here](https://galsim-developers.github.io/GalSim/_build/html/misc.html#galsim.utilities.single_threaded) is the description of single_threaded where we make the recommendation more generally.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -72,12 +72,6 @@ jobs:
               run:
                 pip install .
 
-            - name: Install main branch of GalSim
-              run: |
-                # need to make sure GalSim 2.4.0 is installed before
-                # the next imsim release:
-                pip install git+https://github.com/GalSim-developers/GalSim.git@main
-
             - name: Install test deps
               run:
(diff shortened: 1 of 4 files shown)


jchiang87 (approved):
LGTM.  Thanks, Mike!

PR #291, opened 2022-09-02
Remove the old tests that will not have any correlate in the new system.

rmjarvis:
There are only two test files remaining that we haven't converted to something new, and there isn't really anything in them that is aspirational.  They are basically just tests of the old plumbing functionality.  So this PR removes them and the bit in pyproject.toml that skips them from test runs.


jchiang87 (approved):
Thanks, Mike!

PR #293, opened 2022-09-30
improve skycat selection

jchiang87:
In addition to the region selection based on ranges of ra, dec, apply a selection in pixel coordinates.

Also, use the `.get_LSST_flux` method for skyCatalogs objects so that the cached fluxes will be used when they are available.


tests/test_multiprocessing.py
```diff
@@ -0,0 +1,72 @@
...
+    # Clean up the test output.
+    for item in glob.glob('fits/*'):
+        os.remove(item)
+    os.rmdir('fits')
+
+np.testing.assert_(run_imsim(1) is None)
+cleanup()
+
+np.testing.assert_(run_imsim(2) is None)
+cleanup()
```
rmjarvis, line 71: Is there anything in this test that is still warranted?  You made it to show me the error your were getting, which is now fixed.  I suspect it's too slow to include in our regular CI run, so probably should either remove it, or move to a devel branch.  (I guess we don't have one, but maybe devel-tools should be renamed just plain devel to be more inclusive of various scripts that have been useful during development.)
Or, if there is something here that should be kept as a CI test, then we should probably trim it down into something that runs faster and actually has tests other than just that there were no errors raised.

rmjarvis (approved):
Note sure if there was

tests/test_multiprocessing.py
jchiang87: I'm ok with just deleting this.   I'll go ahead and do that.

PR #294, opened 2022-10-04
Field rotation

g-braeunlich:


imsim/diffraction.py
```diff
@@ -42,7 +42,61 @@ class Geometry:
...
+
+    For a ray of photons with positions pos, wavelengths wavelength and pupil plane
+    coordinates pos,
+    randomly generate diffraction angles and change the directions of the photons
+    entering the pupil plane.
+
+    Parameters
+    ----------
+    pos : 2d positions of the intersections of the rays with the pupil plane (shape (n, 2)).
+    v : Direction of the photons (shape (n, 3)).
```
rmjarvis, line 65: This actually comes in with shape (3,n).  Personally, I think the code would be clearer if this *was* (n,3) as the doc string says, since right now I find the transposes required below and the way that v is indexed to be rather confusing.  But right now the doc string and the code don't match.

rmjarvis (changes requested):
I didn't fully follow all the math here.  I'm sure it's only as complicated as it needs to be, but it is rather complicated.  But the upshot is that I think having a test that actually tests the validity of the calculation somehow is even more important than usual.  

You have some nice tests of the field_rotation at various special cases where you can calculate things relatively straightforwardly.  But there is nothing that checks the final resulting spread of the diffraction spikes for something passing close to zenith.  That seems pretty important to include to me.

imsim/diffraction.py
```diff
@@ -42,7 +42,61 @@ class Geometry:
...
+        It should be a callable accepting an array for phi_star and returning the
+        random values for phi
+    """
+    R = field_rotation_matrix(lat, (telescope_orientation @ v).T, t)
+
+    def rot(v):
+        return np.einsum("kij,kj->ki", R, v)
+
+    def rot_inv(v):
+        return np.einsum("kji,kj->ki", R, v)
```
rmjarvis, line 83: One more thing. (I apparently started writing this and forgot to hit submit.)
This strikes me as possibly a tall pole for this calculation. Josh did some work investigating einsum timings and found that dot is quite a bit faster. So if this is possible to reorder into something that works with dot, that would probably be worth doing.
cf. https://github.com/jmeyers314/SeriesNotebooks/blob/master/np.einsum%20vs%20np.dot%20speed.ipynb

imsim/diffraction.py
g-braeunlich: Sorry, actually we still need to feed `v` directly to `batoid.RayVector`. Therefore, for performance reasons, I would recommend to go with shape (3,n)

imsim/diffraction.py
g-braeunlich: Yep, `(R @ pos[..., None]).squeeze(axis=-1)` would be possible (or `:, :` instead of `...`)

g-braeunlich: ... but:
```py
import numpy as np
n = 10000
A = np.random.rand(n, 2, 2)
x = np.random.rand(n, 2)
%timeit y = (A @ x[:, :, None]).squeeze(axis=-1)
# 371 µs ± 12.5 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
%timeit y = np.einsum("kij,kj->ki", A, x)
# 105 µs ± 2.31 µs per loop (mean ± std. dev. of 7 runs, 10,000 loops each)

File: tests/test_photon_ops.py [hunk 3]
```diff
@@ -60,3 +63,74 @@ def test_lsst_optics() -> None:
...
+    sky_pos = galsim.CelestialCoord(0.543 * galsim.radians, -0.174 * galsim.radians)
+    xy_to_v = photon_ops.XyToV(local_wcs, icrf_to_field, sky_pos)
+
+    x, y = np.array(
+        np.meshgrid(np.linspace(-10.0, 10, 20), np.linspace(-10.0, 10, 20))
+    ).reshape((2, -1))
+    v = xy_to_v(x, y)
+    x_after, y_after = xy_to_v.inverse(v)
+    np.testing.assert_array_almost_equal(x, x_after)
+    np.testing.assert_array_almost_equal(y, y_after)
```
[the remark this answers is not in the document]
g-braeunlich: Indeed, the math is pretty involved. To be honest it is not transparent at all, what is happening behind all the WCS's involved here. It would take me quite a while to fully understand the transforms.
Of course, any help is welcome here, but I would not be able to provide a good test until the end of the week.

At least, I tried mapping `(x,y)=(0,0)` to `v`, but the result `v=(3.40953136e-13, -6.53235082e-04, -9.99999787e-01)` also does not look like something predictable.

imsim/diffraction.py
rmjarvis: You can make the array column-major.  Use `order='F'` when making the array.  Then it can be (n,3) but still have contiguous columns.

tests/test_photon_ops.py
```diff
@@ -60,3 +63,74 @@ def test_lsst_optics() -> None:
...
+    )
+    local_wcs = galsim.AffineTransform(
+        0.168,
+        0.108,
+        -0.108,
+        0.168,
+        origin=galsim.PositionD(x=-0.349, y=-0.352),
+        world_origin=galsim.PositionD(x=0.0, y=0.0),
+    )
+    lsst_diffraction.applyTo(photon_array, local_wcs=local_wcs)
```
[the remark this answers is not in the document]
g-braeunlich: I added some tests. And this was indeed necessary! Thanks for this request!

imsim/diffraction.py
g-braeunlich: I used `np.array(...).T` this also leads to a column major array.

imsim/diffraction.py
jmeyers314: > Josh did some work investigating einsum timings and found that dot is quite a bit faster

Wow, that's a deep cut.  I completely forgot I did that :)  I wonder if it holds up after 7 years...

rmjarvis: lol.  You were among the first few hits on whatever google terms I searched for trying to see how einsum might compare to other options.  :)

imsim/photon_ops.py
```diff
@@ -135,12 +119,248 @@ def applyTo(self, photon_array, local_wcs=None, rng=None):
...
+            for optics_key, shift in shift_optics.items():
+                telescope = telescope.withGloballyShiftedOptic(optics_key, shift)
+        self.telescope = telescope
+        self.latitude = latitude
+        self.altitude = altitude
+        self.azimuth = azimuth
+        self.sky_pos = sky_pos
+        self.icrf_to_field = icrf_to_field
+
+        deviate = GaussianDeviate(seed=seed)
```
rmjarvis, line 185: This isn't right.  We need to be using the config rng, not making a new one with some random seed (especially not default seed=None).  Basically, the whole diffraction_rng stuff needs to be in `applyTo` and use the `rng` paramter there.  And remove the `seed` parameter here.

imsim/photon_ops.py
```diff
@@ -135,12 +119,248 @@ def applyTo(self, photon_array, local_wcs=None, rng=None):
...
+        return np.array(batoid.utils.gnomonicToDirCos(thx, thy)).T
+
+    def inverse(self, v: np.ndarray) -> tuple:
+        thx, thy = batoid.utils.dirCosToGnomonic(v[:, 0], v[:, 1], v[:, 2])
+        # field to ICRF
+        ra, dec = self.icrf_to_field.xyToradec(thx, thy, units="rad")
+        # ICRF to u/v
+        u_rad, v_rad = self.sky_pos.project_rad(ra, dec)
+        u = 3600 * np.rad2deg(u_rad)
+        v = 3600 * np.rad2deg(v_rad)
```
rmjarvis, line 356: You're reusing the name `v` here.  That's not a good idea.  Can we come up a different name for the 3d velocity vector?  u,v are used all over the place to mean the tangent plane position, so I think it's better to leave this v alone and change the other one.  Maybe `dir` for direction, since I believe it's a unit vector representing the direction the photon is moving, right?

imsim/photon_ops.py
```diff
@@ -135,12 +119,248 @@ def applyTo(self, photon_array, local_wcs=None, rng=None):
...
+        """
+
+        xy_to_v = XyToV(local_wcs, self.icrf_to_field, self.sky_pos)
+        # Convert xy coordinates to a cartesian 3d velocity vector of the photons
+        v = xy_to_v(photon_array.x, photon_array.y)
+
+        # Adjust for refractive index of air
+        wavelength = photon_array.wavelength * 1e-9
+        n = self.telescope.inMedium.getN(wavelength)
+        v /= n[:, None]
```
rmjarvis, line 216: Are you sure this line is right?  Why does the refractive index of air affect the diffraction angle?

imsim/photon_ops.py
```diff
@@ -135,12 +119,248 @@ def applyTo(self, photon_array, local_wcs=None, rng=None):
...
+    return decorator
+
+
+def config_kwargs(config, base, cls):
+    """Given config and base, extract parameters (including optionals)
+    and optional parameters.
+    """
+    req, opt, single, _takes_rng = get_cls_params(cls)
+    kwargs, _safe = GetAllParams(config, base, req, opt, single)
+    opt_kwargs = specified_opt_kwargs(kwargs, cls._opt_params.keys())
```
rmjarvis, line 278: I think there must not be any test that runs this function, since I'm pretty sure there is no chance this could work.  `GetAllParams` cannot handle something with type `batoid.Optic`.  That should not be in the `_req` dict.  It needs to be created separately.  Presumably with the `load_telescope` function you made in batoid_utils.

rmjarvis: I also don't understand the point of the `specified_opt_kwargs` function.  Seems like what you are doing with that could just be done with the regular `kwargs` rather than doing this bit to separate out the optional kwargs.  Anyway, I suspect you mostly just need a test that runs this function so you can get this working correctly.

tests/test_photon_ops.py
```diff
@@ -26,31 +22,210 @@ def test_lsst_optics() -> None:
...
+    # Check that the angle of the crosses are rotated relative to each other:
+    expected_angle_difference = dt * (
+        diffraction.OMEGA_EARTH
+        * np.cos(lsst_diffraction.latitude / 180.0 * np.pi)
+        * np.cos(lsst_diffraction.azimuth / 180.0 * np.pi)
+        / np.cos(lsst_diffraction.altitude / 180.0 * np.pi)
+    )
+
+    np.testing.assert_almost_equal(
+        cross_rot_angle_1 - cross_rot_angle_0, expected_angle_difference, decimal=2
```
rmjarvis, line 221: decimal=2 kind of hides the level of precision here.  Should use `assert_allclose` and `rtol` to be clearer about the level of agreement.  I think this isn't currently very close.  I'm getting 0.018 vs 0.026, which is pretty different.  Do you understand where the discrepancy is coming from?

imsim/photon_ops.py
g-braeunlich: `dir` is a builtin function. would `v_photon` or `v_phot` be ok for you?

imsim/photon_ops.py
g-braeunlich: This is coming from @jmeyers314 . Do you have more info about this?

imsim/photon_ops.py
g-braeunlich: After the split of the photon ops, we actually dont need it anymore. But before, we had `use_diffraction` with a default value of `True`. But when we did not specify this value in the config, the parameter was filled with a `None` which evaluated to `False` overriding the default of `True`.
Therefore I introduced this, to protect myself from that behavior (I just removed all values with `None` from `kwargs` assuming they were not specified in the config).
This is btw also something, I would put on my wish list for GalSim. Like this, it is a bit of a pitfall.

imsim/photon_ops.py
jmeyers314: It's photon velocity in units of `c`, so the refractive index does need to be accounted for.

tests/test_photon_ops.py
g-braeunlich: I first thought, that the field rotation rate is changing significantly this close to zenith. But I now had a look and after 1sec,
we only have a change in the rate of `-3.600097e-05`. This cannot explain the difference. Have to investigate.

g-braeunlich: According to `field_rotation_matrix`, I would expect a value of `0.02618501` rad at `t=1` and this position. At least so far this seems to agree.
The only place in the code this angle kicks in is at
```py
    shift = rot(
        diffraction_delta(rot_inv(pos), -v[:, 2], wavelength, geometry, distribution)
    )
```
So question (@rmjarvis / @jmeyers314 ):
is one of `rot`, `rot_inv` actually not needed? Is maybe the rotation of `v` already taken care of in the photon array?

imsim/photon_ops.py
rmjarvis: I don't think that's right. If it's not in the config, it won't be in kwargs at all. There's nothing in GalSim, I don't think, that inserts a default value of None.  We have other places with default value that aren't None which work fine.

imsim/photon_ops.py
rmjarvis: Anything to disambiguate is fine with me.

tests/test_photon_ops.py
g-braeunlich: I think, I just messed up the order of arguments in `star_trace`.
Now:
```
ACTUAL: 0.022242493576406397
DESIRED: 0.025522087833554367
```

imsim/photon_ops.py
g-braeunlich: But there is something which injects a `None` at this place.
You can verify yourself:
Go back to a84a7fa3031dc5ad204858e8e03619f4a202de5a (thats one commit before ad2bd07008bb7dfa14e52680eb2cdfd868d84ab7, which introduced my "hack"). There we also still have `use_diffraction`.

If I then run imsim with `galsim 1-star-near-zenith.yaml`, I end up with `use_diffraction=None`.

`1-star-near-zenith.yaml`:

```yml
---
modules: [imsim]
template: imsim-config

input.instance_catalog.file_name: 1-star-near-zenith.txt
input.opsim_meta_dict.file_name: 1-star-near-zenith.txt
input.tree_rings.only_dets: [R22_S11]

image.sky_level: 0.000001

gal:  # Override object in cat with this
  type: DeltaFunction
  sed:
    file_name: vega.txt
    wave_type: nm
    flux_type: flambda
    norm_flux_density: 1e5
    norm_wavelength: 500

output.dir: output/1-star-near-zenith
output.readout: ""
output.det_num:
  type: List
  items: [94]
output.nfiles: 1
output.truth: ""
```

`1-star-near-zenith.txt`:

```
#comment
rightascension 63.3
declination -30.14463
mjd 59797.2854090
altitude 89.9
azimuth 73.7707957
filter 2
rotskypos 69.0922930
rottelpos 1.000
dist2moon 145.1095257
moonalt -11.1383568
moondec -18.7702120
moonphase 59.6288830
moonra 230.9832941
nsnap 2
obshistid 161899
seed 161899
seeing 0.7613760
sunalt -59.1098785
vistime 33.0000000
object MS_567_8a 63.35000000000000 -30.1 14.0 starSED/phoSimMLT/lte033-4.5-1.0a+0.4.BT-Settl.spec.gz 0 0 0 0 0 0 point none CCM 0.0635117705 3.1
```

rmjarvis: I tried to verify by adding a print statement after `GetAllParams`, and kwargs is the following:
```
kwargs =  {'band': 'r', 'boresight': coord.CelestialCoord(coord.Angle(1.1047934165124105, coord.radians), coord.Angle(-0.5261230452954583, coord.radians)), 'telescope': <batoid.optic.Optic object at 0x177e849a0>}
```
There is no `use_diffraction: None` in there.  Did you intend me to do something different to verify?  Maybe I'm not understanding the problem you had.

For reference:
```
(lsst-scipipe-4.0.1) Mike@Saxophone:~/LSSTDESC/imSim[(HEAD detached at a84a7fa)*] $ git status
HEAD detached at a84a7fa
Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
	modified:   imsim/photon_ops.py

no changes added to commit (use "git add" and/or "git commit -a")
(lsst-scipipe-4.0.1) Mike@Saxophone:~/LSSTDESC/imSim[(HEAD detached at a84a7fa)*] $ git diff
diff --git a/imsim/photon_ops.py b/imsim/photon_ops.py
index a5fdb07..d253ab2 100644
--- a/imsim/photon_ops.py
+++ b/imsim/photon_ops.py
@@ -165,7 +165,10 @@ class LsstOpticsFactory(PhotonOpBuilder):
 
     def buildPhotonOp(self, config, base, _logger):
         req, opt, single, _takes_rng = get_cls_params(LsstOptics)
         kwargs, _safe = GetAllParams(config, base, req, opt, single)
+        print('kwargs = ',kwargs)
 
         opt_kwargs = {key: kwargs.get(key) for key in LsstOptics._opt_params.keys()}
         shift_optics = opt_kwargs.pop("shift_optics")
```

g-braeunlich: I have to apologize! I only looked in the constructor of `LsstOptics` and found `use_diffraction` there.
I did not even have a look at the line (visible in your diff)
```py
opt_kwargs = {key: kwargs.get(key) for key in LsstOptics._opt_params.keys()}
```
Which actually adds the `None`.
So big sorry about the wrong accusation, this was introduced by me.

g-braeunlich: OK. So the `specified_opt_kwargs` is gone now.
Currently, I also am not retrieving `telescope` from `kwargs`, but from `base`.
`telescope=base["_telescope"]`. Like this things seem to work (at least when running a simulation from `galsim`).

But I agree, that adding `"telescope"` to `_req_params` is probably not the clean way.
Is using `telescope=base["_telescope"]` a proper way? Or should the telescope be really built via `load_telescope`?

Regarding tests: Do you have functionality to populate `base` and `config` (e.g. according to a yaml config file represented by a `dict`)? Or has that to be done manually?

rmjarvis: > Is using telescope=base["_telescope"] a proper way? Or should the telescope be really built via load_telescope?

I believe `base['_telescope']` is made by the BatoidWCS.  So if we use that, then we are implicitly requiring that we always use that WCS whenever we want to use these photon ops.  Maybe that's ok.  But probably better practice would be to check for the cached `_telescope`, and if it's not present, then make it (and cache it).

rmjarvis: > Do you have functionality to populate base and config (e.g. according to a yaml config file represented by a dict)? Or has that to be done manually?

There are tests in some other imsim test files that test config dicts.  See test_flats, test_psf, and test_batoid_wcs for instance.  But you can also use `galsim.config.ReadConfig` to read in a config dict from a file if that's easier.  The GalSim tests of the config processing use both approaches.

g-braeunlich: ok. I added some tests and refactored `load_telescope`

imsim/photon_ops.py
```diff
@@ -275,7 +241,9 @@ def config_kwargs(config, base, cls):
     """
     req, opt, single, _takes_rng = get_cls_params(cls)
     kwargs, _safe = GetAllParams(config, base, req, opt, single)
-    opt_kwargs = specified_opt_kwargs(kwargs, cls._opt_params.keys())
+    opt_kwargs = {
+        key: kwargs.get(key) for key in cls._opt_params.keys() if key in kwargs
+    }
     return kwargs, opt_kwargs
```
rmjarvis, line 247: I think this would be cleaner to only return `kwargs`, not separate out `opt_kwargs` explicitly.
Then any kwargs that need special handling, you can pop off (e.g. `telescope` and `band`).  The ones that are used as is (e.g. boresight and any optional ones) get naturally applied by ending the call to `LsstOptics` with `**kwargs`.

imsim/batoid_utils.py
```diff
@@ -1,20 +1,53 @@
...
+            type: lsst_optics
+            ...
+            shift_optics:
+              Detector: [0, 0, 1.5e-3]
+              M2: [3.0e-3, 0, 0]
+    """
+    telescope = load_telescope(telescope_id, band)
+    if shift_optics is not None:
+        for optics_key, shift in shift_optics.items():
+            telescope = telescope.withGloballyShiftedOptic(optics_key, shift)
```
rmjarvis, line 52: @jmeyers314 Is this a trivial operation?  Or should we be caching this function too?

tests/test_photon_ops.py
```diff
@@ -217,8 +221,8 @@ def test_lsst_diffraction_shows_field_rotation() -> None:
         / np.cos(lsst_diffraction.altitude / 180.0 * np.pi)
     )
 
-    np.testing.assert_almost_equal(
-        cross_rot_angle_1 - cross_rot_angle_0, expected_angle_difference, decimal=2
+    np.testing.assert_allclose(
+        cross_rot_angle_1 - cross_rot_angle_0, expected_angle_difference, rtol=0.13
```
rmjarvis, line 225: I think this is probably fine, although 0.13 is still a pretty large tolerance.  But probably worth some developer commentary on why you chose this value.  I.e. give the numerical results and speculate why it's not closer than this.

One possibility is that it's due to the rotation angle not being quite constant even over this short time when it's so close to zenith.  Maybe worth trying another position that's a little farther from zenith, so that the rotation is still easily measurable, but maybe would be more uniform and thus agree with the prediction better.

tests/test_photon_ops.py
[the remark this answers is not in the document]
rmjarvis: I was thinking of some special cases like when the sky_pos is close to boresight, then the projection stuff is nearly trivial.  So (x,y) near (0,0) should be straightforward.  Even moreso if the wcs is just a PixelScale, but maybe even simple in the general local_wcs case.

imsim/photon_ops.py
```diff
@@ -285,13 +253,14 @@ def deserialize_lsst_optics(config, base, _logger):
...
 
     return LsstOptics(
-        telescope=base["_telescope"],
+        telescope=load_telescope_with_shift_optics(
+            kwargs["telescope"], kwargs["band"], shift_optics=shift_optics
+        ),
         boresight=kwargs["boresight"],
         sky_pos=base["sky_pos"],
         image_pos=base["image_pos"],
         icrf_to_field=base["_icrf_to_field"],
```
rmjarvis, line 262: This is also implicitly requiring the BatoidWCS.  Can we do the same trick for that as you did for the telescope?

imsim/batoid_utils.py
jmeyers314: It's pretty quick.  About 25 microseconds for LSST optics on my laptop.

imsim/photon_ops.py
jmeyers314: In my not-yet committed edits for adding bending mode perturbations, I rewrote how the batoid telescope instance was handled (there it's loaded by the config file "input" section processing).  I expect that will eventually supersede this too.

imsim/photon_ops.py
g-braeunlich: Done

imsim/photon_ops.py
g-braeunlich: @rmjarvis  This is not as easy as in the case of the telescope.
Right now, assembling `icrf_to_field` involves `BatoidWCSBuilder.makeWCSFactory`, which requires all the args necessary for `BatoidWCS`. Therefore refactoring this would require more time and I am not sure, how many of the args of `BatoidWCS` would be needed in the end.
But maybe the edits of @jmeyers314 will improve this a bit.

rmjarvis: Right.  Let's leave this as is for now, but hopefully we can make this a little more flexible with Josh's upcoming changes.

rmjarvis (approved):

rmjarvis:
LGTM.  @jmeyers314 I think you also wanted to review this?

jmeyers314:
Yeah, I'd like to review the math, or at least review the unit tests to check on the math that way.  I'll try to get on that this weekend.

g-braeunlich:
@jmeyers314 Regarding the math:
I wrote down my approach in [doc/validation/diffraction.rst](doc/validation/diffraction.rst).
Note: although docs are already published now: https://lsstdesc.org/imSim/
diffraction is not yet published.
@rmjarvis Before merging this, I would like to address your 2 remaining comments (investigate a bit more the discrepancy in the field rotation angle in the test and maybe write a test for this chained transform). I would try this on Monday.

tests/test_photon_ops.py
g-braeunlich: So, the fact that the rotation angular rate is not constant over the time interval actually plays a role. However, taking this into account increases the discrepancy.
The actual reason just was the ridiculous small amount of spike photons (36 only).
I increased the amount of photons by a factor of 100 now and this seems to be much more accurate.
I now directly compared against rotation angle instead of the rate.

In exchange, I improved the test relating the angle to its rate, by using numerical integration (test_diffraction.py)

tests/test_photon_ops.py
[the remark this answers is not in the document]
g-braeunlich: I just noticed, that in the current example `sky_pos == boresight`.
But it is still not transparent to me what exactly is happening behind each of the steps involved in this transform

jmeyers314:
@g-braeunlich , I think I'm generally happy now that you've reduced the tolerance for the rotation angle test.  However, the test itself now fails for me on my M1 mac:

```
>       np.testing.assert_allclose(
            cross_rot_angle_1 - cross_rot_angle_0, expected_angle_difference, rtol=1.0e-3
        )
E       AssertionError: 
E       Not equal to tolerance rtol=0.001, atol=0
E       
E       Mismatched elements: 1 / 1 (100%)
E       Max absolute difference: 0.00070978
E       Max relative difference: 0.02710643
E        x: array(0.026895)
E        y: array(0.026185)

tests/test_photon_ops.py:223: AssertionError
```

Checking linux now to see how that goes.  Did GHA not trigger for the latest commits?

jmeyers314:
Slightly different (but still failing) result on USDF at SLAC:

```
>       np.testing.assert_allclose(
            cross_rot_angle_1 - cross_rot_angle_0, expected_angle_difference, rtol=1.0e-3
        )
E       AssertionError: 
E       Not equal to tolerance rtol=0.001, atol=0
E       
E       Mismatched elements: 1 / 1 (100%)
E       Max absolute difference: 0.00067827
E       Max relative difference: 0.02590317
E        x: array(0.026863)
E        y: array(0.026185)

tests/test_photon_ops.py:223: AssertionError
```

@rmjarvis , I forget, do we expect these to be the same?  (Same RNGs across platforms?)

rmjarvis:
> @rmjarvis , I forget, do we expect these to be the same? (Same RNGs across platforms?)

The GalSim-generated random numbers should be identical across platforms.  But numpy doesn't guarantee that, and it looks like there are numpy randoms in the test as well.  So I can't tell offhand whether we should expect this particular test to be stable across platforms.

g-braeunlich:
Sorry, had to adapt the tolerance.
I also rebased on main now and took the opportunity to squash all fixups.
The only thing where I had to deal with conflicts and had to make changes relative to my last commit was the part about the new argument `camera`.
@jmeyers314 Am I correct, that the purpose of adding `camera` as a property to the factory is actually caching the camera?
If yes, my current version of [imsim/photon_ops.py](imsim/photon_ops.py), `deserialize_lsst_optics` should be fine already.

jmeyers314:
Okay, see #303 which fixes the test_object_positions bug I see locally.  That said though, I'm really confused here how CI passed in GHA.  Is it up-to-date with the latest commits?

PR #295, opened 2022-10-05
imSim 2.0 directory and file cleanup

cwwalter:
Remove the old imSim code and cleanup and rename flies and directories in preparation for the imSim 2.0 release.

First pass through files for people to look at.


jchiang87 (changes requested):
A couple comments/requests:
1. We need to restore `etc/.condarc` and possibly move it elsewhere, and if so, edit `.github/workflows/ci.yaml` accordingly.  I'd just leave it as-is.
2. Of all of the apps now in `bin`, I'd say that only `cosmic_ray_extractor.py` has code worth saving.  It will still need some re-writing, of course, since it relies on the old `python/desc/imsim` code.   The remaining ones can be discarded.

Other than these two items, I'm ok with the rest of the changes.

jchiang87 (approved):
lgtm

PR #298, opened 2022-10-17
ci: fix pages workflow

g-braeunlich:
Another try. This time, I was following https://docs.github.com/en/pages/getting-started-with-github-pages/configuring-a-publishing-source-for-your-github-pages-site#creating-a-custom-github-actions-workflow-to-publish-your-site

and I now got:
```
Branch "doc" is not allowed to deploy to github-pages due to environment protection rules.
```
Already a good sign, but it also means that we have to merge to main in order to try this out (or change our protection rules, which I would not recommend).

### DIFF
--- a/.github/workflows/gh-pages.yml
+++ b/.github/workflows/gh-pages.yml
@@ -25,18 +25,27 @@ jobs:
       - name: "Bypass jekyll on github pages"
         run: "touch public/.nojekyll"
       - name: Upload artifacts
-        uses: actions/upload-artifact@v3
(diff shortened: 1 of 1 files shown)


g-braeunlich:
Had to mark it a draft again. Publish fails even though I did not trigger manually. Have to investigate

g-braeunlich:
Fixed

g-braeunlich:
I now tested it on a fork. Seems to work: https://g-braeunlich.github.io/imSim/

PR #300, opened 2022-10-31
Object position tests

jchiang87:


imsim/photon_ops.py
```diff
@@ -43,7 +46,12 @@ class LsstOptics(PhotonOp):
         Enable / disable diffraction (default: enabled)
     """
 
-    _req_params = {"telescope": Optic, "band": str, "boresight": CelestialCoord}
+    _req_params = {
+        "telescope": Optic,
+        "band": str,
+        "boresight": CelestialCoord,
+        "camera": None
```
rmjarvis, line 53: This should be `str` I think.  This is supposed to be whatever type is specified in the config file.  None isn't a type.  But in the imsim-config.yaml edit, you made it a string (copying the string in output.camera), so this should be specified as a string here.
Also, FYI, Geri updated how the telescope is specified in his branch (for the same reason -- an Optic cannot really be given in a yaml file).  That's now a str too.

tests/test_object_positions.py
```diff
@@ -0,0 +1,79 @@
...
+    return np.sqrt((x_avg - data['x'])**2 + (y_avg - data['y'])**2)
+
+for camera in ('LsstCam', 'LsstCamImSim'):
+    print(camera)
+    run_imsim(camera)
+    eimage_files = glob.glob(f'fits_{camera}/eimage*')
+    for eimage_file in eimage_files:
+        offset = compute_pixel_offset(eimage_file)
+        print(eimage_file, offset)
+        np.testing.assert_(offset < 2)
```
rmjarvis, line 79: This test script isn't currently being run by CI.  If we want it to be, then it needs a `test_*` function (or other UnitTest class implementation).

rmjarvis (changes requested):
Not sure if this is ready or not.  Has the underlying position issue been resolved?
Anyway, my comments are orthogonal to that.

jchiang87:
@rmjarvis Thanks for the comments.  I think this ready for merging.

rmjarvis (approved):

PR #301, opened 2022-11-07
Add sky background gradient and vignetting

jchiang87:


jchiang87:
I modeled the vignetting using a spline fit to the radial profile of measured fluxes from stars with the same input magnitude randomly distributed across the focal plane.  The figure on the left below shows the rendered locations on the focal plane, and the figure on the right is the radial profile of the measured fluxes and the spline fit.
LSSTCam_vignetting_profile_fit

The image below shows a mosaic of the sky background for DC2 visit 182850 with the vignetting and sky gradient functions applied.
fp_vignetting_sky_gradient

jchiang87:
To confirm that these changes are compatible with the WCS distortions, e.g., from tree rings, I've simulated a scene with CCD R41_S10, which is one of the outermost CCDs on the focal plane and has particularly strong tree rings:
eimage_00182850-1-i-R41_S10-det165

I think this PR is ready for review.

devel-tools/vignetting/sample_focal_plane.py
```diff
@@ -0,0 +1,55 @@
...
+from lsst.afw import cameraGeom
+import lsst.geom
+import imsim
+
+# Pointing info for DC2 visit 182850:
+ra = 51.99085849573259566
+dec = -40.31737846575015283
+rottelpos = 28.8262515
+mjd = 59822.28563761110854102
+band = 'i'
```
rmjarvis, line 17: Does the vignetting change much for different bands?  Are we ok with just using i band here?

devel-tools/vignetting/sample_focal_plane.py
```diff
@@ -0,0 +1,55 @@
...
+#df0.to_parquet('fp_sample.parq')
+
+template = "object %(object_id)i %(ra).10f %(dec).10f 20 starSED/phoSimMLT/lte027-2.0-0.0a+0.0.BT-Settl.spec.gz 0 0 0 0 0 0 point none CCM 0.04507462 3.1\n"
+
+instcat_file = 'stars.txt'
+with open(instcat_file, 'w') as fobj:
+    for object_id, (_, row) in enumerate(df0.iterrows()):
+        ra = row.ra
+        dec = row.dec
+        fobj.write(template % locals())
```
rmjarvis, line 55: I think the files in this directory are missing the step where the image is actually drawn.  You should include whatever config file you used to do that here along with a comment describing the full process for someone to reproduce the json file you made if necessary.

imsim/ccd.py
```diff
@@ -34,6 +34,7 @@ def setup(self, config, base, file_num, logger):
         # Figure out the detector name for the file name.
         detnum = galsim.config.ParseValue(config, 'det_num', base, int)[0]
         camera = get_camera(config['camera'])
+        base['camera'] = config['camera']
```
rmjarvis, line 37: I don't think this is how we want to do this.  We have several places in the config that need the camera, and we always use `camera: "@output.camera"` to refer to the one in `output`.  We should do that here too where the image type has its own `camera` item.  Then if the user wants to replace any of these with something else (e.g. not use our output type, but keep the LSST_Image image type) there is a place to specify the `camera` item properly.

imsim/lsst_image.py
```diff
@@ -0,0 +1,169 @@
...
+                'apply_vignetting': bool, 'apply_sky_gradient': bool,
+                'vignetting_data_file': str}
+        params = GetAllParams(config, base, opt=opt, ignore=ignore+extra_ignore)[0]
+
+        size = params.get('size',0)
+        full_xsize = params.get('xsize',size)
+        full_ysize = params.get('ysize',size)
+
+        self.apply_vignetting = params.get('apply_vignetting', False)
+        vignetting_data_file =  params.get('vignetting_data_file', False)
```
rmjarvis, line 116: False is the wrong default here.  None maybe?
Or better probably, only read it if apply_vignetting is True, in which case, make it required with no default.

imsim/lsst_image.py
```diff
@@ -0,0 +1,169 @@
...
+            image:          The image onto which to add the noise.
+            config:         The configuration dict for the image field.
+            base:           The base configuration dict.
+            image_num:      The current image number.
+            obj_num:        The first object number in the image.
+            current_var:    The current noise variance in each postage stamps.
+            logger:         If given, a logger object to log progress.
+        """
+        base['current_noise_image'] = base['current_image']
+        sky = GetSky(config, base, full=True)
```
rmjarvis, line 148: I'm not sure that this necessarily returns an image (which is assumed below).  If the wcs is uniform, then the sky level is constant, and I think this will just return that constant as a float, not an image full of this value, even with full=True.  (Probably bad API design, TBH -- I suspect it would have been better to let this parameter mean always return a full image.)  
Anyway, I think we need a check below if doing the vignetting or gradient to turn the sky into an image if it isn't one already.

imsim/lsst_image.py
```diff
@@ -0,0 +1,169 @@
...
+
+        if sky:
+            if self.apply_sky_gradient:
+                ny, nx = sky.array.shape
+                sky_model = galsim.config.GetInputObj('sky_model', config, base,
+                                                      'LSST_ImageBuilder')
+                sky_gradient = SkyGradient(sky_model, config['wcs']['current'][0],
```
rmjarvis, line 155: Should use image.wcs, not current['wcs']['current'][0].

rmjarvis (changes requested):

imsim/lsst_image.py
rmjarvis: Update -- I looked into this more, and there would be other inefficiencies if this always returned an image.  The API is such that with full=True, then the return value can be added to an existing image to add the right sky value everywhere.  I.e. `image += sky`.  So if it's really constant, there is no point in the extra array allocation.  If full=False, it just returns the value at the center, which in other contexts is sufficient (e.g. when calculating an estimate of the overall noise variance in an image).

Anyway, the upshot is that we should add the following at this point:
```
if not isinstance(sky, galsim.Image):
    sky = galsim.Image(bounds=image.bounds, wcs=image.wcs, init_value=sky)
```

devel-tools/vignetting/sample_focal_plane.py
jchiang87: I ran the focal plane sampling simulation for `u, g, i, y` bands.  There is up to a ~few percent difference in vignetting response comparing u-band to y-band around 330 mm from the focal plane center:
vignetting_band_dependence
The top plot shows the spline fits for the `ugiy` sims with the measured fluxes for g-band plotted.  It's zoomed in on the region of the greatest disparity between spline fits. The bottom plot shows the ratio of the spline fits in `ugi` to `y` band.  Note that each spline function is normalized to unity at the center of the focal plane.

It would be straight-forward to have vignetting functions for each band, if we all agree that is worth doing.

devel-tools/vignetting/sample_focal_plane.py
jchiang87: added the config file and README with instructions.

devel-tools/vignetting/imsim-vignetting.yaml
```diff
@@ -0,0 +1,33 @@
+# Use baseline configuration
+modules: [imsim]
+template: imsim-config
```
rmjarvis, line 3: I think let's copy the config file and edit it, rather than use a template.  I'm concerned that we'll make various changes to the base imsim-config over time, and this will become broken.

devel-tools/vignetting/sample_focal_plane.py
rmjarvis: I think for now, we should change this line to get the band from the instance catalog (filter=3), as that is how the README says to change the filter.  We should make sure `band` here always matches.

jchiang87: In that case, we should just get all of the observation info from the main instance catalog file.

rmjarvis: Sure.  Do we even need to use the instance catalog for this?  We could just specify everything in the config file directly.

jchiang87: I actually did try that, but I couldn't get it to work properly--the objects didn't sample the CCDs uniformly.  If you can provide a config file that does that, I'm happy to use that instead.

jmeyers314: Huh.  That's more variation than I expected, though I admit I've never actually tried to check before.  The filters are slightly powered optics, so it's certainly possible for there to be some small effect on the vignetting curve.

jchiang87:
@rmjarvis Assuming we will just proceed with the radial profile of the vignetting for i-band that I derived, I think I've addressed all of your review comments at this point.  Is there anything else we need to do for this PR?

rmjarvis (approved):

rmjarvis:
I'm happy with this.  I'm still trying to understand what is going wrong with the ConnectionRefusedError in the proxy stuff, but that seems to be unrelated to all of this.

PR #305, opened 2022-11-14
Automatic imsim-env docker image publication CI

stuartmcalpine:
Includes a Docker file and CI workflow to automatically publish a "latest" imsim-env image to the lsstdesc DockerHub repo.

The image is built with the latest version of the LSST Science Pipelines stack, imSim, and imSim's dependencies.

In the future/next step, I think we should also have a docker image associated with each release of imSim, to accompany the latest version.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,7 +9,7 @@ RUN git clone https://github.com/LSSTDESC/imSim.git &&\
 RUN source /opt/lsst/software/stack/loadLSST.bash &&\
     setup lsst_distrib &&\
     pip install --upgrade galsim &&\
-    pip install dust_extinction palpy batoid &&\
+    pip install dust_extinction palpy batoid gitpython &&\
     pip install git+https://github.com/LSSTDESC/skyCatalogs.git@master &&\
     cd imSim && pip install -e . && cd .. &&\
     cd rubin_sim && pip install -e .
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,12 @@
 # Start from the latest release LSST stack image.
-from lsstsqre/centos:7-stack-lsst_distrib-w_latest
+FROM lsstsqre/centos:7-stack-lsst_distrib-w_latest
+
+# Information about image.
+ARG BUILD_DATE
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -0,0 +1,26 @@
...
+    pip install dust_extinction palpy batoid gitpython &&\
+    pip install git+https://github.com/LSSTDESC/skyCatalogs.git@master &&\
+    cd imSim && pip install -e . && cd .. &&\
+    cd rubin_sim && pip install -e .
+
+# Download Rubin Sim data.
+RUN mkdir rubin_sim_data &&\
+    curl https://s3df.slac.stanford.edu/groups/rubin/static/sim-data/rubin_sim_data/skybrightness_may_2021.tgz | tar -C rubin_sim_data -xz &&\
+    curl https://s3df.slac.stanford.edu/groups/rubin/static/sim-data/rubin_sim_data/throughputs_aug_2021.tgz | tar -C rubin_sim_data -xz
+
```
jchiang87, line 21: We can also download the SED data using
```
curl https://s3df.slac.stanford.edu/groups/rubin/static/sim-data/sed_library/seds_170124.tar.gz  | tar -C rubin_sim_data/sims_sed_library -xz
```
and then set the env var below appropriately.

jchiang87 (changes requested):
Other than the suggestion to download the SED library rather than relying on the NERSC location, this looks good to me.

heather999:
Is it a conscious choice to manually download the rubin_sim data rather than use their script, `rs_download_data`? https://github.com/lsst/rubin_sim#install-into-an-lsst-stack-environment

jchiang87:
When I tried to use that script, it seemed to require software beyond the minimal needed to run the imSim code.  For the GHA CI, it seemed advisable to limit the third party dependencies.   If we needed more than just those few datasets, then it might make sense to use their script.

heather999:
I didn't notice any problem running it, but of course that was at NERSC. It's probably better to avoid pulling in extra data files you don't need anyhow so doing the curls in the Dockefile works.

jchiang87 (approved):
Looks good.  Thanks!

Issue #299, opened 2022-10-20
stamp.py: if branch with undefined variables


g-braeunlich:
In [stamp.py](imsim/stamp.py), `LSST_SiliconBuilder.draw()`, there is an `if` branch with undefined vars `full_bounds` and `image_pos`:

```py
        if method == 'fft':
            # When drawing with FFTs, large offsets can be a problem, since they
            # can blow up the required FFT size.  We'll guard for that below with
            # a try block, but we can minimize how often this happens by making sure
            # the offset is close to 0,0.
            if abs(offset.x) > 2 or abs(offset.y) > 2:
                # Make a larger image that has the object near the center.
                fft_image = galsim.Image(full_bounds, dtype=image.dtype)
                fft_image[image.bounds] = image
                fft_offset = image_pos - full_bounds.true_center
```

rmjarvis:
Guess that means we don't have any tests that hit that branch.  I can't remember what the impetus for this check was in the old version of this code, from which we copied this over.  Probably @jchiang87 knows?  But for now I guess you can probably just remove this.  Or maybe turn it into a warning, so we can revisit it if something ever causes it to hit this condition.

jchiang87:
I don't recall the reason for this test.  Changing it to a warning seems reasonable.

PR #306, opened 2022-11-15
FFT photon ops: Diffraction

g-braeunlich:
Closes: #299

### DIFF
--- a/config/imsim-config.yaml
+++ b/config/imsim-config.yaml
@@ -205,6 +205,13 @@ stamp:
     world_pos:
         type: InstCatWorldPos
 
+    fft_photon_ops:
+        -
+            type: lsst_diffraction
+            telescope: LSST
+            band: { type: OpsimMeta, field: band }
+            latitude: -30.24463
+
     photon_ops:
         -
             type: TimeSampler
--- a/config/imsim-config.yaml
+++ b/config/imsim-config.yaml
@@ -207,9 +207,15 @@ stamp:
 
     fft_photon_ops:
         -
+            type: TimeSampler
+            t0: 0.0
+            exptime: "@output.exp_time"
+        -
+            type: PupilAnnulusSampler
+            R_outer: 4.18
+            R_inner: 2.55  # M1 inner diameter is 2.558, but we need a bit of slack for off-axis rays
+        -
             type: lsst_diffraction
-            telescope: LSST
-            band: { type: OpsimMeta, field: band }
             latitude: -30.24463
 
     photon_ops:
--- a/imsim/stamp.py
+++ b/imsim/stamp.py
@@ -480,6 +480,11 @@ class LSST_SiliconBuilder(StampBuilder):
             # can blow up the required FFT size.  We'll guard for that below with
             # a try block, but we can minimize how often this happens by making sure
             # the offset is close to 0,0.
+            if not faint and 'fft_photon_ops' in config:
+                photon_ops = galsim.config.BuildPhotonOps(config, 'fft_photon_ops', base, logger)
+            else:
+                photon_ops = []
+
             if abs(offset.x) > 2 or abs(offset.y) > 2:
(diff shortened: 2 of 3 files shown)


imsim/stamp.py
```diff
@@ -480,6 +480,11 @@ def fix_seds(prof):
             # can blow up the required FFT size.  We'll guard for that below with
             # a try block, but we can minimize how often this happens by making sure
             # the offset is close to 0,0.
```
rmjarvis, line 482: This comment is now in the wrong place.  It refers to the below test re offset.  But I think we agreed a while back that that block is broken and we should just remove it.  You can probably go ahead and do that on this PR.

imsim/photon_ops.py
```diff
@@ -200,7 +200,7 @@ def applyTo(self, photon_array, local_wcs=None, rng=None):
             geometry=LSST_SPIDER_GEOMETRY,
             distribution=self.diffraction_rng(rng),
         )
-        photon_array.x, photon_array.y = xy_to_v.inverse(v)
+        photon_array.x[()], photon_array.y[()] = xy_to_v.inverse(v)
```
rmjarvis, line 203: Why this change?

imsim/stamp.py
```diff
@@ -494,7 +499,11 @@ def fix_seds(prof):
                                method='fft',
                                offset=fft_offset,
                                image=fft_image,
-                               wcs=wcs)
+                               wcs=wcs,
+                               photon_ops=photon_ops,
+                               sensor=galsim.Sensor(),
```
rmjarvis, line 504: This won't be necessary once GalSim 2.4.6 is released.  I fixed the test that had been triggering the assert failure.

config/imsim-config.yaml
```diff
@@ -198,6 +198,13 @@ stamp:
     world_pos:
         type: InstCatWorldPos
 
+    fft_photon_ops:
+        -
+            type: lsst_diffraction
+            telescope: LSST
+            band: { type: OpsimMeta, field: band }
+            latitude: -30.24463
```
rmjarvis, line 206: As mentioned on the Galsim PR, this photon op requires pupil sampling, so you need to include a pupil sampler op here too.  Probably also a time sampler.

rmjarvis (changes requested):

imsim/photon_ops.py
g-braeunlich: Indeed, I think, here we dont gain anything by this change.
Before: rhs is a new numpy.ndarray which then will replace lhs which should be garbage collected.
After: rhs is again a new numpy.ndarray which then will be copied to lhs and garbage collected afterwards.

Will remove it again

rmjarvis: Actually, the setter for x and y already write into the existing array.  So in both cases, the temporary returned by xy_to_v.inverse is the one that will be garbage collected.

g-braeunlich: OK. I removed this commit.

imsim/stamp.py
rmjarvis: One other thing here.  The photon sampling of the FFT image uses an rng, so you should also pass `rng=self.rng` here.

rmjarvis:
I would have liked to see a test that just ran the code in fft mode using the config without all this mock stuff.  I seriously cannot tell what the test is even testing, since practically everything is mocked up.  But that's probably just a style difference -- I prefer things that are closer to integration tests, whereas this is closer to what TDD people would call a pure unit test.  Since Chris is in a hurry to merge this, let's just call this fine.  It probably does test the actual bits that you changed in this PR.
Anyway, I think you need to mark it "Ready for review" though, since it's still a work in progress, so I can't officially sign off on it yet.

g-braeunlich:
@rmjarvis Thx for the review! I now marked this PR as ready

rmjarvis (approved):

PR #308, opened 2022-11-17
instcat MJD-OBS bug

jchiang87:
The `MJD-OBS` header keyword was not being set correctly for the `eimage*` and `amp *` output files when using an instance catalog as the source of the metadata for the observation.   This resulted in files with observing dates outside the validity range of the DC2 calibration data, which in turn caused failures when analyzing with the LSST code.


rmjarvis (approved):
LGTM

PR #309, opened 2022-11-29
account for interface changes in rubin_sim v1.0.0

jchiang87:


imsim/sky_model.py
```diff
@@ -60,11 +60,11 @@ def get_sky_level(self, skyCoord):
         with warnings.catch_warnings():
             # Silence astropy IERS warnings.
             warnings.simplefilter('ignore')
-            rubin_sim_sky_model.setRaDecMjd(skyCoord.ra.deg, skyCoord.dec.deg,
+            rubin_sim_sky_model.set_ra_dec_mjd(skyCoord.ra.deg, skyCoord.dec.deg,
```
rmjarvis, line 63: Can we count on everyone having v1.0.0?  Or do we need a version check here?  (Or try/except block.)

rmjarvis (approved):

imsim/sky_model.py
jchiang87: A version check should work, though try/except would be safer, I think.

PR #311, opened 2022-12-08
use skyCatalogs 1.1.0 in CI

jchiang87:


jchiang87:
With Joanne's help, this is now using a tagged release of skyCatalogs for a version prior to the recent changes.  It should be ready to go.

rmjarvis (approved):

PR #314, opened 2022-12-12
set maxN in drawImage for FFT rendering to limit memory usage

jchiang87:
With `fft_photon_ops` enabled, the memory can increase to tens of GB for bright stars without `maxN` set.  Using `maxN=int(1e6)` limits the memory jumps to < 0.5 GB.

### DIFF
--- a/imsim/stamp.py
+++ b/imsim/stamp.py
@@ -475,6 +475,12 @@ class LSST_SiliconBuilder(StampBuilder):
 
         wcs = base['wcs']
 
+        # Set limit on the size of photons batches to consider when
+        # calling gsobject.drawImage.
+        maxN = int(1e6)
+        if 'maxN' in config:
+            maxN = galsim.config.ParseValue(config, 'maxN', base, int)[0]
+
         if method == 'fft':
             if not faint and 'fft_photon_ops' in config:
                 photon_ops = galsim.config.BuildPhotonOps(config, 'fft_photon_ops', base, logger)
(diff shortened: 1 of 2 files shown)


cwwalter (approved):
LGTM

imsim/stamp.py
```diff
@@ -493,7 +493,8 @@ def fix_seds(prof):
                                photon_ops=photon_ops,
                                rng=self.rng,
                                sensor=galsim.Sensor(),
-                               n_subsample=1
+                               n_subsample=1,
+                               maxN=int(1e6)
```
rmjarvis, line 497: I guess this is ok.  I think I'd rather it be a config parameter, so the user can change it, but this is how we currently do the other maxN (when method=phot), so this is fine for now.

rmjarvis (approved):

imsim/stamp.py
jchiang87: I thought about making it a config parameter, but I wasn't entirely sure where it should go.  I guess we should make both occurrences of `maxN` separate config parameters?  Anyway, I'd vote for fixing this later.

rmjarvis: I'd make just one maxN parameter that applies to both branches.  The reasons to change it would apply to both.

rmjarvis: If you're willing to do it now, this is the snippet I'd recommend (at line 477 before the method branching):
```
maxN = int(1e6)
if 'maxN' in config:
    maxN = galsim.config.ParseValue(config, 'maxN', base, int)[0]
```
Then in both calls to drawImage, use `maxN=maxN`.

jchiang87: Thanks, I've implemented this.

PR #315, opened 2022-12-16
Add checkpointing

rmjarvis:
This PR adds checkpointing for the main work of building the eimage.  It uses a fairly generic Checkpointer class, so we could use this for checkpointing other things as well.  (The atmosphere already has its own way of checkpointing, which is separate from this.)

The image builder splits the work into 10 (default) batches and checkpoints at the end of each batch.  It stores the current image being built, the number of objects complete, and the "extra output" builders, which in our case mostly has the centroid file information.  Upon reloading, it starts up where it left off, but splitting the remaining work again into 10 batches.  This seemed useful so if the last batch was taking a really long time and constantly timing out, it would potentially get a few more checkpoint writes in before getting killed to at least make some progress.  It also lets you update (e.g. increase) the number of batches for subsequent runs if a job is not making progress with the current number without having to restart from scratch.

I also tried to be careful about the possibility of the run being killed while writing the checkpoint file.  So it does a little dance where it moves the current file to a backup location before writing the new one.  Then it moves the new file to the regular name, and then finally deletes the backup.  When loading, it checks for a backup, which it uses if present, since it means this process was interrupted during checkpointing.

So far I've only done manual testing, but it seems to work on the config file that I'd been doing profile tests with.  I tested letting it run to completion before restarting (in which case it gets the whole final image from the checkpoint file and doesn't draw anything at all) and also interrupting after a few batches before restarting.  I compared all three output files: eimage, amp, and centroid.  The centroid files were identical.  

The eimage and amp files typically had a few single photon differences.  However, they were different each run.  I investigated this for a while, and I found that they also happen for just repeated runs of the whole thing, without any checkpointing or even batching at all.  Here is a typical difference:
```
$ python check_diff.py */eimage_00017871-1-i-R22_S11-det094.fits
Files output_no_chk/eimage_00017871-1-i-R22_S11-det094.fits and output_skycat/eimage_00017871-1-i-R22_S11-det094.fits differ
    HDU 0 shows differences in 6 pixels
    The maximum absolute difference is 1.000000e+00.
    The maximum relative difference is 8.271299e-04.
    The positions where there are differences are:
         i  	 j  	    im1    	    im2
        1713	3696	2040.000000	2041.000000
        1714	3696	1209.000000	1208.000000
        3540	749	2191.000000	2192.000000
        3540	750	1411.000000	1410.000000
        3660	839	4501.000000	4502.000000
        3661	839	3375.000000	3374.000000
```
So 3 photons moved over 1 pixel.  Sometimes it's only 1 photon, or even 0 occasionally.  The most I saw was 5.  

I narrowed it down to being something in the Silicon accumulate step, and it's always in fairly bright, but non-saturated stars.  I think it's probably due to the openmp stuff doing calculations in slightly arbitrary order leading to numerical rounding differences depending on which steps were done first.  I don't think it's a bug, but it means that we cannot quite rely on images being 100% identical across different runs if OPENMP is enabled.


rmjarvis:
FYI, I still need to write unit tests.  But Jim, if you want to use this for your upcoming runs, I think it's working correctly.

rmjarvis:
OK, I think this is ready for review now.  I added a bunch of tests and fixed some errors in various edge cases.  None of these should affect any runs you've been doing with this Jim, but if you update to the current version of this branch, then any checkpoint files you had made will now be incompatible with the current code.

imsim/lsst_image.py
```diff
@@ -145,16 +288,20 @@ def addNoise(self, image, config, base, image_num, obj_num, current_var, logger)
...
             obj_num:        The first object number in the image.
             current_var:    The current noise variance in each postage stamps.
-            logger:         If given, a logger object to log progress.
+            logger:         A logger object to log progress.
         """
         base['current_noise_image'] = base['current_image']
 
         if logger.isEnabledFor(logging.INFO):
             skyCoord = base['world_center']
+            # Normally skyCoord is a CelestialCoord, but if not make sure this doesn't bork.
```
jchiang87, line 297: I'm curious:  When would `skyCoord` not be a `CelestialCoord`?

jchiang87 (approved):
LGTM, though I noted one minor typo in a log message.   I haven't tried this version yet, but I've been using the earlier code extensively in the current sims, and it seems to have worked well.  Once the images with existing checkpoints are finished, I'll switch over to this version, but that shouldn't prevent this code from being merged.

imsim/checkpoint.py
```diff
@@ -0,0 +1,122 @@
...
+                self.logger.info("Also found file %s. Deleting.", self.file_name_new)
+                os.remove(self.file_name_new)
+        else:
+            # Case A
+            self.logger.info("No checkpoint file %s detected.", self.file_name)
+
+    def save(self, name, data):
+        """Save some data to the checkpoint file under the give name.
+        """
+        self.logger.debug('checkoutpointing: %s',name)
```
rmjarvis, line 69:
```suggestion
        self.logger.debug('checkpointing: %s',name)
```

imsim/lsst_image.py
rmjarvis: It won't be a CelestialCoord if the WCS is not a CelestialWCS.  E.g. in the unit tests for this, I just did a PixelScale wcs, rather than something realistic, which failed the way it had been written.

PR #316, opened 2023-01-06
Use GalSim 2.4.7 fix for atmospheric screen memory sharing.

rmjarvis:
This PR updates the AtmosphericPSF input type to use the new features in GalSim 2.4.7 that let it share memory properly.  Basically, it tells GalSim not to use the proxies, but instead start the workers with the initialization functions that copy over pointers to the global dict where the screens live.

I also got rid of the mokey patch for the InputProxy class, since the correct version of that is now in GalSim.

### DIFF
--- a/etc/standalone_conda_requirements.txt
+++ b/etc/standalone_conda_requirements.txt
@@ -1,6 +1,6 @@
 # conda install --file etc/standalone_conda_requirements should install all required dependencies of imSim including a conda based version of the Rubin pipelines.
 
 stackvana>=0.2021.30
-galsim>=2.4
+galsim>=2.4.7
 dust_extinction>=1.0
 palpy>=1.8.1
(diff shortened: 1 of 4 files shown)


jmeyers314:
This looks good to me, except that building the individual screens appears to be a serial operation right now in a single process.  This can be parallelized to save a little bit of time. See https://github.com/GalSim-developers/GalSim/blob/releases/2.4/galsim/phase_screens.py#L153-L166.  I'll take a quick look to see if I can figure out how to achieve this, but @rmjarvis may also have some ideas.

jmeyers314:
Actually, nope, I lied.  I just had `nproc=1` hard-coded in my test config file by mistake.  Building in parallel looks like it works too.  I think this is good to merge.

jchiang87 (approved):
lgtm.  Running this on Perlmutter with one process per core, i.e., 128 in total, I don't see the memory per process exceeding 3GB, whereas running as individual jobs, the most processes I can run concurrently is ~70, or ~7GB per job.

PR #317, opened 2023-01-19
add checkpointing to flat generation code

jchiang87:
@rmjarvis This would be convenient to add, especially for the flats that use SEDs, since those take so long to run.


rmjarvis (approved):
This all looks right to me.

PR #318, opened 2023-01-20
Optimize diffraction

g-braeunlich:


tests/test_diffraction.py
```diff
@@ -81,82 +81,79 @@ def test_dist_circle() -> None:
...
 def test_field_rotation_matrix_is_correct_near_zenith() -> None:
     """Test if a star near zenith has a field rotation angle similar
     to the angle predicted by the field rotation rate formula."""
 
-    alt = 89.9
-    az = 45.0
-    lat = -30.24463
+    alt = 89.9 / 180.0 * np.pi
+    az = 45.0 / 180.0 * np.pi
+    lat = -30.24463 / 180.0 * np.pi
```
rmjarvis, line 144: Can we switch to using `coord.Angle` instances for these rather than passing floats around with implicit units?
So here `alt = 89.9 * coord.degrees`.  Then when using them, `np.cos(alt)` just works, or you can get the radian version if you need it with `alt.rad`.

tests/test_photon_ops.py
```diff
@@ -328,28 +430,74 @@ def test_config_lsst_diffraction():
...
+                        "dec": "-0.5261230452954583 radians",
+                    },
+                    "latitude": -30.24463,
+                    "disable_field_rotation": True,
+                }
+            ]
         },
+    }
+    galsim.config.ProcessInput(config)
+    galsim.config.BuildPhotonOps(config["stamp"], "photon_ops", config)
```
rmjarvis, line 489: These tests don't seem to actually test anything other than that no exceptions were raised.  Can we add some actual tests of the images here?

rmjarvis (changes requested):

g-braeunlich:
@rmjarvis Thanks a lot for the review.
I pushed changes for I hope everything except the test (still to be done).
What do you think about the option "disable_diffraction" do we want to keep it  or shall I drop the corresponding commit?

g-braeunlich:
@jmeyers314 Got a failing batoid_wcs test after changing the height parameter to `SIMONYI_LOCATION.height`.

```
>       np.testing.assert_allclose(
            np.rad2deg(aob)*3600, cmds['azimuth']*3600,
            rtol=0, atol=2.0
        )
E       AssertionError: 
E       Not equal to tolerance rtol=0, atol=2
E       
E       Mismatched elements: 1 / 1 (100%)
E       Max absolute difference: 2.08372057
E       Max relative difference: 5.34821008e-06
E        x: array(389608.747408)
E        y: array(389610.831128)
```
Do you think this is acceptable?

tests/test_photon_ops.py
rmjarvis: I would expect there to be some test that the code you changed is doing what you think it should do.

g-braeunlich: If it is only that, I tested the code I changed that it is doing what I think it should do by `test_rubin_diffraction_optics_is_same_as_diffraction_and_optics`, `test_rubin_diffraction_produces_spikes` and `test_rubin_diffraction_does_not_show_field_rotation_when_deactivated`.
The deserialization tests make sure that I obtain instances of `RubinDiffraction`, `RubinDiffractionOptics` and `RubinOptics`. From there on the 3 above tests take over.
Indeed, there is a small gap in coverage: I dont test that the values from config end up in the instance being what I expect.
On the other hand side, this is actually should already be covered by galsim.
Still, I could add checks for the values in the instances after deserializing or I also could repeat the 3 configless tests from above.

rmjarvis: There ought to be some test that the image produced here is correct.  I'll leave to you to decide what that means, but right now this test (and the other two like it) is incomplete.

g-braeunlich: Added a commit which compares the behavior of the deserialized photon ops to some reference ops

rmjarvis (approved):

PR #319, opened 2023-01-24
Update docker image

stuartmcalpine:
1) Updated the imsim-env docker image (update dependencies)

2) Updated the install docs to include information about the imsim-env docker image

3) Added a ci workflow that tests imSim within the docker image

### DIFF
--- a/.github/workflows/ci_with_docker.yml
+++ b/.github/workflows/ci_with_docker.yml
@@ -1,0 +1,31 @@
+# Perform imSim unit tests within the latest imSim docker environment image.
+
+name: imSim CI within docker
+
+on:
+  push:
+    branches:
+      - update_docker_image
+  workflow_dispatch: null
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    # Point to the imSim docker image.
+    container:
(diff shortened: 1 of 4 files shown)


cwwalter (approved):
I'm not really a docker user. But the parts of this I understand look OK to me.

I think my only comment would be that for normal users, I'm not sure recommending  docker as the preferred installation method in the text here:

> for that reason we recommend working within the *imSim* *Docker* image, where
 the LSST software stack is pre-installed. Alternatively, you can use the
 *imSim* *Docker* file as a template to build your own *Docker* image, or,
 another option is to work within a *Conda* environment that gets the LSST stack
 through *Stackvana*. The three installation options are detailed below.

is really the best advice.  For most people, using stackvana (or maybe cmvfms if they have it) is going to be easier.

stuartmcalpine:
Thanks Chris,

I didn't realise it was that was round. I've reworded the docs to favour Conda (or at least mentioned that first)

stuartmcalpine:
@jchiang87 good to merge this?

jchiang87:
> @jchiang87 good to merge this?

It's been approved, so I would think yes.

PR #320, opened 2023-02-08
test if checkpointing is enabled before trying to checkpoint

jchiang87:
Sorry about prematurely merging #317.  I didn't notice that the CI didn't run (and neglected to test it myself).  Anyway, this adds a test that checkpointing is enabled before trying to save.


rmjarvis (approved):
Sorry, I should have caught that.

PR #330, opened 2023-04-03
Change docker permissions to work with shifter

stuartmcalpine:
Shifter cant run as root, so cant have imSim as editable install in docker file, reverted it back to static.

Also added a badge on the readme pointing to the built docker image at dockerhub.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -25,8 +25,8 @@ RUN source /opt/lsst/software/stack/loadLSST.bash &&\
     setup lsst_distrib &&\
     mamba install -y --file imSim/etc/docker_conda_requirements.txt &&\
     python3 -m pip install batoid skyCatalogs==1.2.0 gitpython &&\
-    python3 -m pip install -e rubin_sim/ &&\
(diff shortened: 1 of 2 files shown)


cwwalter:
I heard there maybe some issue now about free use of docker.  Do we need to migrate to somewhere else?

jchiang87 (approved):
lgtm

heather999:
I was more worried about this when Docker was planning to drop their free team plan (which we use) - but they received so many complaints that they backed off and have no plans to discontinue free teams.  So for now Dockerhub should be fine.  We can of course start to consider alternatives like the github container registry.  Given the structure of our LSSTDESC GitHub organization plan - github's container registry should be free to use.  We may want to watch how space we utilize, how much data we download from a registry, etc.
The other wrinkle is that images on dockerhub are easily pulled into shifter and that's very convenient.  NERSC is moving to podman and I think that would more naturally allow access to other container registries like github's.  So this is probably all part of a larger shift in how we deal with images and images that we run at NERSC.  Likely worthy of a Computing WG discussion..

Issue #321, opened 2023-02-10
Telescope doesn't load when nproc>1


jmeyers314:
I think this is a bug with the `InputProxy` stuff for the [telescope loader](https://github.com/LSSTDESC/imSim/blob/main/imsim/telescope_loader.py#L185-L195), but haven't been able to pinpoint the problem yet.  (In particular, I'm not sure if the problem is GalSim or ImSim).  

Arises with the following wavefront sensor donut config.  This works fine when `output.nproc=1`, but not when `output.nproc=2`.

```yaml
modules:
  - imsim
  - astropy
  - astropy.time

eval_variables:
  cboresight:
    type: RADec
    ra: 0 deg
    dec: 0 deg
  sband: &band r

input:
  telescope:
    file_name: LSST_r.yaml
    perturbations:
      M2:
        shift: [3e-3, 0, 0]

  atm_psf:
    airmass: 1.1
    rawSeeing:  0.5
    band:  *band
    boresight: $boresight

    # Optional parameters:  (Unless otherwise stated, these are the default values.)
    t0: 0               # seconds
    exptime: 30.0
    kcrit: 0.2          # in units of 1/r0
    screen_size: 409.6  # Default=812.2, which takes a lot of memory, so use this for testing.
    screen_scale: 0.1   # meters
    doOpt: False


image:
  type: Scattered
  xsize: $xsize  # $xsize is populated by output.LSST_CCD.  Could this automatically apply to Scattered.xsize?
  ysize: $ysize
  nobjects: 4
  image_pos: $image_center  # Draw one donut at the center of each WFS CCD

  bandpass:
    file_name:
      type: FormattedStr
      format: LSST_%s.dat
      items:
        - *band
    wave_type: nm
  stamp_size: 3000

  wcs:
    type: Batoid
    camera: "@output.camera"
    boresight: $boresight
    obstime:  # Tried placing this in `eval_variables`, wouldn't work.
      type: Eval
      str: astropy.time.Time('J2022')  # makes ra=dec=0 appear at ~45 zenith angle
    det_name: $det_name
    # band: *band

  noise:
      type: CCD
      gain: 1
      read_noise: 0  # The read noise is applied later.  Only sky noise here.
  random_seed: 57721

gal:
  type: DeltaFunction
  sed:
    file_name: vega.txt
    wave_type: nm
    flux_type: flambda
    # norm_flux_density: 1.65e5
    norm_flux_density: 1.65e4
    norm_wavelength: 500

[...]

PR #334, opened 2023-04-17
#321 Fix telescope input class to work correctly in multiprocessing

rmjarvis:
The telescope proxy was broken for use in multiprocessing as described by @jchiang87 in #321.

The problem is that the proxies we use for multiprocessing communication of input data can only call methods of the underlying object, not access attributes or call hidden (i.e. leading underscore) method like `__getitem__`.

So I changed our telescope dict to a bona fide class, called Telescopes, and forced the access to use `.get` rather than `.__getitem__` (aka `[]`).


jchiang87 (approved):
lgtm, though I'm willing to wait for @jmeyers314 to have the opportunity to comment.

jmeyers314:
LGTM.  Thanks!

jmeyers314 (approved):

PR #336, opened 2023-04-24
Refactor skyCatalogs access

jchiang87:
- This avoids unneeded overhead at startup when simulating multiple CCDs using multiprocessing with galsim by implementing `.getApproxNObjects` and deferring catalog access until needed.
- Also uses `skyCatalogs.PolygonalRegion` to perform object selection for the CCD footprint on the sky.

### DIFF
--- a/imsim/skycat.py
+++ b/imsim/skycat.py
@@ -155,7 +155,7 @@ class SkyCatalogInterface:
         -------
         galsim.GSObject
         """
-        if not self._objects:
+        if not self.objects:
             raise RuntimeError("Trying to get an object from an empty sky catalog")
 
         skycat_obj = self.objects[index]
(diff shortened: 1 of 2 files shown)


imsim/skycat.py
```diff
@@ -171,7 +155,7 @@ def getObj(self, index, gsparams=None, rng=None, exp_time=30):
         -------
         galsim.GSObject
         """
-        if not self.objects:
+        if not self._objects:
```
rmjarvis, line 158: I think this should be `objects`, not `_objects`.  In practice, it might not matter, since probably `getNObjects` will always be called before `getObj`.  But I think logically, this test should be if the loaded object list is empty, not if the objects haven't been loaded from the skyCatalog yet.

rmjarvis (changes requested):
Just one thing I'm not sure is right.

imsim/skycat.py
jchiang87: I guess it is clearer to test `objects`, but my thought process was that `_objects` would be an empty list if skyCatalogs didn't find any objects within the search region.  Anyway, I'll make the change.

rmjarvis: But it could also still be None if you haven't checked yet.

rmjarvis (approved):

PR #342, opened 2023-05-15
Fix telescope multiprocessing

### LINKED ISSUE #338: Race when perturbing telescope?
I've been testing out simulating donuts and then processing them with the active optics framework.  In a recent test, I tried perturbing the telescope (by adding a Zernike-shaped figure error to M1) and then making images at the centers of each raft and each wavefront sensor.  I then iterated over telescope perturbations (which Zernike polynomial to add to M1).  However, something is off when I plot the donuts over different perturbations from a single wavefront sensor:

Donuts

These should all be moderate perturbations around the same fiducial donut, but some images have been made completely in-focus!  Similarly, if I plot an in-focus spot across different telescope perturbations, most look plausible but a few have been turned into donuts!

In-focus

I'm pretty sure which images are effected is random from running this multiple times.

Not sure if it's relevant, but I have been using multiprocessing.  Maybe something is wrong with `TelescopeLoader.setupImage` (I don't see anything obvious though).  Or maybe there's some kind of race where `GetInputObj('telescope', config, base, 'telescope')` returns before `setupImage` is done or is improperly cached?

jmeyers314:
Fixes #338 .

The original problem was that shifting the detector z-position occurred in a single global variable: `Telescopes._d`.  So competing processes would inevitably clobber each others shifted telescope objects leading to the reported race condition.  The solution suggested by @rmjarvis is to store the detector-shifted telescope instances into the `base` config dict, of which each process has its own copy.

In addition to the above change, I added some logger infrastructure in a few places and added a small optimization to `BatoidWCSFactory._field_to_focal`.

### DIFF
--- a/imsim/batoid_wcs.py
+++ b/imsim/batoid_wcs.py
@@ -103,6 +103,16 @@ class BatoidWCSFactory:
         self.rh = self.H2O_pressure/es  # relative humidity
         self.wl = self.wavelength * 1e-3  # nm -> micron
 
+    def _get_det_telescope(self, z_offset, _telescope=None):
+        if _telescope is not None:
+            return _telescope
+        elif z_offset != 0.0:
+            return self.telescope.withLocallyShiftedOptic(
+                "Detector", [0.0, 0.0, -z_offset]  # batoid convention is opposite of DM
+            )
+        else:
+            return self.telescope
+
     def _ICRF_to_observed(self, rc, dc, all=False):
         """
         Parameters
@@ -310,14 +320,7 @@ class BatoidWCSFactory:
             optic=self.telescope,
             wavelength=self.wavelength*1e-9
         )
-        if _telescope is not None:
-            det_telescope = _telescope
-        else:
-            det_telescope = self.telescope
-            if z_offset != 0.0:
-                det_telescope = det_telescope.withLocallyShiftedOptic(
-                    "Detector", [0.0, 0.0, -z_offset]  # batoid convention is opposite of DM
-                )
+        det_telescope = self._get_det_telescope(z_offset, _telescope)
         det_telescope.trace(rv)
         # x/y transpose to convert from EDCS to DVCS
         return rv.y*1e3, rv.x*1e3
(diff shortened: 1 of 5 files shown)


imsim/batoid_wcs.py
```diff
@@ -103,6 +104,8 @@ def __init__(
         self.rh = self.H2O_pressure/es  # relative humidity
         self.wl = self.wavelength * 1e-3  # nm -> micron
 
+        self.logger = logger
```
rmjarvis, line 107: I don't see where this is actually used, but I'd recommend not doing this.  Loggers aren't pickleable, so this immediately makes this class not pickleable either.  Maybe doesn't matter, but if we can avoid doing that, it's probably better.
Generally I prefer to pass in the logger to any function where we would want to use it.  Then if we call if from multiple processes, each process would be able to log to its own logger, not some global self.logger.

imsim/batoid_wcs.py
```diff
@@ -332,6 +338,12 @@ def _focal_to_field(self, fpx, fpy, z_offset=0.0):
         thx, thy : array
             Field angle in radians
         """
+        det_telescope = self.telescope
+        if z_offset != 0.0:
+            det_telescope = det_telescope.withLocallyShiftedOptic(
+                "Detector", [0.0, 0.0, -z_offset]  # batoid convention is opposite of DM
+            )
```
rmjarvis, line 345: Admittedly, this is only 5 lines, one of which is just a closing paren, but it's repeated twice.  Should we make it quick helper function somewhere?  E.g.
```
    def _get_det_telescope(self, z_offset, _telescope=None):
        if _telescope is not None:
            return _telescope
        elif z_offset != 0.0:
            return self.telescope.withLocallyShiftedOptic(
                "Detector", [0.0, 0.0, -z_offset]  # batoid convention is opposite of DM
            )
        else:
            return self.telescope
```

imsim/photon_ops.py
```diff
@@ -53,13 +53,15 @@ def __init__(
...
         camera,
+        logger=None
     ):
         self.telescope = telescope
         self.detector = camera[det_name]
         self.boresight = boresight
         self.sky_pos = sky_pos
         self.image_pos = image_pos
         self.icrf_to_field = icrf_to_field
+        self.logger = logger
```
rmjarvis, line 64: Here too.  I don't think we want to save self.logger.

rmjarvis (changes requested):
Just some small change suggestions.

imsim/batoid_wcs.py
jmeyers314: Ah, okay.  I was using this (possibly incorrectly/poorly) while debugging this issue and thought it might be nice to keep around for the future.  But if it messes with the pickleability then happy to get rid of it.

rmjarvis (approved):

PR #343, opened 2023-05-16
Fft diffraction

g-braeunlich:


imsim/diffraction_fft.py
```diff
@@ -0,0 +1,226 @@
...
+    x_range = np.arange(-w, w + 1)
+    y_range = np.arange(-h, h + 1)
+    [x, y] = xy = np.c_[
+        np.meshgrid(
+            x_range,
+            y_range,
+            indexing="ij",
+            copy=False,
+        )
+    ]
```
rmjarvis, line 105: `np.c_` is pretty obscure and not really necessary here.  This would be more straightforward as
```
x, y = np.meshgrid(x_range, y_range, indexing="ij", copy=False)
xy = np.array([x, y])
```

imsim/diffraction_fft.py
```diff
@@ -0,0 +1,226 @@
...
+        return 1.0, otherwise return l_pix / (r * d_alpha)."""
+        return 1.0 / np.maximum(r * d_alpha / l_pix, 1.0)
+
+    d_pix = 0.5 * l_pix  # Half a pixel
+    return (
+        int_spike_profile(r + d_pix) - int_spike_profile(r - d_pix)
+    ) * arclength_dose(r)
+
+
+def prepare_psf(w: int, h: int, resolution: float, alpha: float) -> np.ndarray:
```
rmjarvis, line 49: I think you don't call this from anywhere?

imsim/diffraction_fft.py
```diff
@@ -0,0 +1,226 @@
...
+    """
+
+    psf_w = psf_h = psf_size  # neglect the outer region (~1/r^2)
+    sin_cos = np.empty(2)
+    e_z_0, e_z = prepare_e_z(latitude)
+    e_focal = e_equatorial(latitude=latitude, azimuth=azimuth, altitude=altitude)
+    field_rotation_sin_cos(e_z_0, e_z, e_focal, exptime, sin_cos)
+    d_alpha = np.arctan2(sin_cos[1], sin_cos[0])
+    rottelpos = np.pi/4. - rottelpos
+    psf = prepare_psf_field_rotation(
```
rmjarvis, line 152: `psf` isn't a great name fro this.  This is just the diffraction spikes, not the whole psf.  Probably `spike_image` would be better.  Or `spike_per_pixel` to point out that it's just the diffraction spikes corresponding to a single pixel in the undiffracted psf.

imsim/diffraction_fft.py
```diff
@@ -0,0 +1,226 @@
...
+        psf_w,
+        psf_h,
+        resolution=pixel_length,
+        alpha=rottelpos,
+        d_alpha=d_alpha,
+    )
+    for region_row, region_col in saturated_clusters(image, full_well):
+        img_region = image[region_row, region_col]
+        diffracted = scipy.signal.convolve2d(psf, img_region, mode="full")
```
rmjarvis, line 161: I think this part of the code would be significantly simpler if `psf` was set up to be the same size/shape as `image`.  Then you could call `convolve2d` with `mode='same'`, and the output would also be the same size as image.  Then adding is just `image += diffracted` without all the extra cropping.

imsim/diffraction_fft.py
```diff
@@ -0,0 +1,226 @@
...
+    regions = []
+    for i, j in pixel_coordinates.T:
+        neighbor_indices = [n for n, r in enumerate(regions) if dist_2d(r, (i, j)) <= 1]
+        pix_region = (slice(i, i + 1), slice(j, j + 1))
+        if not neighbor_indices:
+            regions.append(pix_region)
+            continue
+        neighbor_regions = [regions.pop(n) for n in reversed(neighbor_indices)]
+        regions.append(merge_regions(neighbor_regions + [pix_region]))
+    return regions
```
rmjarvis, line 206: This seems like a lot of work to avoid doing this on any pixels below threshold.  Since convolve2d wants rectangular arrays, why not just return a single range that encompasses all the central pixels.  It's not like it's wrong to include lower surface brightness pixels near the center.  And I suspect it would be faster to just have a single call to convolve2d, even though it would technically be doing more work.

tests/test_diffraction_fft.py
```diff
@@ -0,0 +1,477 @@
...
+    config = create_test_config(
+        XSIZE,
+        YSIZE,
+        STAMP_SIZE,
+        exptime=exptime,
+        mode=mode,
+        enable_diffraction=enable_diffraction,
+    )
+    image = galsim.config.BuildImage(config, logger=None)
+    galsim.fits.write(image, file_name=filename)
```
rmjarvis, line 443: FYI, this is more simply written `image.write(filename)`.

tests/test_diffraction_fft.py
```diff
@@ -0,0 +1,477 @@
...
+    with assert_no_error_logs() as logger:
+        image = galsim.config.BuildImage(config, logger=logger)
+    # To save the image produced here, use:
+    # galsim.fits.write(image, file_name="/tmp/spikes.fits")
+    brightness = image.array
+    # Center of star:
+    [c_x, c_y] = center_of_brighness(brightness)
+    # 2 Pixel tolerance:
+    np.testing.assert_allclose(
+        np.array([c_x, c_y]), np.array([1923.0, 3161.0]), atol=2.0, rtol=0.0
```
rmjarvis, line 261: Where did (1923, 3161) come from?  I was expecting to see one image built with raytracing and the other with fft.
I'd also expect them to be closer than 2 pixels apart.  The center of brightness should be much less than 1 pixel different I would think.  (Especially for non-rotation, but even with rotation.)

tests/test_diffraction_fft.py
```diff
@@ -0,0 +1,477 @@
...
+    )
+    angle, angle_stddev = folded_spike_angle(brightness, c_x, c_y, r_min=10.0)
+    expected_angle = (
+        45.0 * galsim.degrees - config["stamp"]["diffraction_psf"]["rottelpos"]
+    ).deg
+    # 1° tolerance:
+    np.testing.assert_allclose(np.rad2deg(angle), expected_angle, atol=1.0, rtol=0.0)
+    expected_max_spike_width = 5.0
```
rmjarvis, line 269: Where did this come from?  Again, I expected it to match ray tracing.  Not be necessarily less than some nominal value.

tests/test_diffraction_fft.py
```diff
@@ -0,0 +1,477 @@
...
+    np.testing.assert_array_less(np.rad2deg(angle_stddev), expected_max_spike_width)
+    slope, intercept, slope_stderr, intercept_stderr = brightness_log_profile(
+        brightness, c_x, c_y
+    )
+    np.testing.assert_allclose(slope, -2.0, atol=0.6, rtol=0.0)
+    np.testing.assert_allclose(intercept, 12.5, atol=3.0, rtol=0.0)
```
rmjarvis, line 275: ???  What are these numbers?

tests/test_diffraction_fft.py
```diff
@@ -0,0 +1,477 @@
...
+    # Assert some rough evidence, that brightness decays as
+    # brightness ~ exp(intercept) * r**slope:
+    np.testing.assert_array_less(slope_stderr, 0.2)
+    np.testing.assert_array_less(intercept_stderr, 0.8)
```
rmjarvis, line 279: I don't see how these tests match the comment that it's checking "brightness ~ exp(intercept) * r**slope"

tests/test_diffraction_fft.py
```diff
@@ -0,0 +1,477 @@
...
+    np.testing.assert_allclose(np.rad2deg(angle), expected_angle, atol=2.5, rtol=0.0)
+    np.testing.assert_allclose(
+        np.rad2deg(angle_stddev), expected_spike_width / 2., atol=1.0, rtol=0.0
+    )
+    slope, intercept, slope_stderr, intercept_stderr = brightness_log_profile(
+        brightness, c_x, c_y
+    )
+    # Brightness should decay as ~1/r**2:
+    np.testing.assert_allclose(slope, -2.0, atol=0.6, rtol=0.0)
+    np.testing.assert_allclose(intercept, 12.6, atol=3.0, rtol=0.0)
```
rmjarvis, line 314: OK, here is an explanation for the slope=-2.  But I why 12.6?  And 12.5 above?  With tolerance of +- 3?

imsim/opsim_meta.py
```diff
@@ -33,6 +33,30 @@ def get_opsim_md(config, base):
...
+
+    Currently only the type of an Angle will will have an inpact on the processed fields.
+    Angles are assumed to be degree values in opsim meta.
+    """
+    missing_values = set(required_fields) - set(config)
+    if not missing_values:
+        return {}
+    opsim_meta = galsim.config.GetInputObj(
+        "opsim_meta_dict", config, base, "opsim_meta_dict"
+    )
```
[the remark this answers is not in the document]
g-braeunlich: I actually agree: its a bad idea to introduce such a backdoor.
I currently use it to grab alt/az, exptime and rottelpos.
Is there a way to get those values without opsim meta?
If not: it is always quite some ceremony to extract / convert fields from opsim meta and I have to load the values mentioned above in stamp.py and photon_ops.py.
What would you think about renaming `opsim_fallback_values` to `load_opsim_values` and only request the values which I really need? Any other missing values would raise an error. I think this way it would not be a backdoor anymore.

imsim/stamp.py
```diff
@@ -1,8 +1,54 @@
...
+
+    def apply(self, image: galsim.Image) -> None:
+        if self.enabled:
+            apply_diffraction_psf(
+                image.array,
+                rottelpos=self.rottelpos.rad,
+                exptime=self.exptime,
+                latitude=self.latitude.rad,
+                azimuth=self.azimuth.rad,
+                altitude=self.altitude.rad,
```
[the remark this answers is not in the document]
g-braeunlich: Here I am passing over to the lower level routines. This call acts as an interface to that layer. In lower level routines, I am always using radians so there, I dont need to check.
As a consequence, I dont have any overhead from the `Angle` wrapper and I also dont need to introduce a dependency to `galsim` / `coord` in the core routines.
With this in mind it would make life simpler if in the future this part of the code would be rewritten in a compiled language.

Do you still prefer keeping the `Angle` type here?

imsim/diffraction_fft.py
```diff
@@ -0,0 +1,226 @@
...
+    - Assume that the rotation rate is constant over time.
+    - A pixel is approximated by an "angular pixel" (r0-1/2 <= r <= r0+1/2, 0 <= \phi <= \phi_pix(r0), where \phi_pix(r) is chosen such that the arclength r \phi_pix(r) is 1.
+    """
+
+    l_pix = resolution  # length of a pixel
+
+    def arclength_dose(r: np.ndarray) -> np.ndarray:
+        """If the arclength r * d_alpha is smaller than the length of a pixel,
+        return 1.0, otherwise return l_pix / (r * d_alpha)."""
+        return 1.0 / np.maximum(r * d_alpha / l_pix, 1.0)
```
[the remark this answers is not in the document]
g-braeunlich: I did not do that to also allow `d_alpha = 0.`

imsim/diffraction_fft.py
```diff
@@ -0,0 +1,226 @@
...
+    psf_size: int,
+):
+    """2d convolve, leaving image dimension invariant.
+    image: Brightness pixel data
+    rottelpos: Telescope rotation angle [rad]
+    exptime: Exposure time (used to determine the spike width)
+    latitude: Geographic latitude of observators
+    azimuth: Azimuth of the telescope
+    altitude: Altitude of the telescope
+    full_well: Threshold value for saturated pixels
```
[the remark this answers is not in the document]
g-braeunlich: :+1: I will use `brightness_threshold` to make it even clearer.

tests/test_diffraction_fft.py
g-braeunlich: All those magic numbers (here and below) actually do come from a run with raytracing.
I designed the tests as suggested by @cwwalter such that the reference values are generated once using rytracing mode and do not need to be recomputed during CI testing.

(thats also why the comment 
```
    # To compare with the raytracing method, instead use:
    # config = create_test_config(XSIZE, YSIZE, STAMP_SIZE, exptime=0.0, mode=Mode.RAYTRACING)
```
is there).
But here, I guess it would not add that much additional compute time to also integrate the raytracing mode runs.
What do you think?

tests/test_diffraction_fft.py
g-braeunlich: Regarding the intercept value, I just was not sure on how close this should come to the raytracing value / dont have a good intuition for this value. Thus I just left it here as an additional regression test.
Also: any suggestion on how to deal with this value is appreciated.

imsim/diffraction_fft.py
g-braeunlich: Regarding this and your next comment: Are you suggesting to just do a full convolution `image` * `pixel_spikes`, where `pixel_spikes` has the same dimension as `image`?

imsim/diffraction.py
```diff
@@ -343,10 +344,40 @@ def field_rotation_matrix(
     e_z_rot = e_z(t)
     e_h = np.cross(e_focal, e_z_rot)
     e_h_0 = np.cross(e_focal, e_z_0)[None, :]
-    rot = np.empty(np.shape(t) + (2, 2))
     nrm = np.linalg.norm(e_h, axis=-1) * np.linalg.norm(e_h_0, axis=-1)
-    rot[:, 1, 1] = rot[:, 0, 0] = np.einsum("ij,ij->i", e_h, e_h_0) / nrm
-    rot[:, 0, 1] = np.einsum("ij,ij->i", e_z_rot, e_h_0) / nrm
+    out[..., 0] = np.einsum("ij,ij->i", e_h, e_h_0) / nrm
+    out[..., 1] = np.einsum("ij,ij->i", e_z_rot, e_h_0) / nrm
```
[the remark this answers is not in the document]
g-braeunlich: Good idea. But I checked, and it does not work when `out[..., 0]` is a scalar. This is the case when using the FFT variant (would be not the case when only using raytracing)

imsim/diffraction_fft.py
```diff
@@ -0,0 +1,226 @@
...
+    l_pix = resolution  # length of a pixel
+
+    def arclength_dose(r: np.ndarray) -> np.ndarray:
+        """If the arclength r * d_alpha is smaller than the length of a pixel,
+        return 1.0, otherwise return l_pix / (r * d_alpha)."""
+        return 1.0 / np.maximum(r * d_alpha / l_pix, 1.0)
+
+    d_pix = 0.5 * l_pix  # Half a pixel
+    return (
+        int_spike_profile(r + d_pix) - int_spike_profile(r - d_pix)
```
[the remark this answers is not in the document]
g-braeunlich: I tried using the approximation instead. I actually also was not expecting this, this would result in a difference in the angle of ~ 1° (and a difference in its stddev of ~ 3°). I ll leave the integral version and explain it

imsim/opsim_meta.py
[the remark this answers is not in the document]
rmjarvis: Yes, just make them required.  And users can use OpsimMeta types in the config.

imsim/stamp.py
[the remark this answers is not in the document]
rmjarvis: Yes.  If you can document that there is noticeable overhead from using it, I would reconsider, but I'm pretty sure that there is essentially no overhead from the Angle class.

imsim/diffraction_fft.py
rmjarvis: No.  Still find a small region around the center of the PSF with all the >threshold pixels.  
But the thing you convolve with that could be the same size as the image or at least the part of the image that you do the convolution on (it could be some smaller image[bounds] if you want).  
The point is to let scipy handle all the edge details for you with the mode='same' option.

tests/test_diffraction_fft.py
rmjarvis: If the ray tracing is too time consuming to compute in regular CI runs, then you should still have code in there to generate it, but have it not run normally.
E.g.
```
file_name = 'data/raytrace_diffraction_values.npz'
if not os.path.isfile(file_name):
    # ... Run ray tracing code to generate values.
    np.savez(file_name, c_x=c_x, c_y=c_y, ...)

raytrace_data = np.load(file_name)

# ... Use raytrace_data['c_x'], raytrace_data['c_y'], to compare FFT version.
```
Then you run this locally and commit the data file to the repo.  If something changes where we need to recompute things, we delete the file and run again.

imsim/stamp.py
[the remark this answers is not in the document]
g-braeunlich: It is not only, overhead. The `Angle` type is very invasive and I dont only use `np.cos` / `np.sin` on those values.
(E.g., I also use the `%` operator for the 4 fold symmetry, which is not supported for `Angle`).
Therefore at some point, I need to convert.

imsim/diffraction.py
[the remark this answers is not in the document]
g-braeunlich: Resolving this now, but feel free to unresolve, if there is a way to make this work.

imsim/diffraction_fft.py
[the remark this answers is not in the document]
g-braeunlich: Resolving this now, but feel free to unresolve, if you think this might be solved in a simpler way.

imsim/diffraction_fft.py
g-braeunlich: Removed

imsim/diffraction_fft.py
g-braeunlich: Still not sure if I get it.
If I only take a small region with all the >threshold pixels, and make the cutoff spike psf that same size as the region, then the resulting spike width would be determined by the small region - i.e. in most cases << 4000px (="spike_width_cuttoff").
Or do you suggest to have the small region to be at least "spike_width_cuttoff"?

g-braeunlich: OK. But then the resulting convolution still has to be put back to some region in the total image. This still would require cropping, or am I wrong?

rmjarvis: Sure, but that's much easier than what you are currently doing, since you can use the galsim.Bounds functionality.
```
b_spike = galsim.Bounds(star_pos).withBorder(spike_length_cutoff)
b_overlap = b_spike & image.bounds
spike_image = galsim.Image(spike_array, b_spike)
image[b_overlap] += spike_image[b_overlap]
```

imsim/diffraction_fft.py
```diff
@@ -0,0 +1,226 @@
...
+
+    Idea: Calculate the photon dose a single pixel in a distance of r to the
+    spike center receives.
+
+    Approximations:
+    - Assume that the rotation rate is constant over time.
+    - A pixel is approximated by an "angular pixel" (r0-1/2 <= r <= r0+1/2, 0 <= \phi <= \phi_pix(r0), where \phi_pix(r) is chosen such that the arclength r \phi_pix(r) is 1.
+    """
+
+    l_pix = resolution  # length of a pixel
```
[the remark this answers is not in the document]
rmjarvis: Josh, can you look at this part of the code at least?  I still don't understand this scaling parameter.  What are its units even?

tests/test_diffraction_fft.py
```diff
@@ -258,32 +289,56 @@ def test_fft_diffraction_is_similar_to_raytracing_for_0_exptime():
...
     )
+    # Brightness should decay as ~1/r**2:
     np.testing.assert_allclose(slope, -2.0, atol=0.6, rtol=0.0)
-    np.testing.assert_allclose(intercept, 12.5, atol=3.0, rtol=0.0)
-    # Assert some rough evidence, that brightness decays as
-    # brightness ~ exp(intercept) * r**slope:
+    np.testing.assert_allclose(raytrace_data["slope"], -2.0, atol=0.6, rtol=0.0)
+
+    np.testing.assert_allclose(
+        intercept, raytrace_data["intercept"], atol=3.0, rtol=0.0
```
rmjarvis, line 312: atol=3 still seems high to me.  This is only accurate to about 25% tolerance?

g-braeunlich: They are actually much closer. I set the tolerance to 0.5 now

imsim/diffraction_fft.py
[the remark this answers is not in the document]
jmeyers314: @g-braeunlich , I think the relevant scalings you're looking for is https://github.com/GalSim-developers/GalSim/blob/releases/2.4/galsim/phase_psf.py#L642-L654 (especially the third bullet point: aperture grid to real-space (sky) grid) ?

rmjarvis (approved):

cwwalter:
A bit confused... Mike just approved the changes.  Did you add a bunch of new ones?

PR #344, opened 2023-05-17
Enable eups setup

jchiang87:


cwwalter (approved):

cwwalter (approved):
Looks good to me.  

I guess with this change I won't need the 

PYTHONPATH=$IMSIM_DIR:$PYTHONPATH

in my setup file anymore right?

jchiang87:
> I guess with this change I won't need the
> PYTHONPATH=$IMSIM_DIR:$PYTHONPATH
> in my setup file anymore right?

That's right (I have the same.)

Issue #346, opened 2023-06-01
Pass readout config entries to CcdReadout class


jchiang87:
These entries in the `output` section of the config
```
    readout:
        # Convert from e-image to realized amp images
        readout_time: 3.
        dark_current: 0.02
        bias_level: 1000.
        pcti: 1.e-6
        scti: 1.e-6
```
should be passed to the `CcdReadout` class.

rmjarvis:
@jchiang87 Are you working on this?  If not, I can take this one.

jchiang87:
@rmjarvis Please go ahead.  Thanks!

PR #347, opened 2023-06-02
Fix CameraReadout to actually read parameters from the config file.

rmjarvis:
It turns out that CameraReadout wasn't reading the relevant parameters from the config file.  It was just always using the default values for things.  This PR fixes this.

I also added a test that setting bias_level, dark_current, and read_noise =0.0 is sufficient to have the readout process leave a null image unchanged, which was what Vittorio was trying to do, which precipitated issue #346.


jchiang87 (approved):
lgtm

PR #351, opened 2023-06-07
Update telescope_loader.py

jmeyers314:
Fix Kelvin -> Celsius lie.  

(Note, I would have been comfortable with a "commit directly to main" here, but branch protection prevents that.  Is there a workaround for simple documentation fixes like this?)

### DIFF
--- a/imsim/telescope_loader.py
+++ b/imsim/telescope_loader.py
@@ -74,11 +74,11 @@ def parse_fea(config, base, telescope):
     # the bulk temperature and 4 temperature gradients be supplied.
     fea:
       m1m3_temperature:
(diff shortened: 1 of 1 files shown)


rmjarvis (approved):
I think for delta temperature, Kelvin = Celsius.  So I'm not sure that this is actually wrong.  But I'm happy to approve if you prefer calling these Celsius.

jmeyers314:
Thanks.

It's not wrong for the temperature gradients, but the bulk temperatures are supposed to be absolute, not deltas.  Just so happens that 0 Celsius is about the "nominal" temperature.

PR #356, opened 2023-06-16
Implementation of Fringing in imsim

Zhiyuan-G:
This pull request addresses the implementaion of CCD fringing into imsim. 
A few things to notice: 
1. Right now only includes file for yband fringing.
2. The spatial and temporal variation of fringing from sensor to sensor is waited to be implemented as we are still waiting for the data that help characterizes those variations. 

Please let me know if I need to make any changes or adjustments. Thank you!


imsim/sky_model.py
```diff
@@ -108,6 +115,50 @@ def __call__(self, x, y):
...
+        
+        # Load the 4k x 4k normalized map
+        fringing_filename = os.path.join(data_dir, 'fringing_data',
+                                    'e2v-321-fringe-sim-norm-center.fits.gz')
+        
+        fringe_im = fits.open(fringing_filename)[0].data
+        
+        # Interpolate
+        x = np.linspace(0,fringe_im.shape[-1]-1,fringe_im.shape[-1],dtype= int)
+        y = np.linspace(0, fringe_im.shape[0]-1,fringe_im.shape[0],dtype = int )
```
rmjarvis, line 137: Probably worth adding spaces after the commas here.  I'm not a stickler for pep8 style, but these two lines are a bit hard to read without them.

imsim/sky_model.py
```diff
@@ -108,6 +115,50 @@ def __call__(self, x, y):
...
+        the fringing amplitude from sensor to sensor based on the time
+        and its location on the focal plane.
+        Right now just return 1 since we are still waiting for the data.
+        '''
+        
+        level = 1.
+        
+        return(level)
+
+    def __call__(self, x, y):
```
rmjarvis, line 155: Better to make this a named method, rather than call.  Something like `calculate_amplitude` maybe.

imsim/sky_model.py
```diff
@@ -75,6 +79,9 @@ def get_sky_level(self, skyCoord):
         # Return photons/arcsec^2
         value = flux * self.eff_area * self.exptime
         return value
+    
+    def print(self):
+        print('This is a test')
```
rmjarvis, line 84: Spurious from testing I presume?

rmjarvis (changes requested):
The code looks mostly ok.  The main thing that is missing is unit tests.  There are none yet.  You need something that will run the code you added and make sure things seem to be working the way you expect.

Zhiyuan-G:
>

Zhiyuan-G:
opps, sry I miss-clicked the 'close' tab.. @rmjarvis Thank you for your comments. Could you please point me to some unit test examples? I will add it to my code.

cwwalter:
You can see them in:

imSim/tests

config/imsim-config.yaml
```diff
@@ -161,6 +161,8 @@ image:
     vignetting_data_file: LSSTCam_vignetting_data.json
 
     apply_sky_gradient: True
+    
+    apply_fringe: False # Only set to True if simulating y-bands images!
```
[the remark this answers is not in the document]
Zhiyuan-G: I'm getting this error when changing it to your suggestion. 
galsim.errors.GalSimConfigError: Unable to evaluate string "band == 'y'" as a <class 'bool'>
NameError("name 'band' is not defined")

Zhiyuan-G:
I have updated code with the new method and also make adjustments based on Mike's comment. I think the only thing left is to resolve the bug I encountered when trying to set the condition for applying fringing in the config file. See [here](https://github.com/LSSTDESC/imSim/pull/356#discussion_r1270455772).

config/imsim-config.yaml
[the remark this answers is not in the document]
rmjarvis: Are you using the current version of the template config file (on main branch)?  That has band defined in eval_variables as
```
sband: { type: OpsimData, field: band }
```
which should define the name `band` to use in other eval statements.

imsim/sky_model.py
```diff
@@ -108,7 +113,102 @@ def __call__(self, x, y):
...
+            # Set variation level to unity if True, interpolated value if False.
+            if np.isnan(level):
+                return 1
+            else:
+                return level
+        else:
+            return 1
+        
+
+    def cal_fringe_amplitude(self,x,y):
```
rmjarvis, line 200: Please type out the word calculate here.  Cal can be short for too many things, so better to be clear than brief.

rmjarvis:
I think you probably forgot to commit your new test file.  There are still no unit tests here.

Zhiyuan-G:
I have rebased the branch, updated the code and added a test file. I think the code is good to go.

rmjarvis:
Looks like you messed up the rebase.  There are lots of other things on this branch now.  To fix:
```
git checkout main
git pull
git checkout fringing
git rebase -i main
```
Select only the commits that belong on this branch by deleting anything from other branches.  Feel free to reorder them if you want or squash ones you want to combine into a single commit.  But mostly just get rid of the extra ones that got added.  Exit the text file.
Resolve any conflicts if any show up.  `git rebase --continue` to continue the rebase after fixing.
Then
```
git push --force
```
Hope this helps!

Zhiyuan-G:
Thanks! Ok, I think I've fixed it.

tests/test_config/1-star.yaml
```diff
@@ -0,0 +1,19 @@
+# For testing purpose
```
cwwalter, line 1: Is this file used? Is anything else in this test_config directory?

tests/test_fringing.py
```diff
@@ -0,0 +1,47 @@
...
+    
+    # Testing a CCD outside the interpolation range of OH spatial variation
+    ra = 53
+    dec = -33
+    wcs = make_batoid_wcs(ra, dec, rottelpos, mjd, band, 'LsstCam')
+    ccd_fringing = CCD_Fringe(img_wcs = wcs,c_wcs=[cra, cdec],spatial_vary= True)
+    fringe_map = ccd_fringing.calculate_fringe_amplitude(xarr,yarr)
+    
+    # Check if the default fringing varaition is within +/- 0.2%
+    np.testing.assert_approx_equal(fringe_map.max(),1.002,significant=4)
```
cwwalter, line 43: It looks like you are testing the maximum. What if it doesn't work at all and zero is returned everywhere?

Will that pass the test?

cwwalter (commented):
Hi ZhiYuan I added some comments and questions.

imsim/lsst_image.py
```diff
@@ -43,7 +45,8 @@ def setup(self, config, base, image_num, obj_num, ignore, logger):
         extra_ignore = [ 'image_pos', 'world_pos', 'stamp_size', 'stamp_xsize', 'stamp_ysize',
                          'nobjects' ]
         opt = { 'size': int , 'xsize': int , 'ysize': int, 'dtype': None,
-                'apply_sky_gradient': bool, 'camera': str, 'nbatch': int }
+                'apply_vignetting': bool, 'apply_sky_gradient': bool, 'apply_fringe': bool,
```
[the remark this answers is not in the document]
cwwalter: Confirming they are not on the version on main anymore. Here is what is there now.

```
        opt = { 'size': int , 'xsize': int , 'ysize': int, 'dtype': None,
                'apply_sky_gradient': bool, 'camera': str, 'nbatch': int }
```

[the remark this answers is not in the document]
Zhiyuan-G: Thanks. Fixed!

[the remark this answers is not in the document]
Zhiyuan-G: Fixed.

imsim/lsst_image.py
```diff
@@ -264,11 +267,23 @@ def addNoise(self, image, config, base, image_num, obj_num, current_var, logger)
             det_name = base['det_name']
             camera = get_camera(self.camera_name)
             logger.info("Applying vignetting according to radial spline model.")
-            radii = Vignetting.get_pixel_radii(camera[det_name])
-            sky.array[:] *= self.vignetting.apply_to_radii(radii)
-
+            sky.array[:] *= self.vignetting(camera[det_name])
```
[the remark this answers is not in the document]
Zhiyuan-G: I think I unintentionally delete some code related to the vignetting block. I have fixed this by updating this part of code by copying the one on the main branch in the last push.

imsim/sky_model.py
```diff
@@ -108,7 +113,102 @@ def __call__(self, x, y):
...
+        kpow = -(H + 1.0) / 1.2
+        
+        A = np.zeros((n, n), complex)
+
+        kvec = np.fft.fftfreq(n)
+        k0 = kvec[n // 64]
+        kx, ky = np.meshgrid(kvec, kvec, sparse=True, copy=False)
+        ksq = kx ** 2 + ky ** 2
+        m = ksq > 0
+        random_seed = random.randint(1, 1000000)
```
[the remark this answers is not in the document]
Zhiyuan-G: Thanks for pointing this out. I've fixed this issue in the lasted update. Now the random seed number for each sensor is set to be equal to be its 'det_num'.

tests/data/1-star.txt
```diff
@@ -0,0 +1,20 @@
...
+moonphase 3.8193650
+moonra 256.4036553
+nsnap 2
+obshistid 230
+rottelpos 0.0000000
+seed 230
+seeing 0.8662850
+sunalt -32.7358290
+vistime 33.0000000
+seqnum 0
```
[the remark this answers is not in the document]
Zhiyuan-G: Deleted. I previously used this file to generate some images for visual inspection on fringing.

tests/test_config/1-star.yaml
Zhiyuan-G: save as above.

tests/test_fringing.py
Zhiyuan-G: I have added checks for checking zero or nan values in the test script in the lasted update.

imsim/lsst_image.py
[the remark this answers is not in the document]
jchiang87: They're not the same.  For running in mutlprocessing, you need to use the `.apply_to_radii(radii)` version.  [Here is what is currently in main](https://github.com/LSSTDESC/imSim/blob/e34819956dc3f314929859f67b23a5ecfc494fff/imsim/lsst_image.py#L268).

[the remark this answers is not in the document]
Zhiyuan-G: Right. I have fixed it in my latest push.

Zhiyuan-G:
I have made adjustments based on all the previous comments. Major changes include stretching the spatial variation OH line to cover the whole focal plane and updating testing file. The code is ready to check in.

cwwalter (approved):
This LGTM now, but it would be good to have another review from an expert like Mike, Jim, or Josh also.

imsim/lsst_image.py
```diff
@@ -266,9 +268,27 @@ def addNoise(self, image, config, base, image_num, obj_num, current_var, logger)
             logger.info("Applying vignetting according to radial spline model.")
             radii = Vignetting.get_pixel_radii(camera[det_name])
             sky.array[:] *= self.vignetting.apply_to_radii(radii)
-
+            
+        if self.apply_fringing:
+            # get det number
+            det_num = base['output']['det_num']['current'][0]
+            # get center ra/dec
+            cra = config['wcs']['boresight']['cboresight']['_kd'][1]\
```
jchiang87, line 276: I poked around a bit and found something that's cleaner and probably more robust than these 10 levels of `[...]`s:
```
boresight, _ = galsim.config.ParseValue(config['wcs'], 'boresight', base, galsim.CelestialCoord)
```
I would change the interface to `CCD_Fringing`, so that you call it like this:
```
ccd_fringing = CCD_fringing(img_wcs, boresight, seed=det_num, spatial_vary=True)
```
Then in `fringe_variation_level`, I would do
```
dx = self.boresight.ra.deg - self.img_wcs.center.deg
dy = self.boresight.dec.deg - self.img_wcs.center.deg
```
I'm wondering a bit about the data in `data/skyline_var.pkl`, however.  Does just doing linear interpolation in ra, dec still work reasonably for declinations close 90 degrees?  It seems like simulating sky regions near the poles of the coordinate system would be problematic.

imsim/sky_model.py
```diff
@@ -108,7 +113,103 @@ def __call__(self, x, y):
...
+    
+    The function call operator returns the normlized fringing level
+    as a function of pixel coordinates relative to the value at the 
+    CCD center.
+    """
+    def __init__(self,img_wcs,c_wcs,seed,spatial_vary):
+        self.img_wcs = img_wcs
+        self.c_wcs = c_wcs
+        self.spatial_vary = spatial_vary
+        self.seed = seed
```
jchiang87, line 129: You might consider using the galsim RNG instead of seeding your own.

jchiang87 (changes requested):

imsim/lsst_image.py
rmjarvis: This is better, but still not great.  You shouldn't rely on the `config['wcs']` field to have a `boresight` parameter.  That's specific to the BatoidWCS, and is not generic.  If a user uses a different WCS type, this will break.
Rather, you should add a `boresight` parameter to the `image` field, which is required whenever `apply_fringing` is True.  In the default config file, you should set this as `boresight: '$boresight'`, since we have this in our `eval_variables` section, so it can be trivially evaluated.

imsim/sky_model.py
```diff
@@ -108,7 +113,103 @@ def __call__(self, x, y):
...
+        
+        A = np.zeros((n, n), complex)
+
+        kvec = np.fft.fftfreq(n)
+        k0 = kvec[n // 64]
+        kx, ky = np.meshgrid(kvec, kvec, sparse=True, copy=False)
+        ksq = kx ** 2 + ky ** 2
+        m = ksq > 0
+        random_seed = self.seed
+        gen = np.random.RandomState(random_seed)
```
rmjarvis, line 146: Don't use `np.random.RandomState`.  They don't guarantee consistent values across numpy versions, so that means the simulated values won't necessarily be portable across different systems.  Always use GalSim random number generators for any random values in simulations.  In this case, I think we do want to use the same random seed for a given detector every time.  So we don't actually want to tie it to the RNG that other things in the config file are using.  (Otherwise, different exposures would get different fringing patterns.)  So using det_num for the seed is a good idea.  But this line should be
```
gen = galsim.BaseDeviate(random_seed).np
```
The `np` property returns a numpy-style generator that can be used just like a np.random.Generator() instance.

imsim/lsst_image.py
Zhiyuan-G: I tried adding `boresight: '$boresight'` to the `image` field in parallel with `apply_fringing` parameter. However, I'm getting the error `galsim.errors.GalSimConfigError: Unexpected attribute boresight found`.

jchiang87: I think you need to add
```
'boresight': galsim.CelestialCoord,
```
to the `opt` dictionary [here](https://github.com/LSSTDESC/imSim/blob/30f7396ccceb8309989da3369c7923f55d3f537d/imsim/lsst_image.py#L46).

imsim/lsst_image.py
```diff
@@ -43,7 +43,8 @@ def setup(self, config, base, image_num, obj_num, ignore, logger):
         extra_ignore = [ 'image_pos', 'world_pos', 'stamp_size', 'stamp_xsize', 'stamp_ysize',
                          'nobjects' ]
         opt = { 'size': int , 'xsize': int , 'ysize': int, 'dtype': None,
-                'apply_sky_gradient': bool, 'camera': str, 'nbatch': int }
+                 'apply_sky_gradient': bool, 'apply_fringing': bool, 
+                 'boresight':galsim.CelestialCoord,'camera': str, 'nbatch': int}
```
rmjarvis, line 47:
```suggestion
                 'boresight':galsim.CelestialCoord, 'camera': str, 'nbatch': int}
```

imsim/lsst_image.py
```diff
@@ -57,8 +58,9 @@ def setup(self, config, base, image_num, obj_num, ignore, logger):
             self.vignetting = None
 
         self.apply_sky_gradient = params.get('apply_sky_gradient', False)
-
+        self.apply_fringing = params.get('apply_fringing',False)
```
rmjarvis, line 61:
```suggestion
        self.apply_fringing = params.get('apply_fringing', False)
```

imsim/lsst_image.py
```diff
@@ -57,8 +58,9 @@ def setup(self, config, base, image_num, obj_num, ignore, logger):
...
         self.camera_name = params.get('camera')
+        self.boresight = params.get('boresight')
```
rmjarvis, line 63: boresight is required if apply_fringing is True.  You need to check that and raise an exception if not.

rmjarvis: And also test that the error is raised if the user messes this up.

tests/test_fringing.py
```diff
@@ -0,0 +1,43 @@
...
+    np.testing.assert_approx_equal(np.std(np.diag(fringe_map)),0.0014,significant=2)
+    
+    # Check the min/max of fringing varaition for the current offset.
+    np.testing.assert_approx_equal(fringe_map.max(),1.00205,significant=4)
+    np.testing.assert_approx_equal(fringe_map.min(),0.99794,significant=4)
+    
+    
+if __name__ == '__main__':
+    test_fringing()
+
```
rmjarvis, line 43:
```suggestion
```

imsim/sky_model.py
```diff
@@ -108,7 +113,103 @@ def __call__(self, x, y):
...
+        phase = 2 * np.pi * gen.uniform(size=(n, n))
+        A[m] = ksq[m] ** kpow * gen.normal(size=(n, n))[m] * np.exp(1.j * phase[m]) * np.exp(-ksq[m] / k0 ** 2)
+
+        return np.fft.ifft2(A)
+
+
+    def simulate_fringes(self,n = 1.2,n1=1.5, amp=0.002, nwaves_rms=10.):
+        '''
```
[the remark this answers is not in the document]
rmjarvis: I think amp should be a run-time modifiable parameter.  That seems like a scientifically useful option.

tests/test_fringing.py
```diff
@@ -0,0 +1,42 @@
...
+    dec = -35.76
+    wcs = make_batoid_wcs(ra, dec, rottelpos, mjd, band, 'LsstCam')
+    
+    ccd_fringing = CCD_Fringing(img_wcs = wcs,boresight=[cra, cdec],seed = det_num, spatial_vary= True)
+
+    fringe_map = ccd_fringing.calculate_fringe_amplitude(xarr,yarr)
+    
+    # Check for zero values in the fringe map.
+    if (np.all(fringe_map) != True) or (True in np.isnan(fringe_map)):
+        raise ValueError(" 0 or nan value in the fringe map!")
```
rmjarvis, line 31: This doesn't make sense in a test.  Seems like this code should be in the main code, and then a test should make sure it happens and test that the right exception is raised.

imsim/sky_model.py
```diff
@@ -108,8 +112,130 @@ def __call__(self, x, y):
...
+            
+            # Check if the angluar offset of the current sensor is outside 
+            # the coverage field of the OH data we currently have. 
+            # Set variation level to unity if True, interpolated value if False.
+            if np.isnan(level):
+                return 1
+            else:
+                return level
+        else:
+            return 1
```
rmjarvis, line 218: Are all three of these branches tested?

rmjarvis, line 218: Are all three of these branches tested?

tests/test_fringing.py
rmjarvis, line 31: This doesn't make sense in a test.  Seems like this code should be in the main code, and then a test should make sure it happens and test that the right exception is raised.

tests/test_fringing.py
```diff
@@ -0,0 +1,42 @@
...
+    
+    # Check std of the diagnoal of fringe map.
+    np.testing.assert_approx_equal(np.std(np.diag(fringe_map)),0.0014,significant=2)
+    
+    # Check the min/max of fringing varaition for the current offset.
+    np.testing.assert_approx_equal(fringe_map.max(),1.00205,significant=4)
+    np.testing.assert_approx_equal(fringe_map.min(),0.99794,significant=4)
```
rmjarvis, line 38: Please add spaces after these (and in general all) commas.

tests/test_fringing.py
```diff
@@ -0,0 +1,42 @@
...
+    
+    xarr, yarr = np.meshgrid(range(4096), range(4004))
+
+    
+    # Testing a CCD with an arbitrary location on the focal plane.
+    ra = 54.86
+    dec = -35.76
+    wcs = make_batoid_wcs(ra, dec, rottelpos, mjd, band, 'LsstCam')
+    
+    ccd_fringing = CCD_Fringing(img_wcs = wcs,boresight=[cra, cdec],seed = det_num, spatial_vary= True)
```
rmjarvis, line 25: Normal style puts a space after commas, but not around = signs in a parameter listing.

rmjarvis: cf. https://peps.python.org/pep-0008/

imsim/lsst_image.py
```diff
@@ -266,9 +268,24 @@ def addNoise(self, image, config, base, image_num, obj_num, current_var, logger)
             logger.info("Applying vignetting according to radial spline model.")
             radii = Vignetting.get_pixel_radii(camera[det_name])
             sky.array[:] *= self.vignetting.apply_to_radii(radii)
-
+            
+        if self.apply_fringing:
+            # get det number
+            det_num = base['output']['det_num']['current'][0]
+            # get boresight ra/dec
+            boresight = [self.boresight.ra.deg,self.boresight.dec.deg]
```
rmjarvis, line 276: Don't do this.  Let CCD_Fringing take a CelestialCoord object for its boresight parameter, rather than using an adhoc data structure (i.e. the list of ra, dec, both in degrees).

imsim/sky_model.py
```diff
@@ -108,8 +112,130 @@ def __call__(self, x, y):
...
+        
+        if self.spatial_vary:
+            # Load 2d interpolator for OH spatial variation
+            filename = os.path.join(data_dir, 'fringing_data',
+                                        'skyline_var.pkl')
+            with open(filename, 'rb') as f:
+                interp = pickle.load(f)
+                
+            dx = self.boresight[0] - self.img_wcs.center.ra.deg 
+            dy = self.boresight[1] - self.img_wcs.center.dec.deg
```
rmjarvis, line 205: If boresight is a CelestialCoord, then this becomes:
```
dx, dy = self.boresight.project(self.img_wcs.toWorld(image.true_center))
dx = dx.deg
dy = dy.deg
```
(Although I would delay getting the degree versions until you use them in the line below.)

This has the added advantage that it would work with any wcs, rather than specifically the batoid wcs.  (The `center` attribute is an implementation detail of that wcs type.)

Although note that this function currently doesn't have access to the `image`, so you probably want to compute and store the combination `image.wcs.toWorld(image.true_center)` in the constructor rather than store `img_wcs` as you currently do.

rmjarvis: Oh, and it also has the other advantage that the scale in the RA direction doesn't change as you move far from the equator.  I'm pretty sure that must be a bug in the current implementation.

jmeyers314: That's a bug in BatoidWCS?  What's the bug?

rmjarvis: Not in BatoidWCS.  It's a bug in these lines.  If you're near the north pole, delta RA is much larger than the actual angle on the sky.  There's a cos(dec) factor that is neglected here, but which `project` does correctly.

rmjarvis: Actually, thinking about this more, does this imply that you have the same bug in the way you made the interpolation file?  If you based this on observations of fringing data, then I think this means you would have hard-coded a cos(dec) factor in there corresponding to whatever declination the observations were made.

rmjarvis: I talked to Chris about my above concern, and I think the interpolation is probably fine.  But you should switch to using project for the position (dx,dy) you use for calling the interpolation.

imsim/sky_model.py
```diff
@@ -108,8 +112,130 @@ def __call__(self, x, y):
...
+        """
+        Parameters
+        ----------
+        img_wcs : 
+            WCS of the current image.
+        boresight : `array`
+            Array stores the center RA/DEC of boresight.
+        seed : `int`
+            Random seed number for each CCD. 
+            This number should be the det_num for the current sensor
```
jchiang87, line 134: I don't think `det_num` should be used as the seed. It's reused across different cameras, so not unique to a particular CCD, and one can make yaml configs for a given camera where the det_num to CCD mapping isn't preserved.  If we switch to using the GalSim random number generator, then we can use the detector serial number as the seed, which would have the desired behavior.   Also the detector serial numbers will tell us which kind of CCD we are considering, i.e., e2V vs ITL from the first three letters of the serial number, so we can account for different fringing behavior for the different types that way.

Zhiyuan-G: Using serial number sounds good. Do you know where can I access that info?

jchiang87: Here's some example code:
```
import imsim
camera = imsim.get_camera()

serial_number = camera[det_name].getSerial()
```

Zhiyuan-G: Thank you. I will adjust the code accordingly.

imsim/lsst_image.py
```diff
@@ -266,9 +273,25 @@ def addNoise(self, image, config, base, image_num, obj_num, current_var, logger)
...
+            # Use serial number as random seed number to make sure the height map 
+            # for the same sensor is the same for different exposures.
+            camera = get_camera(self.camera_name)
+            det_name = base['det_name']
+            serial_number = camera[det_name].getSerial()
+            # Only apply fringing to e2v sensors.
+            if serial_number[:3] == 'E2V':
+                ccd_fringing = CCD_Fringing(true_center=image.wcs.toWorld(image.true_center),
+                                            boresight=self.boresight,
+                                            seed=int(serial_number[-3:]), spatial_vary=True)
```
rmjarvis, line 287: This is ok, but maybe fragile.  If they change the structure of the string, this might start failing.  More robust is `seed=hash(serial_number)`.

imsim/sky_model.py
```diff
@@ -108,8 +112,124 @@ def __call__(self, x, y):
...
+    The function call operator returns the normlized fringing level
+    as a function of pixel coordinates relative to the value at the 
+    CCD center.
+    """
+    def __init__(self, true_center, boresight, seed, spatial_vary):
+        """
+        Parameters
+        ----------
+        img_wcs : 
+            WCS of the current image.
```
rmjarvis, line 129: Should add a doc line for true_center.

imsim/sky_model.py
```diff
@@ -108,8 +112,124 @@ def __call__(self, x, y):
...
+        boresight : `array`
```
rmjarvis, line 130: This is now a CelestialCoord.

imsim/sky_model.py
rmjarvis: The final line in the doc for seed is now incorrect.

tests/test_fringing.py
```diff
@@ -0,0 +1,60 @@
...
+            'nobjects': 0,
+        },
+    }
+
+    galsim.config.ProcessInput(config)
+    image = galsim.config.BuildImage(config)
+
+    ccd_fringing = CCD_Fringing(true_center=image.wcs.toWorld(image.true_center),
+                                boresight=world_center,
+                                seed=serial_num, spatial_vary=True)
```
rmjarvis, line 44: Should add a test with spatial_vary=False too.

imsim/lsst_image.py
Zhiyuan-G: Got it.  I am thinking the line getting the sensor type `if serial_number[:3] == 'E2V'` maybe also fail? Is there any better options for that one?

rmjarvis: Mm.  That's also maybe fragile.  But less so.  It's probably unlikely they'll change the first three characters to something else.  But I could imaging them wanting to add letters at the end or underscores somewhere.  Who knows.  hash will convert any string to an integer.

imsim/lsst_image.py
```diff
@@ -57,9 +57,16 @@ def setup(self, config, base, image_num, obj_num, ignore, logger):
             self.vignetting = None
 
         self.apply_sky_gradient = params.get('apply_sky_gradient', False)
-
+        self.apply_fringing = params.get('apply_fringing', False)
+        if self.apply_fringing:
+            try:
+                self.boresight = params.get('boresight')
+            except:
```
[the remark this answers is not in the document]
rmjarvis: Also, you should write a test that checks that this works the way you expect.  I.e. that a GalSimConfigError is raised when boresight is missing.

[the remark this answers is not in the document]
Zhiyuan-G: I have fixed it in the code. But I'm not sure what's the best approach to test it. I tried adding a `boresight` parameter in the config dict created in the test file but I was getting galsim error.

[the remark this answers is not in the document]
rmjarvis: It's quicker for me to just do this, but please take a look over the commits I added.  The last commit in particular is the one I meant for this point.

rmjarvis (approved):

jchiang87 (approved):

PR #357, opened 2023-06-22
Fix scattered positions

jmeyers314:
Same as https://github.com/GalSim-developers/GalSim/pull/1190, just for `LSST_ImageBuilder`.

### DIFF
--- a/imsim/lsst_image.py
+++ b/imsim/lsst_image.py
@@ -179,7 +179,9 @@ class LSST_ImageBuilder(ScatteredImageBuilder):
                 "Both image_pos and world_pos specified for LSST_Image.",
                 (config['image_pos'], config['world_pos']))
 
-        if 'image_pos' not in config and 'world_pos' not in config:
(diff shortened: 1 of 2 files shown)


jmeyers314:
@rmjarvis , I think you wrote this test?  I'm not sure what the failing line is trying to do.  Can you take a look?

rmjarvis:
I'm pretty sure that bit was to provide coverage of [this branch](https://github.com/LSSTDESC/imSim/blob/main/imsim/lsst_image.py#L177).  I think with your code change here, you can change the failing line to
```
config['image']['world_pos'] = config['image']['image_pos'] = config['stamp'].pop('image_pos')
```
The point is just that it is an error if world_pos and image_pos are both present in the config file.

jmeyers314:
> you can change the failing line to ...

Yep, that seems to have worked. Thanks.

rmjarvis (approved):

PR #358, opened 2023-06-22
Remove galsim 2.4 requirement in templates.py

cwwalter:
We now need at least GalSim 2.4 to run at all. So, this should be a dependency. Not specified here.

Removing TODO and if statement.


cwwalter:
Gah... started this branch off wrong branch. closing.

PR #365, opened 2023-06-27
Update the installation part of the documentation.

cwwalter:
Updating the installation part of the sphinx code.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -49,7 +49,6 @@ jobs:
 
             - name: Install pip deps
               run: |
-                # We need to get batoid onto conda, but for now, this is a separate step.
                 pip install skyCatalogs==1.4.0-rc5
                 conda info
                 conda list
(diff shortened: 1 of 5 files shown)


cwwalter:
Jim,  you don't need to look at the documentation changes I made here.

But, there are changes to the CI file (and conda standalone requirements) I made so that things worked with a recent version of the stack/python.

Can you take a look at and review that? Thanks.

.github/workflows/ci.yml
```diff
@@ -49,9 +49,6 @@ jobs:
 
             - name: Install pip deps
               run: |
-                # We need to get batoid onto conda, but for now, this is a separate step.
-                pip install batoid
-                pip install batoid_rubin
```
jchiang87, line 54: Just to be clear: we don't need these because they are available from the conda distribution now?

cwwalter: That's right. I added them to the standalone-requirements file.

Actually we don't need rubin-sim either.  But, I kept it this way so as to not download the parts we need.

jchiang87 (approved):
LGTM.

PR #368, opened 2023-06-30
make vignetting an InputType so that it can be used in stamp.py and in lsst_image.py

jchiang87:


jmeyers314 (approved):
This looks pretty good to me.  A couple questions though:
 - Should we require `input.vignetting` if using a `RubinOptics` photon op?
 - What happens if `input.vignetting` is populated and we're photon-shooting without `RubinOptics`?  are the FFT and shoot objects still consistent in this case (i.e., does the vignetting affect the non-RubinOptics shoot objects?)

jchiang87:
Thanks, Josh!

> Should we require input.vignetting if using a RubinOptics photon op?

I think so, but it wasn't obvious to me how to couple those two.

>  What happens if input.vignetting is populated and we're photon-shooting without RubinOptics? are the FFT and shoot objects still consistent in this case (i.e., does the vignetting affect the non-RubinOptics shoot objects?)

In that case, the rendered fluxes won't be consistent.  We'd need to apply the vignetting scaling explicitly to the non-RubinOptics shoot objects as well.

I'll make a new issue for these items, and we can have another PR to implement them.

rmjarvis:
> Should we require input.vignetting if using a RubinOptics photon op?

I'd suggest not.  At most we might issue a warning that FFT objects won't be vignetted.  There may be reasons a user would want to do that, but emitting a warning seems appropriate.

rmjarvis (approved):

Issue #352, opened 2023-06-07
Add surface figure errors as optional output


jmeyers314:
It'd be nice to write out the figure errors of each surface produced by the FEA and AOS inputs as an optional output for reality checks.

cwwalter:
Josh is this a short or long term wish?  Should it be included in the immediate 2.0 milestone?

jmeyers314:
I think this is fairly straightforward to implement, so let's make it a 2.0 milestone.

cwwalter:
OK.

rmjarvis:
Done in #372

PR #372, opened 2023-07-10
Add surface sag as optional output

jmeyers314:
Adds a `sag` output field for optionally writing out maps of each optical surface's "sag", i.e., the height above or below the surface's vertex along the optic axis.  Addresses #352 .

### DIFF
--- a/imsim/sag.py
+++ b/imsim/sag.py
@@ -1,0 +1,84 @@
+import galsim
+from galsim.config import ExtraOutputBuilder, RegisterExtraOutput, GetInputObj
+from galsim.config.input import ParseValue, GetAllParams
+import numpy as np
+import batoid
+from .telescope_loader import infer_optic_radii
+
+
+class SagBuilder(ExtraOutputBuilder):
+    """Build surface sag maps.  The sag is the height above or below a plane
+    perpendicular to the optic axis.  It characterizes the shape of the optic.
+
+
+    Required config inputs:
+        file_name: str
+            Name of file to write sag images to.
+
+    Optional config inputs:
+        nx: int
+            Size of sag map images in pixels.  Default: 255.
+    """
+    def initialize(self, data, scratch, config, base, logger):
+        req = {'file_name': str}
+        opt = {'nx': int}
+        kwargs, safe = GetAllParams(config, base, req=req, opt=opt)
+        self.nx = kwargs.pop('nx', 255)
+        self.final_data = None
+
+    def finalize(self, config, base, main_data, logger):
+        telescope = GetInputObj(
+            'telescope',
+            config,
+            base,
+            'opd'
+        ).fiducial
+        self.final_data = []
+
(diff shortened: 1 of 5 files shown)


imsim/sag.py
```diff
@@ -0,0 +1,82 @@
+import galsim
+from galsim.config import ExtraOutputBuilder, RegisterExtraOutput, GetInputObj
+from galsim.config.input import ParseValue, GetAllParams
+import numpy as np
+import batoid
+from .telescope_loader import infer_optic_radii
+
+
+class SagBuilder(ExtraOutputBuilder):
+    """Build surface sag maps.
```
rmjarvis, line 10: I'm sure for the people who know, sag is appropriate jargon.  But at least once, you should define what this word means.

imsim/sag.py
```diff
@@ -0,0 +1,82 @@
...
+
+    Required config inputs:
+        file_name: str
+            Name of file to write OPD images to.
```
rmjarvis, line 14: Also OPD

imsim/sag.py
```diff
@@ -0,0 +1,82 @@
...
+            out = np.nan * np.ones((self.nx, self.nx))
+            out[ww] = optic.surface.sag(xx[ww], yy[ww])
+
+            dx = (xs[1] - xs[0])*outer
+            dy = dx
+
+            x0, y0, z0 = optic.coordSys.origin
+            R = optic.coordSys.rot
+
+            sag_img = galsim.Image(np.array(out.data))
```
rmjarvis, line 55: Is there an appropriate pixel_scale to use here?  Maybe it never matters, but an Image typically has a scale, so worth considering if something makes sense in this case.  

Edit: I see dx, dy below, so maybe a jacobian WCS using dx,dy, which I suppose are allowed to be different.

tests/test_sag.py
```diff
@@ -0,0 +1,78 @@
...
+                    file_name: sag.fits
+                    nx: 127
+            """
+        )
+
+        config = yaml.safe_load(config)
+        galsim.config.ProcessInput(config)
+        galsim.config.SetupExtraOutput(config)
+        # Need to mock file_num for following method to be happy:
+        config['file_num'] = 0
```
rmjarvis, line 35: This works, but the canonical thing to call to setup file-related things is 
```
galsim.config.SetupConfigFileNum(config, 0, 0, 0)
```
which does a bit that could be necessary in some contexts.

rmjarvis (approved):
LGTM.  Just a few minor suggestions.

imsim/sag.py
jmeyers314: Typo.  Should have been `sag`

imsim/sag.py
jmeyers314: It's hard for me to imagine wanting different values for dx, dy, so I just made this a scale.

PR #375, opened 2023-07-12
photon_ops.ray_vector_to_photon_array: Rotate photon velocities from focal -> pixel

g-braeunlich:


g-braeunlich:
@jmeyers314 I am not sure about the scaling m -> mm in the velocity space. Maybe I have to remove this again

jmeyers314:
Hey Geri,
  I added a line to turn the jacobian for the velocities into a pure rotation.  I think that's correct.  We need -1 < dxdz < 1 for instance, and that was definitely not happening without this normalization.

jmeyers314 (approved):

PR #377, opened 2023-07-18
add vignetting named method

jchiang87:
* Enable multiprocessing in `config/imsim-skycat.yaml`
* Use `Vignetting.apply_to_radii` instead of `Vignetting.__call__` to avoid problems with proxy objects for multiprocessing
* Compute radii for pixels in a CCD using a static method to work around un-pickleable Detector objects


jmeyers314 (approved):
LGTM as long as CI passes.

Issue #376, opened 2023-07-12
Warn if atm_psf.doOpt is true and using RubinOptics


jmeyers314:
We don't want to simulate the optics twice!  This is a bit pernicious since the default for doOpt is true: https://github.com/LSSTDESC/imSim/blob/main/imsim/atmPSF.py#L111

rmjarvis:
+1  And probably should switch the default to False.  That feels like the wrong default behavior.

rmjarvis:
Done in #380

PR #380, opened 2023-07-20
Chromatic seeing

jmeyers314:
Adds fwhm \propto wave^-0.2 or so effect of chromatic seeing.  I ended up using galsim.ChromaticAtmosphere with the existing imsim AtmosphericPSF, though I had to split apart the parametric optics from the atm to make sure the chromaticity was only applied to the atm component.  I also split out the 2nd kick component of the atmosphere, since I'm pretty sure this shouldn't have the same wavelength dependence (in the case kcrit -> 0 or inf (can't remember which), the second kick becomes an Airy, which would have a wave^+1.0 chromaticity).

I also addressed #376 by making `doOpt` default to False and raising a warning when double counting the optics.

I added a workaround for https://github.com/GalSim-developers/GalSim/issues/1228 in stamp.py:fix_seds().

### DIFF
--- a/imsim/atmPSF.py
+++ b/imsim/atmPSF.py
@@ -220,7 +220,7 @@ class AtmosphericPSF(object):
         # Tokovinin 2002, PASP, v114, p1156
         # https://dx.doi.org/10.1086/342683
         kolm_seeing = galsim.Kolmogorov(r0_500=r0_500, lam=wavelength).fwhm
-        r0 = r0_500 * (wavelength/500)**1.2
+        r0 = r0_500 * (wavelength/500)**(6./5)
         arg = 1. - 2.183*(r0/L0)**0.356
         factor = np.sqrt(arg) if arg > 0.0 else 0.0
         return kolm_seeing*factor
@@ -232,7 +232,7 @@ class AtmosphericPSF(object):
     @staticmethod
     def _r0_500(wavelength, L0, targetSeeing):
         """Returns r0_500 to use to get target seeing."""
-        r0_500_max = min(1.0, L0*(1./2.183)**(-0.356)*(wavelength/500.)**1.2)
+        r0_500_max = min(1.0, L0*(1./2.183)**(-0.356)*(wavelength/500.)**(6./5))
         r0_500_min = 0.01
         return bisect(
             AtmosphericPSF._seeingResid,
--- a/imsim/atmPSF.py
+++ b/imsim/atmPSF.py
@@ -98,6 +98,11 @@ class AtmosphericPSF(object):
     @param screen_size  Size of the phase screens in meters.  default: 819.2
     @param screen_scale Size of phase screen "pixels" in meters.  default: 0.1
     @param doOpt        Add in optical phase screens?  default: False
+    @param exponent     Chromatic seeing exponent.  The atmospheric component of
+                        the PSF will dilate as (wavelength/500 nm)**exponent.  For
+                        Kolmogorov turbulence, this should be -0.2.  For Von Karman
+                        turbulence, this is somewhere around -0.3 (depends on the
+                        outer scale).  default: -0.3
     @param logger       Optional logger.  default: None
     @param nproc        Number of processes to use in creating screens. If None (default),
                         then allocate one process per phase screen, of which there are 6,
@@ -108,8 +113,8 @@ class AtmosphericPSF(object):
     """
     def __init__(self, airmass, rawSeeing, band, boresight, rng,
                  t0=0.0, exptime=30.0, kcrit=0.2,
-                 screen_size=819.2, screen_scale=0.1, doOpt=False, logger=None,
-                 nproc=None, save_file=None):
+                 screen_size=819.2, screen_scale=0.1, doOpt=False, exponent=-0.3,
+                 logger=None, nproc=None, save_file=None):
         self.airmass = airmass
         self.rawSeeing = rawSeeing
         self.boresight = boresight
@@ -125,13 +130,15 @@ class AtmosphericPSF(object):
         self.exptime = exptime
         self.screen_size = screen_size
         self.screen_scale = screen_scale
+        self.exponent = exponent
+        self.doOpt = doOpt
 
         if save_file and os.path.isfile(save_file):
             self.logger.warning(f'Reading atmospheric PSF from {save_file}')
             self.load_psf(save_file)
         else:
             self.logger.warning('Building atmospheric PSF')
-            self._build_atm(kcrit, doOpt, nproc)
+            self._build_atm(kcrit, nproc)
             if save_file:
                 self.logger.warning(f'Saving atmospheric PSF to {save_file}')
                 self.save_psf(save_file)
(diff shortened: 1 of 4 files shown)


imsim/atmPSF.py
```diff
@@ -196,7 +216,9 @@ def __eq__(self, rhs):
 
     @staticmethod
     def _vkSeeing(r0_500, wavelength, L0):
-        # von Karman profile FWHM from Tokovinin fitting formula
+        # von Karman profile FWHM from fitting formula in eqn 19 of
+        # Tokovinin 2002, PASP, v114, p1156
+        # https://dx.doi.org/10.1086/342683
         kolm_seeing = galsim.Kolmogorov(r0_500=r0_500, lam=wavelength).fwhm
         r0 = r0_500 * (wavelength/500)**1.2
```
rmjarvis, line 223: Is this right?  Not (`1+self.seeing`)?  Also above in `_build_atm` you use 6/5.

imsim/atmPSF.py
```diff
@@ -279,16 +301,39 @@ def getPSF(self, field_pos, gsparams=None):
...
+                    aper=self.aper,
+                    theta=theta,
+                    t0=self.t0,
+                    exptime=self.exptime,
+                    gsparams=gsparams,
+                    second_kick=False
+                ),
+                alpha=self.exponent,
+                base_wavelength=self.wlen_eff,
+                zenith_angle=0*galsim.degrees
```
rmjarvis, line 318: Probably a comment that this means no DCR, since we do DCR separately elsewhere.
But that raises the question, why not do DCR here now?

imsim/stamp.py
```diff
@@ -515,8 +515,13 @@ def fix_seds(prof):
                 #       Something like sed.make_tabulated()
                 if (not isinstance(sed._spec, galsim.LookupTable)
                     or sed._spec.interpolant != 'linear'):
-                    new_spec = galsim.LookupTable(wave_list, sed(wave_list), interpolant='linear')
-                    new_sed = galsim.SED(new_spec, 'nm', 'fphotons')
+                    f = np.broadcast_to(sed(wave_list), wave_list.shape)
```
rmjarvis, line 518: Maybe mention here the GalSim issue for which this is a workaround.

tests/test_psf.py
```diff
@@ -255,5 +255,99 @@ def test_r0_500(self):
...
+            config400['input']['atm_psf']['exponent'] = exponent
+            config400['gal']['sed'] = sed400
+            img400 = galsim.config.BuildImage(config400)
+            sigma400 = galsim.hsm.FindAdaptiveMom(img400).moments_sigma
+
+            config900 = copy.deepcopy(template)
+            config900['input']['atm_psf']['exponent'] = exponent
+            config900['gal']['sed'] = sed900
+            img900 = galsim.config.BuildImage(config900)
+            img900.wcs = galsim.PixelScale(0.05) # just lie a little bit
```
rmjarvis, line 342: why?

imsim/atmPSF.py
```diff
@@ -125,13 +132,17 @@ def __init__(self, airmass, rawSeeing, band, boresight, rng,
...
+        self.opt = None
 
         if save_file and os.path.isfile(save_file):
             self.logger.warning(f'Reading atmospheric PSF from {save_file}')
             self.load_psf(save_file)
         else:
             self.logger.warning('Building atmospheric PSF')
             self._build_atm(kcrit, doOpt, nproc)
+            if _no2k:
+                self.second_kick = galsim.DeltaFunction()
```
rmjarvis, line 145: This feels like the wrong place for this.  I think better to put it in _build_atm, and pass another argument do2k, like we have doOpt.

tests/test_psf.py
```diff
@@ -255,5 +255,99 @@ def test_r0_500(self):
...
+            sigma900 = galsim.hsm.FindAdaptiveMom(img900).moments_sigma
+
+            np.testing.assert_allclose(
+                sigma400/sigma900,
+                (400/900)**exponent,
+                rtol=0.02  # 2% error on value of around ~1.2 to 1.3
```
rmjarvis, line 348: I think I would have expected somewhat better than 2% here.  Does it improve with more photons?  I.e. is this just shot noise or a sign of something else going on?

rmjarvis (changes requested):

imsim/atmPSF.py
jmeyers314: I'm lost, where do you suggest I add `1+self.seeing`?  Happy to change 6/5 -> 1.2.

This bit is implementing the following:
Screen Shot 2023-07-20 at 11 57 03 AM

where \epsilon_0 is the fwhm from a Kolmogorov profile and \epsilon_vK is with the outer scale added in (a VonKarman profile).

rmjarvis: OK.  I guess I just assumed the 1.2 was related to the 0.2 of Kolmogorov seeing.  If not, ignore me.

imsim/atmPSF.py
jmeyers314: > why not do DCR here now?

We probably could, but would need to parse and validate zenith/parallactic angles.  Since that's already working with the PhotonDCR operator, I'm inclined to just keep DCR there.  I do think there's a case for not applying any DCR if no zenith/parallactic/etc angles are passed to ChromaticAtmosphere in GalSim, but it currently raises if you don't pass something like zenith=0.

tests/test_psf.py
jmeyers314: Good catch.  That was from an earlier iteration where I was using image.calculateMomentRadius() instead of hsm.  I'll remove it.

tests/test_psf.py
jmeyers314: At the very least, the pixel convolution is there.  I did try going to more photons, didn't seem to help much.  I can try shrinking the pixel size some more...  I'm not sure what kind of systematic uncertainties to expect from hsm too.

imsim/atmPSF.py
jmeyers314: It's just the normal 6/5.  I guess 6/5 might be a bit clearer.

tests/test_psf.py
rmjarvis: Ah, the pixel convolution would definitely impact this.  That's probably the main thing.

jmeyers314: Okay.  Got it down to 0.1% using 4x smaller pixels.

jmeyers314:
I think I've hit everything, @rmjarvis .  Let me know if you have any new/more comments.

tests/test_psf.py
rmjarvis: Great.  I'm satisfied that that's the reason it's not perfect then.

imsim/atmPSF.py
```diff
@@ -279,16 +301,40 @@ def getPSF(self, field_pos, gsparams=None):
...
+                    aper=self.aper,
+                    theta=theta,
+                    t0=self.t0,
+                    exptime=self.exptime,
+                    gsparams=gsparams,
+                    second_kick=False
+                ),
+                alpha=self.exponent,
+                base_wavelength=self.wlen_eff,
+                zenith_angle=0*galsim.degrees
```
rmjarvis, line 318:
```suggestion
                zenith_angle=0*galsim.degrees  # Turns off DCR, since we apply that later using PhotonDCR.
```

rmjarvis (approved):

PR #387, opened 2023-07-21
Parse LSSTSiliconBuilder Options in stamp.py

cwwalter:
Add option parsing to the LSST_SiliconBuilder class.

Note that unlike what I reported, the

OpsimDataLoader
TreeRings
Checkpointer

classes do not need to be changed. They are
using the builtin kwargs function from the InputLoader class.

### DIFF
--- a/imsim/stamp.py
+++ b/imsim/stamp.py
@@ -88,15 +88,19 @@ class LSST_SiliconBuilder(StampBuilder):
         # First do a parsing check to make sure that all the options passed to
         # the config are valid and no required options are missing.
 
-        req = {'fft_sb_threshold': float,
+        req = {'fft_sb_thresh': float,
                'airmass': float,
-               'rawseeing': float,
-               'band': str
+               'rawSeeing': float,
+               'band': str,
+               'world_pos': galsim.CelestialCoord,
+               'diffraction_psf': dict
                }
 
         opt = {'max_flux_simple': float,
-               'method': str,
-               'maxN': int
+               'draw_method': str,
+               'maxN': int,
+               'size': int,
+               'photon_ops': dict
                }
 
         galsim.config.CheckAllParams(config, req, opt)
--- a/imsim/stamp.py
+++ b/imsim/stamp.py
@@ -84,6 +84,23 @@ class LSST_SiliconBuilder(StampBuilder):
         Returns:
             xsize, ysize, image_pos, world_pos
         """
+
+        # First do a parsing check to make sure that all the options passed to
+        # the config are valid and no required options are missing.
+
+        req = {'fft_sb_threshold': float,
+               'airmass': float,
+               'rawseeing': float,
+               'band': str
+               }
+
+        opt = {'max_flux_simple': float,
+               'method': str,
+               'maxN': int
+               }
+
+        galsim.config.CheckAllParams(config, req, opt)
+
         gal = galsim.config.BuildGSObject(base, 'gal', logger=logger)[0]
         if gal is None:
             raise galsim.config.SkipThisObject('gal is None (invalid parameters)')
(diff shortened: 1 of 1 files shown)


imsim/stamp.py
```diff
@@ -84,6 +84,27 @@ def setup(self, config, base, xsize, ysize, ignore, logger):
         Returns:
             xsize, ysize, image_pos, world_pos
         """
+
+        # First do a parsing check to make sure that all the options passed to
+        # the config are valid, and no required options are missing.
+
+        req = {'fft_sb_thresh': float,
```
rmjarvis, line 91: This is optional.

imsim/stamp.py
```diff
@@ -84,6 +84,27 @@ def setup(self, config, base, xsize, ysize, ignore, logger):
...
+               'band': str,
+               'world_pos': galsim.CelestialCoord,
```
rmjarvis, line 93: Not required.  (And probably not worth mentioning in opt either.  This is usually automatically computed by GalSim.)

imsim/stamp.py
```diff
@@ -84,6 +84,27 @@ def setup(self, config, base, xsize, ysize, ignore, logger):
...
+               'diffraction_psf': dict
+               }
+
+        opt = {'airmass': dict,
+               'rawSeeing': float,
+               'max_flux_simple': float,
+               'draw_method': str,
+               'maxN': int,
+               'size': int,
+               'photon_ops': dict
```
rmjarvis, line 103: Some of these are in ignore, so you don't need to include them (They are standard GalSim optional items, which GalSim has already put into ignore here.): draw_method, maxN, size, photon_ops, and world_pos (from above).

rmjarvis (changes requested):

imsim/stamp.py
```diff
@@ -84,6 +84,27 @@ def setup(self, config, base, xsize, ysize, ignore, logger):
         Returns:
             xsize, ysize, image_pos, world_pos
         """
+
+        # First do a parsing check to make sure that all the options passed to
+        # the config are valid, and no required options are missing.
+
+        req = {'fft_sb_thresh': float,
+               'band': str,
```
rmjarvis, line 92: This is optional (sort of).  It's only required for the same code path that hits rawSeeing and airmass.  I.e. simulating moderately bright stars (flux>10) if stamp.size is not given explicitly.

imsim/stamp.py
```diff
@@ -84,6 +84,27 @@ def setup(self, config, base, xsize, ysize, ignore, logger):
...
+        opt = {'airmass': dict,
+               'rawSeeing': float,
+               'max_flux_simple': float,
+               'draw_method': str,
+               'maxN': int,
+               'size': int,
+               'photon_ops': dict
+               }
+
+        galsim.config.CheckAllParams(config, req, opt)
```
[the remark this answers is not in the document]
rmjarvis: Since I think we've converted everything to optional, you can more simply just call
```
ignore = ['fft_sb_thresh', 'band', 'airmass', 'rawSeeing', 'max_flux_simple'] + ignore
galsim.config.CheckAllParams(config, ignore=ignore)
```
If only calling Check, rather than Get, things in ignore are effectively treated the same as things in opt.

imsim/stamp.py
cwwalter: So, I added that because I was getting:

```
galsim.errors.GalSimConfigError: Unexpected attribute world_pos found
```

as it is in our imsim-config.yaml.  

I can add it to the opt list.

cwwalter: Or perhaps it is already in a global ignore list and the option to the call would fix it?

cwwalter: Ah... yes, that fixed it. Thanks.

imsim/stamp.py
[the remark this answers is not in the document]
cwwalter: I think we still currently need the diffraction_psf as required.

cwwalter:
OK this works.

I'm not really sure why we would have 'fft_sb_thresh' and 'max_flux_simple' in ignore rather than optional here though.

rmjarvis (approved):

PR #388, opened 2023-07-21
Changes in response to PR which was closed-early.

cwwalter:
Changes suggested by Mike. These changes were actually suggested in PR #378, but it was closed before his review was visible.


cwwalter:
Note that this PR removes the imsim-user.yaml in config, and keeps the one in example. the install and usage instructions use the version in example for non-user config templates etc only.

doc/config.rst
```diff
@@ -507,130 +517,138 @@ Stamp Type: LSST_Silicon
...
-For reading out LSST CCDs.
+For modifying the ``LSST_Image`` by adding cosmic rays and relevant headers and optionally writing out the "eimage" FITS files via the standard *GalSim* output.
 
 Optional keywords to set:
 """"""""""""""""""""""""""
 
     * ``cosmic_ray_rate`` = *float_value* (default = 0) The rate of cosmic rays per second in a sensor.
     * ``cosmic_ray_catalog`` = *str_value* (default = Distributed with *imSim*) A file containing cosmic ray images to paint on the sensor.
-    * ``header`` = *dictionary* (default = None) Extra items to add to the FITS header in the output files.
+    * ``header`` = *dict* (optional) Extra items to add to the FITS header in the output files.
```
rmjarvis, line 639: Should add here a few more, which aren't so obvious in the code, since we just use the standard GalSIm processing to handle them:

`nproc`: How many processes to use for working on multiple CCDs at a time.
`nfiles`: How many CCDs to build.  ("files" here refers to the eimage files that are produced by the standard processing)
`file_name`: The name of the eimage file to write out.
`dir`: The directory to write the output files to (default is the current directory)

Then there are also a couple imSim parameters you missed:

`camera` (default = 'LsstCam'), `exptime` (default=30), `det_num` (I think required currently, although the config files says it defaults to all 189, but I don't see that implemented anywhere).

doc/config.rst
```diff
@@ -507,130 +517,138 @@ Stamp Type: LSST_Silicon
...
 
-Parameters modifying the LSSTCamera electronics readout.
+Parameters modifying the LSSTCamera electronics readout and writing the amplifier files.
 
 Optional keywords to set:
 """"""""""""""""""""""""""
 
-    * ``camera_name`` = *str_value* (default = None) The camera object to use.
+    * ``file_name`` = *str_value* (default = None) The file name for amplifier file output(s).
+    * ``camera_name`` = *str_value* (default = 'LsstCam') The name of tthe camera to use.
```
rmjarvis, line 651: This is right, but we should switch it to `camera` to match everything else that takes a camera name.

rmjarvis (changes requested):
A few more suggestions...

rmjarvis (approved):

rmjarvis:
Chris, I went ahead and made the edits I suggested.  I think this is good to go now.

cwwalter:
Thanks a lot  Mike!

I want to make one more change before merging to make the examples file use the IMSIM_HOME env variable again that you removed. 

I want to encourage people to run in a working directory (instead of the imSim directory itself) and make sure this can run from any directory successfully. The install test instructions have people doing this.

 I will double check to make sure the new check you added in the CI still works.

rmjarvis:
OK.  I'll just register my objection to using IMSIM_HOME then.  I don't like that style at all.  But you do you.

PR #392, opened 2023-07-25
Fix various aspects of the make_fft_psf code.

rmjarvis:
This PR fixes a number of things.

1. First, the PSF we use for FFT drawing needs to be achromatic, so the new ChromaticAtmosphere PSF needs to be evaluated at the bandpass effective wavelength before we do further processing in make_fft_psf.
2. The unit test I wrote to test this discovered that our modified FFT PSF (to make sure not to use the atmospheric screens in FFT mode, which is super slow) doesn't end up with a good match to the one with screens.  There were a few things to change to make it work better:
    a. Get rid of the SecondKick, which we don't need.
    b. Add back in an Airy, which is implicitly part of a SecondKick piece.
    c. Use VonKarman with the L0 from the screen, rather than Kolmogorov
3. The LSST_Silicon stamp type was implicitly dependent on there being an output field using LSST_CCD, which we don't want.  All it needed from that was the det_name.  So now it's a required parameter in LSST_Silicon, where the template config gets that from the output field.
4. It also got the camera from the output field, which again we don't want.  So that's now an optional parameter as per how we set this elsewhere in imsim.

I also added a run of the new examples/imsim-user.yaml to our CI tests to make sure that runs, which is what triggered this bug fix.  Now CI should notice when that breaks.

### DIFF
--- a/imsim/stamp.py
+++ b/imsim/stamp.py
@@ -444,17 +444,26 @@ class LSST_SiliconBuilder(StampBuilder):
         elif isinstance(psf, galsim.Convolution):
             obj_list = [self.make_fft_psf(p, logger) for p in psf.obj_list]
             return galsim.Convolution(obj_list, gsparams=psf.gsparams)
+        elif isinstance(psf, galsim.SecondKick):
+            # The Kolmogorov version of the phase screen gets most of the second kick.
+            # The only bit that it missing is the Airy part, so convert the SecondKick to that.
+            return galsim.Airy(lam=psf.lam, diam=psf.diam, obscuration=psf.obscuration)
         elif isinstance(psf, galsim.PhaseScreenPSF):
             # If psf is a PhaseScreenPSF, then make a simpler one the just convolves
             # a Kolmogorov profile with an OpticalPSF.
             r0_500 = psf.screen_list.r0_500_effective
-            atm_psf = galsim.Kolmogorov(lam=psf.lam, r0_500=r0_500,
-                                        gsparams=psf.gsparams)
+            L0 = psf.screen_list[0].L0
+            atm_psf = galsim.VonKarman(lam=psf.lam, r0_500=r0_500, L0=L0, gsparams=psf.gsparams)
 
             opt_screens = [s for s in psf.screen_list if isinstance(s, galsim.OpticalScreen)]
             logger.info('opt_screens = %r',opt_screens)
             if len(opt_screens) >= 1:
                 # Should never be more than 1, but if there weirdly is, just use the first.
+                # Note: Technically, if you have both a SecondKick and an optical screen, this
+                # will add the Airy part twice, since it's also part of the OpticalPSF.
+                # It doesn't usually matter, since we usually set doOpt=False, so we don't usually
+                # do this branch. If it is found to matter for someone, it will require a bit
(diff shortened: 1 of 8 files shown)


tests/test_psf.py
```diff
@@ -347,6 +349,117 @@ def test_chromatic_psf(self):
...
+        print('Rmom of FFT PSF: ',img.view(scale=0.2).calculateMomentRadius())
+
+        # The FFT image is about 10^3 x brighter than the reference image.
+        # Scale it down to make it easier to compare to the reference image.
+        img /= 1.e3
+
+        # Peaks should now be similar
+        np.testing.assert_allclose(ref_img.array.max(), img.array.max(), rtol=0.05)
+
+        # The sizes should also be pretty close
```
jmeyers314, line 453: Looks like CI failed here pretty dramatically.  Do we still expect the RNGs to be close to platform independent?  Other than this, PR all looks okay to me.

jmeyers314 (approved):

tests/test_psf.py
rmjarvis: Weird.  I certainly expected the random numbers to be portable.  We even have unit tests in GalSim to verify that the same random values are produced on all systems.  Hm...  I'll try running on some other systems besides my laptop to see if I can see what is going on with that.

rmjarvis: I'm a little stymied by this.  I can't reproduce the failure on any system I've tried.  USDF, Perlmutter, and BNL all give exactly the same numbers as my laptop, and thus they all pass.  I have no idea why GitHub CI is getting different values.

rmjarvis: OK, I more or less figured it out.  It turns out that on GitHub, the image position was ending up as galsim.PositionD(5.636331611506773e-11,-2.8624992133171645e-11) rather than (0,0).  This ended up causing a different centering of the stamp, which would be fine for normal things where it gets dropped onto a larger image in the right place.  But in my test, my size estimators were centered on the image center, which wasn't the center of the object. 
Anyway, this led me to adjust the stamp builder to allow users to input image_pos, rather than world_pos, so I can give (0,0) explicitly, rather than have the wcs calculate it.  This is something we should really have had already to better match how the default stamp builder works, so that's a bonus.

imsim/stamp.py
```diff
@@ -203,10 +205,17 @@ def setup(self, config, base, xsize, ysize, ignore, logger):
 
         logger.info('Object %d will use stamp size = %s',base.get('obj_num',0),image_size)
 
-        # Also the position
-        world_pos = galsim.config.ParseWorldPos(config, 'world_pos', base, logger)
-        self.sky_coord = world_pos
-        image_pos = None  # GalSim will figure this out from the wcs.
+        # Determine where this object is going to go:
+        # This is the same as what the base StampBuilder does:
```
jmeyers314, line 209: All looks good again to me.

Just curious, what happens if both image_pos and world_pos are present in the config, but are mutually inconsistent with the WCS?

rmjarvis: IIRC, you can do this, and it will respect your values of both, ignoring the absolute calibration of the wcs.  The Jacobian of the wcs would still be used where appropriate, but not used to place the objects on the stamp.

PR #394, opened 2023-08-01
bug-fix for bright objects without saturated pixels

jchiang87:

PR #395, opened 2023-08-01
bright galaxy stamp fix

jchiang87:
Here are three renderings of the galaxy that I mentioned in [this slack thread](https://lsstc.slack.com/archives/CHXKSF3HC/p1673543975668059) at different stages of addressing this issue.  The first image on the left uses the current code in main.  Here the main issue is that the stamp size returned by `.getGoodImageSize` ([here](https://github.com/LSSTDESC/imSim/blob/c42c10884d085b0f82062f4760f3aa237c6a7dac/imsim/stamp.py#L182)) is so large that the stamp has fewer than 10 photons per pixel on average, so that [this line](https://github.com/LSSTDESC/imSim/blob/c42c10884d085b0f82062f4760f3aa237c6a7dac/imsim/stamp.py#L196) isn't executed and the [somewhat brighter sb limit](https://github.com/LSSTDESC/imSim/blob/c42c10884d085b0f82062f4760f3aa237c6a7dac/imsim/stamp.py#L202) is used instead.  After addressing that by testing the `image_size` vs `self._Nmax`, the middle image below is the outcome.  So I lowered the nominal sb limit used by `._getGoodPhotImageSize` from `np.sqrt(noise_var)/3.` to `np.sqrt(noise_var)/8.`.   With that change, the image on the right is the result.

bug-fix_stages

Note that this fix also needs #394 to work properly for objects like this.

### DIFF
--- a/imsim/stamp.py
+++ b/imsim/stamp.py
@@ -190,9 +190,11 @@ class LSST_SiliconBuilder(StampBuilder):
 
             gal_at_eff_wl = gal.evaluateAtWavelength(bandpass.effective_wavelength)
             # For bright things, defined as having an average of at least 10 photons per
-            # pixel on average, try to be careful about not truncating the surface brightness
+            # pixel on average, or objects for which GalSim's estimate of the image_size is larger
+            # than self._Nmax, compute the image_size using the surface brightness limit, trying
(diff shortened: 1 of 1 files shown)


rmjarvis (approved):
lgtm

PR #396, opened 2023-08-02
use skyCatalogs 1.5.0-rc2

jchiang87:


rmjarvis (approved):

rmjarvis:
Oops. I approved based on looking at the code, but it seems we need to define a new environment variable now.  Unit tests are failing with `KeyError: 'SKYCATALOGS_DIR'`.

PR #398, opened 2023-08-03
use camera model saturation

jchiang87:


config/imsim-config.yaml
```diff
@@ -66,6 +66,10 @@ input:
         visit: 19500
         snap: 1
 
+    camera_data:
```
cwwalter, line 69: Sadly, the changes in this file are also going to need to go into the two new template files with the copy.

(can handle after that PR merges to main..  just noting now)

imsim/camera.py
```diff
@@ -160,14 +181,31 @@ class Camera(dict):
     Class to represent the LSST Camera as a dictionary of CCD objects,
     keyed by the CCD name in the focal plane, e.g., 'R01_S00'.
     """
-    def __init__(self, camera_class='LsstCam'):
+    _req_params = {'camera': str}
```
cwwalter, line 184: I think these aren't being used/checked here.  Maybe they belong with the others in the CameraReadout class?

imsim/camera.py
```diff
@@ -183,3 +221,21 @@ def update(self, other):
...
+    req = {'det_name': str}
+
+    kwargs, safe = GetAllParams(config, base, req=req)
+    det_name = kwargs['det_name']
+
+    value = camera_object[det_name].full_well
+    return value, False
+
+
+RegisterInputType('camera_data', InputLoader(Camera, takes_logger=False))
```
cwwalter, line 239: I think the parameters for this new camera_data type aren't being checked anywhere with a Get or Check parameters call.

imsim/readout.py
```diff
@@ -503,6 +522,7 @@ def finalize(self, config, base, main_data, logger):
             'pcti': float,
             'full_well': float,
             'read_noise': float,
+            'bias_levels_file': str,
```
cwwalter, line 525: This is also being checked here (as marked as a question in other file).

imsim/stamp.py
```diff
@@ -16,8 +16,8 @@ class DiffractionPSF:
     azimuth: galsim.Angle
     altitude: galsim.Angle
     rotTelPos: galsim.Angle
+    brightness_threshold: float
```
cwwalter, line 19: And because of this works with the MISSING below this parameter is being checked here.

tests/test_camera_readout.py
```diff
@@ -128,11 +128,12 @@ def test_no_opsim(self):
         ccd = imsim.Camera(camera_name)[det_name]
         for amp in range(16):
             read_noise = list(ccd.values())[amp].read_noise
+            gain = list(ccd.values())[amp].gain
             amp_mean = np.mean(readout.final_data[amp+1].data)
             amp_var = np.var(readout.final_data[amp+1].data)
             assert np.isclose(amp_mean, 1000 + 0.02*dark_time, rtol=1.e-3)
-            print(amp_var, 0.02*dark_time + read_noise**2, (0.02*dark_time + read_noise**2-amp_var)/amp_var)
-            assert np.isclose(amp_var, 0.02*dark_time + read_noise**2, rtol=3.e-3)
+            print(amp_var, 0.02*dark_time/gain + read_noise**2, (0.02*dark_time/gain + read_noise**2-amp_var)/amp_var)
```
cwwalter, line 135: Are we switching ow from ADUs to electrons here (now dividing by gain)?

cwwalter (approved):
I can't find anything obviously wrong in the new code.  Here are some comments/questions.

cwwalter:
Also, once the checking of things is worked out.. it would be good to add the new parameters to the appropriate section in doc/config.rst around the section "Extra Output Type: readout".

imsim/camera.py
jchiang87: I think they must be used inside of the galsim code when it provides the `CcdSaturation` value to the diffraction_psf object.

imsim/ccd.py
```diff
@@ -169,6 +169,7 @@ def parse(item, type, default):
         image.header['DAYOBS'] = dayobs
         image.header['SEQNUM'] = seqnum
         image.header['CONTRLLR'] = 'P', 'simulated data'
+        image.header['RUNNUM'] = parse('observationId', int, -999)
```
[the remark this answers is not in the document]
jchiang87: The -999 is just a default value.  For opsim db input, `observationId` is provided from the `observations` table.  Similarly, in an instcat file, the `obshistid` entry provides this value.   It may be that the default value here is never used, but the `parse` interface requires that argument to be provided.

tests/test_camera_readout.py
jchiang87: It's the other way around.  Dividing by the gain here converts from electrons to ADU.  This was always needed, but the gain values in the earlier weeklies of obs_lsst were sufficiently close to unity, that the test still passed.  The current gain values are now 1.5-1.7.

imsim/camera.py
rmjarvis: @cwwalter The `_req_params` and `_opt_params` class variables mean that they get checked.

imsim/stamp.py
rmjarvis: Why is this required now?  Can't we give this the usual default of 1.e5 still?

imsim/readout.py
```diff
@@ -206,15 +206,14 @@ def cte_matrix(npix, cti, ntransfers=20):
     return my_matrix
 
 
-# TODO: get lsst_num from camera object.
-def get_primary_hdu(eimage, lsst_num='LCA-11021_RTM-000', camera_name=None,
+def get_primary_hdu(eimage, lsst_num, camera_name=None,
```
rmjarvis, line 209: Is lsst_num really the right name for this parameter?  It doesn't look like a number.  Maybe `lsst_serial` is more apt, since the value comes from `ccd.getSerial()`.

imsim/camera.py
```diff
@@ -160,14 +181,31 @@ class Camera(dict):
...
+            if not os.path.isfile(bias_levels_file):
+                bias_levels_file = os.path.join(data_dir, bias_levels_file)
+                if not os.path.isfile(bias_levels_file):
+                    raise FileNotFoundError(
+                        f"{bias_levels_file} not found.")
+            with open(bias_levels_file) as fobj:
+                bias_levels = json.load(fobj)
+        else:
+            # Set the input dict of per-amp bias levels to None.
+            bias_levels = defaultdict(lambda : None)
```
rmjarvis, line 203: There don't seem to be any tests for most of these new code paths.  I think all the unit tests still only test the else branch here, which doesn't use the bias_levels_file.  You should add something that tests the thing the main config file does.  And also one where the file name is fully specified (rather than implicitly in data_dir).  And ideally one that checks the OSError when it's not present.

imsim/camera.py
```diff
@@ -183,3 +221,21 @@ def update(self, other):
...
+    """Retrieve detector data from the Camera object."""
+    camera_object = GetInputObj('camera_data', config, base, 'CcdSaturation')
+
+    req = {'det_name': str}
+
+    kwargs, safe = GetAllParams(config, base, req=req)
+    det_name = kwargs['det_name']
+
+    value = camera_object[det_name].full_well
+    return value, False
```
rmjarvis, line 236: Likewise, I don't think you have any unit tests for this class.

rmjarvis (changes requested):
My main request is that we only have one place in the config file with the new bias_levels_file, rather than 2.  Otherwise there is significant potential for users to let these get out of sync.

imsim/readout.py
```diff
@@ -282,43 +281,55 @@ class CcdReadout:
...
     per image HDU, simulating the electronics readout effects for each amp.
     """
-    full_well: float = 1.0e5
-    
-    def __init__(self, eimage, logger, camera=None,
+    def __init__(self, eimage, logger, camera_name=None,
                  readout_time=2.0, dark_current=0.02, bias_level=1000.0,
-                 scti=1.0e-6, pcti=1.0e-6, full_well=full_well, read_noise=None):
+                 scti=1.0e-6, pcti=1.0e-6, full_well=None, read_noise=None,
+                 bias_levels_file=None):
```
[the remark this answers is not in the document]
jchiang87: I couldn't figure out how to do exactly this using the config system without substantially rewriting both the CcdReadout and Camera code.  They probably should be rewritten, but I wasn't going to be able to do it before PCW.

imsim/readout.py
jchiang87: The project header keyword name for this value is `LSST_NUM`, hence the name used here.   We can use `lsst_serial` if you want, but that's the reason `lsst_num` was originally selected.

imsim/readout.py
```diff
@@ -282,43 +281,55 @@ class CcdReadout:
...
         """
         Parameters
         ----------
         eimage : GalSim.Image
             The eimage with the rendered scene, wcs, and header information.
         logger : logging.Logger
             Logger object
-        camera : str, Camera class to use, e.g., 'LsstCam', 'LsstCamImSim'.
+        camera_name : str, Camera class to use, e.g., 'LsstCam', 'LsstCamImSim'.
```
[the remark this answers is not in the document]
jchiang87: Thanks for explaining the constraints.  They are not at all clear, to me at least.

jchiang87:
> My main request is that we only have one place in the config file with the new bias_levels_file, rather than 2. Otherwise there is significant potential for users to let these get out of sync.

I wasn't happy about it either, but it's the only way I could see to get the saturation info to the DiffractionPsf class without rewriting a lot of code. If you can suggest a concrete set of changes to do this, I'd be happy to adopt them.

imsim/stamp.py
jchiang87: One of the reasons for these changes was to make use of the per CCD values obtained from the camera the default behavior, both here and in the readout code.  Letting this be 1e5 for all CCDs would be defeat that aim.

jchiang87: Thinking about this some more, it occurs to me that linking this brightness threshold for the diffraction spikes to the CCD saturation values really doesn't make any sense.  They are physically unrelated.  So we can just use set this to some reasonable value, and I can remove all of the awkward config-related code here that was intended to preserve that coupling.

imsim/camera.py
cwwalter: Ah from `get_cls_params` in the inputLoader. Thanks.

jchiang87:
I'm closing this and will make a new branch and PR since there's a lot that would need to be cleaned up here.

Issue #384, opened 2023-07-20
Make diffraction_psf optional


rmjarvis:
Currently the diffraction_psf bit is a required keyword in LSST_Silicon, but you can set enabled=False.  This is an awkward thing to require for people who don't want to bother with it.  We should make it so omitting the keyword entirely is the same as setting enabled=False.

PR #401, opened 2023-08-07
Make most stamp parameters optional

rmjarvis:
There are a number of parameters to our stamp type, `LSST_Silicon`, which are currently required, but are really only needed for specific use cases.

1. diffraction_psf is only appropriate if one is using one of our diffraction photon ops.  Making it optional was requested in issue #384.
2. airmass, band, rawSeeing are only needed to calculate an appropriate image size for bright stars.  And there are already reasonable defaults if the user doesn't care about this being super precise.
3. det_num is only needed if one is doing vignetting.

These are now all optional, except det_num, which is required if vignetting is enabled.

Also, I think the name diffraction_psf is confusing here, since it doesn't make clear that this only applied to FFT-rendered objects.  So I switched the name to diffraction_fft.  I also added a not in the docs that point out that it should be enabled if one is using one of our diffraction-enabling photon ops.

### DIFF
--- a/doc/config.rst
+++ b/doc/config.rst
@@ -519,6 +519,12 @@ Required keywords to set:
 
     * ``det_name`` = *str_value* (only required if doing vignetting) The name of the detector.
 
+      .. note::
+            If using the output type LSST_CCD, then ``det_name`` will automatically be added
+            to the ``eval_variables`` section for you.  In this case, you can simply use
+            ``det_name: '$det_name'``.  If not using LSST_CCD, then the value should be of a
+            form such as R22_S11.  (This is the central CCD in the focal plane.)
+
 Optional keywords to set:
 """""""""""""""""""""""""
     * ``fft_sb_threshold`` = *float_value* (default = 0) Over this number of counts, use a FFT instead of photon shooting for speed.  If set to 0 don't ever switch to FFT.
--- a/doc/config.rst
+++ b/doc/config.rst
@@ -289,7 +289,8 @@ Key Name: atmosphericPSF
 This keyword enables an atmospheric PSF with 6 randomly generated atmospheric screens.  Photons are raytraced through this atmosphere to produce a realistic atmospheric PSF.
 
 .. warning::
-    You should not attempt to use the option to add parametric optics (through the ``doOpt`` option) if you are using fully ray-traced optics.  Otherwise, you will simulate the optics twice.  See See :ref:`the stamp keyword <stamp-label>` below how to activate the ray-traced mode.
+
+    You should not attempt to use the option to add parametric optics (through the ``doOpt`` option) if you are using fully ray-traced optics.  Otherwise, you will simulate the optics twice.  See :ref:`the stamp keyword <stamp-label>` below how to activate the ray-traced mode.
 
 
 Required keywords to set:
@@ -531,9 +532,8 @@ Optional keywords to set:
     * ``camera`` = *str_value* (default = 'LsstCam') The name of the camera to use.
 
 
-Note there is an extra required diffraction_psf keyword you must include in the stamp section above that configures how diffraction passing through the telescope spiders is handled.  Here is how to configure it.
-
-Key Name: diffraction_psf:
+Note there is an extra required diffraction_fft keyword you must include in the stamp section above that configures how diffraction passing through the telescope spiders is handled.  Here is how to configure it.
+Key Name: diffraction_fft:
 ^^^^^^^^^^^^^^^^^^^^^^^^^^
 
 Required keywords to set:
@@ -578,7 +578,11 @@ Optional keywords to set:
 type: **RubinDiffraction**
 ^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-Apply diffractive Effects.
+Apply diffractive effects.
+
+.. warning::
+
+    This only applies to objects rendering using photon shooting.  To be consistent, if you use this, you should also set diffraction_fft as described above.)
 
 Required keywords to set:
 """""""""""""""""""""""""
--- a/doc/config.rst
+++ b/doc/config.rst
@@ -517,11 +517,7 @@ Stamp Type: LSST_Silicon
 Required keywords to set:
 """""""""""""""""""""""""
 
-    * ``airmass`` = *float_value* (default = 1.2) The airmass to use in FFTs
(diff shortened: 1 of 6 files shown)


doc/config.rst
```diff
@@ -516,11 +517,7 @@ Stamp Type: LSST_Silicon
 Required keywords to set:
 """""""""""""""""""""""""
 
-    * ``airmass`` = *float_value* (default = 1.2) The airmass to use in FFTs
-    * ``rawseeing`` = *float_value* (default = 0.7) The FWHM seeing at zenith at 500 nm in arc seconds for FFTs.
-    * ``band`` = *str_value* (default = None) The filter band of the observation.
-    * ``det_name`` = *str_value* The name of the detector.
-
+    * ``det_name`` = *str_value* (only required if doing vignetting) The name of the detector.
```
cwwalter, line 520: I think this is a bit confusing as stands.

det_name is special in that it currently the only(?) variable which is set for use in the imSim codebase with:

base['eval_variables']['sdet_name'] = det_name

in ccd.py. So in out config file we set this to:

det_name: "$det_name"  # This is automatically defined by the LSST_CCD 

I think maybe the documentation here might be more explicit, but I also wonder if we shouldn't make this assignment by default.  Are there cases where someone sets the CCD they want and then overrides this to be another sensor? I see in the test it is not set because we are not doing vingnetting, but that seems a slightly odd thing to have to think about setting.

Finally, I realized we never give an example of what this string might look like (e.g R22_S11). Some people might think that this means e2V or ITL or a dector number.

imsim/stamp.py
```diff
@@ -88,32 +87,37 @@ def setup(self, config, base, xsize, ysize, ignore, logger):
...
         self.gal = gal
 
-        self.diffraction_psf = DiffractionPSF.from_config(config, base)
-
-        try:
-            self.vignetting = GetInputObj('vignetting', config, base,
-                                          'LSST_SiliconBuilder')
-        except galsim.config.GalSimConfigError:
-            self.vignetting = None
+        self.diffraction_fft = DiffractionFFT.from_config(config, base)
```
cwwalter, line 112: I went through this, and I think it is correct.  But I found it quite confusing when first reading to see the checks above and then a 2nd check using the dataclass here. 

That's because this diffraction_fft object is different, with it's own options and isn't actually used until the draw() function later. 

I think some comments saying something about also check the difraction_fft options for use later in fcn() might be useful.

imsim/stamp.py
```diff
@@ -158,13 +162,11 @@ def setup(self, config, base, xsize, ysize, ignore, logger):
...
                 gsparams = galsim.GSParams(folding_threshold=folding_threshold)
 
-            airmass = galsim.config.ParseValue(config, 'airmass', base, float)[0]
-            rawSeeing = galsim.config.ParseValue(config, 'rawSeeing', base, float)[0]
-            band = galsim.config.ParseValue(config, 'band', base, str)[0]
-            psf = self.Kolmogorov_and_Gaussian_PSF(gsparams=gsparams,
-                                                   airmass=airmass,
-                                                   rawSeeing=rawSeeing,
-                                                   band=band)
+            opt = { 'airmass': float, 'rawSeeing': float, 'band': str }
```
cwwalter, line 165: Could we do these checks with the others above?

imsim/stamp.py
```diff
@@ -59,7 +58,7 @@ class LSST_SiliconBuilder(StampBuilder):
                               wave_type='nm', flux_type='fphotons')
     _Nmax = 4096  # (Don't go bigger than 4096)
 
-    diffraction_psf: DiffractionPSF
+    diffraction_fft: DiffractionFFT
```
cwwalter, line 61: I think this is a type hint.  Does it belong here and not above?

cwwalter (approved):
LGTM a few suggestions to make things clearer to the user and someone reading stamp.py below.

doc/config.rst
rmjarvis: People can use LSST_Silicon without using LSST_CCD.  So if you're using LSST_CCD, you can just use `$det_name`.  But if not, then you need to set it some other way.

imsim/stamp.py
rmjarvis: Honestly, I don't understand this syntax.  It's something Geri put in.  I think it's related to DataClasses somehow, but this class isn't marked with @dataclass, so I'm not really sure what it does.

imsim/stamp.py
rmjarvis: I don't know what you mean by checks.  These are the three kwargs we need for Kolmogorov_and_Gaussian_PSF.  So GetAllParams here returns precisely what we need at this point.

imsim/stamp.py
rmjarvis: It's also using a sub-dict.  The GetAllParams that it calls isn't the same field as we're doing here.  It's config['diffraction_fft'].  So we couldn't include those parameters in the GetAllParams call we do at this point.

imsim/stamp.py
rmjarvis: Oh, I see what you mean.  The return value from GetAllParams here starts out the same as params above.  Yes, we can skip running GetAllParams again now at this point.

imsim/stamp.py
rmjarvis: Actually, I think a better API would have `from_config` take this sub-dict as it's parameter rather than the parent dict.  That would probably make it clearer here why this does a second round of GetAllParams.  I'll switch tot hat.

imsim/stamp.py
cwwalter: I think this is a type hint for something like mpy to use.  It's letting the type checker know that difraction_fft is of type DiffractionFFT.  But, I have never actually used this before.

rmjarvis: Apparently this is a type annotation.  (Maybe you knew that already.)  https://peps.python.org/pep-0526/

Nothing breaks when I get rid of it.  And it strikes me as not very useful, tbh.  We aren't doing type annotations other places in imSim, and I suspect we mostly don't want to.  We rely on duck typing most of the time, rather than static type checking, so I think we should just remove this line.

doc/config.rst
rmjarvis: I added a note in the docs about all this.

cwwalter (approved):
LGTM now.

jchiang87 (approved):

PR #402, opened 2023-08-10
fft diffraction: Fix spike offset

g-braeunlich:


g-braeunlich:
The bug was in principle, that I used `mode="same"` instead of `mode="full"`.
I moved the convolution logic into its own function and added a test which checks convolution + adding images

jchiang87 (approved):
Thanks! lgtm

rmjarvis (approved):
Seems right.  Do you have an update to how things look in practice.  E.g. this kind of thing that precipitated the issue?
image

jchiang87:
I've made the comparison for the test inputs I posted in our [slack thread](https://lsstc.slack.com/archives/CHXKSF3HC/p1691542283369069?thread_ts=1691420959.989129&cid=CHXKSF3HC).  Left is before the fix and right is with these changes.
PR402_fix

rmjarvis:
Great!  Lgtm.

cwwalter:
Great!

PR #403, opened 2023-08-14
Switch to recursive templates

cwwalter:
Now that it is supported in GalSim, switch config system to use nested templates.  Now the examples call the specific input templates, which in turn call the base template.

Note: this change requires you use GalSim 2.4.11 which is included in weekly _32.  The documentation and CI were also updated to reflect this.


config/imsim-config-instcat.yaml
```diff
@@ -9,369 +9,30 @@
...
-            # TODO: ... more probably
-
-gal:
-    type: InstCatObj
+# Get most of the configuration from the imSim config-template
+# for instance catalogs.
+template: imsim-config
 
 ################################################################
 #  Above is template material.
```
rmjarvis, line 17: I guess this is still true.  But since it's one line above now, this comment should probably be amended.

config/imsim-config-instcat.yaml
```diff
@@ -9,369 +9,30 @@
...
+    file_name: default_catalog_file.txt
+    sed_dir: $os.environ.get('SIMS_SED_LIBRARY_DIR')
 
-# input.opsim_data:
-#     # Read the visit meta data by default we use the same file as the input.
-#     # However you could specify a opsim file or a another instance catalog
-#     # instead.
-#     file_name: '@input.instance_catalog.file_name'
+input.opsim_data:
+    # Read the visit meta data by default we use the same file as the input.
```
rmjarvis, line 28: This should be two sentences.  And rather than "input", this should be something like "above instance catalog with the object information".

rmjarvis (approved):
Just a couple minor requests.

config/imsim-config-instcat.yaml
cwwalter: OK I worked on the text in both files.

config/imsim-config-instcat.yaml
cwwalter: Done.

cwwalter:
Ugh.. Thanks Mike. I looks like word wrap was on and that was invisible to me.

doc/install.rst
jchiang87, line 1: I think we need to state the GalSim (as well as skyCatalogs (v1.5.0rc2)) version that's required.  Not everyone will use the stackvana distribution, e.g., I build from the lsstsqre docker images for running on my laptop and in shifter images at NERSC, so I would need to specify those package versions for building those images.

tests/test_header_keywords.py
```diff
@@ -6,6 +6,7 @@
 from astropy.io import fits
 import numpy as np
 import galsim
+import imsim
```
jchiang87, line 9: I don't think this is the best way to make this module available.  It's not explicitly used in the code below, so I think it's confusing to have it imported here.   I think it makes more sense to add a `modules` entry in the `config` below, i.e.,
```
    config = {'modules': ['imsim'],
              'template': template,
```
This would be consistent with how that module would be made available in an actual yaml file.

tests/test_object_positions.py
```diff
@@ -6,6 +6,7 @@
 from astropy.io import fits
 import numpy as np
 import galsim
+import imsim
```
jchiang87, line 9: Same comment applies here as I made for the `test_header_keywords.py` change.

jchiang87 (changes requested):
A few minor comments.  Updating the docs to specify versions is probably the most important.

.github/workflows/ci.yml
```diff
@@ -19,7 +19,7 @@ jobs:
             matrix:
                 # For now, just ubuntu, 3.10.  Can add more later.
```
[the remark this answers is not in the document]
cwwalter: Removed.

tests/test_header_keywords.py
cwwalter: Done.

doc/install.rst
cwwalter: Added information on dependencies to the top of the installation section for those not using cvmfs or stackvana.

BTW: This reminded me that I have a place holder just below this for writing down any special notes or instructions for people with Apple silicon machines (which I don't have, so can't test).

jchiang87 (approved):
Thanks!  looks good.

PR #407, opened 2023-08-17
Add ability to add to OPD directly.

jmeyers314:
This PR adds some features that enable the `with_extra_zk` method inside `batoid_rubin` v0.3.0 (which just landed on PyPI, but might need a few hours to land on conda-forge).  With this method, you can first zero-out and then directly set whatever OPD you want for a given field angle.  (There's a new test targeting the "zero-out" the OPD feature in particular).

Along the way, I made some other changes to the OPD and sag processing and outputs.

On the processing side, I eliminated the ImSim function that inferred each optic's inner and outer radii and instead added this capability to batoid directly (now available in v0.5.0 of batoid, which is on PyPI and conda-forge).  This became useful when I started making the obscurations on optics more detailed, effectively placing simulated masking tape on M1 to help track +x and +y coordinates (see https://lsstc.slack.com/archives/CHXKSF3HC/p1691467190092519?thread_ts=1691456649.673369&cid=CHXKSF3HC for example).  The more detailed obscuration model meant I could no longer easily infer the radii from the obscuration itself, so batoid will now let you set radii directly and fall back to using the obscuration if needed.

On the output side, I put a bit more care into the output WCSs for OPDs and sags such that they connect the data to actual coordinates on surfaces.  The details are slightly different for OPD and sag, mostly because I tried to mimic Zemax's slightly odd behavior in batoid for OPD output.  ImSim derives its OPD WCS directly from the batoid output though, so if I ever change batoid's behavior to be less like Zemax, ImSim will automatically inherit the new behavior.

### DIFF
--- a/tests/test_opd.py
+++ b/tests/test_opd.py
@@ -14,117 +14,124 @@ DATA_DIR = Path(__file__).parent / 'data'
 
 
 def test_opd_zemax():
-    with TemporaryDirectory() as d:
-        config = textwrap.dedent(
-            f"""
-            input:
-                telescope:
-                    file_name: LSST_r.yaml
-                    perturbations:
-                        M2:
-                            shift: [100.e-6, 0.0, 0.0]
-            output:
-                dir: {d}
-                opd:
-                    file_name: opd.fits
-                    wavelength: 694.0  # nm
-                    projection: zemax
-                    fields:
-                        - {{thx: 1.121 deg, thy: 1.231 deg}}
-            """
-        )
+    for nx in [255, 256]:
+        with TemporaryDirectory() as d:
+            config = textwrap.dedent(
+                f"""
+                input:
+                    telescope:
+                        file_name: LSST_r.yaml
+                        perturbations:
+                            M2:
+                                shift: [100.e-6, 0.0, 0.0]
+                output:
+                    dir: {d}
+                    opd:
+                        file_name: opd.fits
+                        wavelength: 694.0  # nm
+                        projection: zemax
+                        nx: {nx}
+                        fields:
+                            - {{thx: 1.121 deg, thy: 1.231 deg}}
+                """
+            )
 
-        config = yaml.safe_load(config)
-        galsim.config.ProcessInput(config)
-        galsim.config.SetupExtraOutput(config)
-        galsim.config.SetupConfigFileNum(config, 0, 0, 0)
-        galsim.config.extra.WriteExtraOutputs(config, None)
+            config = yaml.safe_load(config)
+            galsim.config.ProcessInput(config)
+            galsim.config.SetupExtraOutput(config)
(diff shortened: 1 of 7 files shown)


jmeyers314:
The new versions of batoid and batoid_rubin required for this have now landed.  Ready for review.

rmjarvis:
FYI, you can just click rerun jobs on the Actions tab, rather than push an empty commit.  I personally find that easier.

tests/test_opd.py
```diff
@@ -209,7 +216,85 @@ def test_opd_wavelength():
...
+        # Write a new config with phases to zero out the OPD
+        config = textwrap.dedent(
+            f"""
+            input:
+                telescope:
+                    file_name: LSST_r.yaml
+                    fea:
+                        extra_zk:
+                            zk: {(zk*1e-9).tolist()}
+                            eps: 0.612
```
rmjarvis, line 262: Is there a test we can add that checks that this changes the in-focus image too?  I think it should, and as discussed on the telecon, I would have thought that could be used to add a slight random perturbation to the aberrations similar to what we used to do with Bo's model.

jmeyers314:
> FYI, you can just click rerun jobs on the Actions tab, rather than push an empty commit.  I personally find that easier.

Yeah, I looked for that. I think I must not be on the appropriate access list.

tests/test_opd.py
jmeyers314: Sure, I can add a test that just checks that the PSF isn't identical before and after adding a screen.

The difference between adding the phase here and what's in [OptWF](https://github.com/LSSTDESC/imSim/blob/main/imsim/atmPSF.py#L37) is that the current PR is changing the wavefront uniformly over the full field of view, and the optical perturbations from Bo's model are field angle dependent.  

One could certainly extend the current PR to from adding a constant (over-the-field) single Zernike phase to adding a DoubleZernike that captures the field dependence.  But I think that's more than I need at the moment (and really begs for GalSim 2.5 too).

jmeyers314: I added a test that the RubinOptics photon op gives different output after adding the constant phase.  I think that's sufficient and easier than going all the way to an image.

tests/test_photon_ops.py
```diff
@@ -682,7 +683,7 @@ def test_phase_affects_image():
 
     # Now add some phase and process again
     config = galsim.config.CleanConfig(config)
-    config.update(**TEST_BASE_CONFIG) # restore _icrf_to_field
+    config.update(**deepcopy(TEST_BASE_CONFIG)) # restore _icrf_to_field
```
rmjarvis, line 686: The canonical way to do this, rather than deepcopy is `config = galsim.config.CleanConfig(config)` whenever you change a dict and need to reset the compiled bits.

rmjarvis (approved):

PR #409, opened 2023-08-18
bug-fixes for multiprocessing and Gaia stars

jchiang87:
* With the handling of proper motion and parallax for Gaia stars in skyCatalogs, the function call for accessing the skyCatalogs files needs to have the MJD of the observation provided.
* `lsst.afw.cameraGeom.Detector` objects are not pickleable, so just pass the transformation object to the `Vignetting.at_sky_coord` function since the `Vignetting` instance is a proxy object when using multiprocessing, and it therefore needs to be passed pickleable arguments.
* Add test code to run simulations using `outputs.nproc = 2` and an input catalog that exercises the code path that calls the `Vignetting.at_sky_coord` function.

### DIFF
--- a/imsim/skycat.py
+++ b/imsim/skycat.py
@@ -99,7 +99,7 @@ class SkyCatalogInterface:
             sky_cat = skyCatalogs.open_catalog(
                 self.file_name, skycatalog_root=self.skycatalog_root)
             self._objects = sky_cat.get_objects_by_region(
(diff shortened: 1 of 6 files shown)


cwwalter (approved):
I'm not sure I understand the pix_to_fp pickling stuff but it looks good to me and I tested the mjd part yesterday.

rmjarvis (approved):

PR #410, opened 2023-08-22
Make doOpt=True work again.

rmjarvis:
The option to use a regular optical screen rather than the full batoid optics got broken during the changes in #380.  The problem is that there is a bug in GalSim (which we'll try to fix before 2.5, but has been a bug in the 2.4 series) where convolutions of 2 or more PhaseScreenPSFs don't work with photon shooting.  They each separately populate the pupil angles and then there is an error during convolution of the photon arrays, since that looks like an error.

Anyway, the psfs do work correctly if they are put in the photon_ops list, rather than being convolved with the galaxy.  This does almost the same thing, but the code path is a little different, so it avoids repopulating the pupil angles on the second PhaseScreenPSF.

Along the way, I fixed a few other things:
1. I discovered that LSST_Image didn't work when not using an LSST_CCD output type.  So I fixed that.  Mostly by having it figure out the image size itself and having the det_name value be explicit rather than just looking for it in base['det_name'].
2. A couple other classes had similar implicit dependence on LSST_CCD by looking for base['det_name'], so I fixed them the same way.
3. There were also some places that used xsize, ysize and expnum in base, which are set by LSST_CCD.  Those all have reasonable defaults, so I use the defaults now if they aren't set in base, but use the values if LSST_CCD has set them.
4. I removed most of the mocking that was in the tests in test_stamp.py, since they broke fairly catastrophically with the relatively minor changes I made to the draw function.  What's left might still be fragile to future code changes, but I think much less so than it was.  It's closer now to a regular integration test where the only thing mocked is the drawImage function so we can check the arguments it's getting.

### DIFF
--- a/config/imsim-config.yaml
+++ b/config/imsim-config.yaml
@@ -44,8 +44,8 @@ eval_variables:
     fexptime: { type: OpsimData, field: exptime }
 
     # Note: Another value we use in many eval items is det_name.  If using the LSST_CCD
-    #       output type, as we do here, it's added to the eval_variables automatically.
-    #       Otherwise you should set sdet_name here manually to e.g. R22_S11 or whatever.
+    #       output type, as we do here, it's automatically added as an availble eval variable.
+    #       Otherwise you should set sdet_name here manually to e.g. 'R22_S11' or whatever.
 
 
 # Any input data is set here.  These are read in at the start of the program and
--- a/config/imsim-config.yaml
+++ b/config/imsim-config.yaml
@@ -43,6 +43,10 @@ eval_variables:
 
     fexptime: { type: OpsimData, field: exptime }
 
+    # Note: Another value we use in many eval items is det_name.  If using the LSST_CCD
+    #       output type, as we do here, it's added to the eval_variables automatically.
+    #       Otherwise you should set sdet_name here manually to e.g. R22_S11 or whatever.
+
 
 # Any input data is set here.  These are read in at the start of the program and
 # potentially updated for each output file. Also includes things that need some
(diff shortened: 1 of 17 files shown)


jmeyers314 (approved):
LGTM.  One quick general question: is `det_num` also easy to make as accessible in the config as `det_name`?  I can imagine use cases where you'd set something in the config via `blah: $X if det_num >= 189 else Y` to select wavefront sensors for example.

rmjarvis:
> is det_num also easy to make as accessible in the config as det_name? 

Yep.  Added.  I also switched to using eval_base_variables for these two, rather than adding them to the config's eval_variables.  That does seem a bit neater, IMO.

rmjarvis:
Does anyone else want to look at this?  Or ok to merge?

cwwalter:
OK for me to merge, but can you summarize the change in how we use det_name? I'm not sure I follow all of it, is it the case that we can't use $det_name anymore and should always rely on the value in base instead?

rmjarvis:
> is it the case that we can't use $det_name anymore and should always rely on the value in base instead?

The opposite.  The main point is that we can't internally just use `base['det_name']` all over the place, since it isn't necessarily present.  It's only there if the user is using the `LSST_CCD` output type.  We don't want to force them to use that output type in order to use other imsim modules.

So in any modules that need to know the detector name, this PR makes it a regular required parameter.  If you *are* using LSST_CCD, then putting `det_name: $det_name` is a very reasonable thing to do.  And if not, you can either give a specific value instead, or set up `eval_variables.sdet_name` to be something appropriate.

cwwalter:
Ah, OK. That makes sense now. Thanks.

PR #411, opened 2023-08-24
Enable file_num parsing in fea config

jmeyers314:
Allows the telescope to be changed file-by-file in multifile configs.  For example, one can add a different zernike perturbation to each output file via something like:
```yaml
template: ./fiducial_in_focus.yaml

input.telescope.fea:
  extra_zk:
    zk: $[0]*(4+file_num)+[1e-6]
    eps: 0.612
output.dir: $f"zk_{file_num+4:02d}"
```
This snippet will add some Z4 to file 1, some Z5 to file 2, and so on.

Making this possible basically involved separating the config parsing logic from the telescope assembly logic, making sure the former passes as much as possible through `galsim.config.ParseValue` and related.

### DIFF
--- a/imsim/telescope_loader.py
+++ b/imsim/telescope_loader.py
@@ -1,4 +1,3 @@
-import os
 from galsim.config import (
     InputLoader, RegisterInputType, GetAllParams, get_cls_params, ParseValue,
     LoggerWrapper
@@ -16,7 +15,9 @@ def parse_xyz(xyz, base):
     return parsed_xyz, all(safe)
 
 
-def parse_fea(config, base, telescope):
+def apply_fea(
+    fea_perturbations, telescope, fea_dir=None, bend_dir=None
+):
     """ Parse a finite element analysis config dict.  Used to add detailed
     effects like mirror print through or temperature originating figure
     perturbations.  Also facilitates direct manipulation of standard active
@@ -24,12 +25,14 @@ def parse_fea(config, base, telescope):
 
     Parameters
     ----------
-    config : dict
-        The configuration dictionary to parse.
-    base : dict
-        Base configuration dictionary.
+    fea_perturbations : dict
+        The perturbations dictionary to parse.
     telescope : batoid.Optic
         Telescope to perturb
+    fea_dir : str, optional
+        Directory containing batoid_rubin FEA data.
+    bend_dir : str, optional
+        Directory containing batoid_rubin bending mode data.
 
     Examples of fea config dicts
     ----------------------------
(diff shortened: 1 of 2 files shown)


rmjarvis (approved):
lgtm

cwwalter:
I'm a bit confused but what file file_num refers to above.

It is being used to set an output directory and I assume this file is not the per CCD files.  Is I for multiple telescope files?

jmeyers314:
> I'm a bit confused but what file file_num refers to above.

Yeah, that's something I learned yesterday.  It does increment for every detector, but you can make more complicated configs that track "exposure_num" instead of just file_num by dividing by the number of sensors.  Here's a recent snippet I've been playing with:

```yaml
template:...

eval_variables:
  lwavefront_sensors: [191, 192, 195, 196, 199, 200, 203, 204]
  lscience_sensors: $list(range(189))
  lsensors: $wavefront_sensors  # do just the wavefront sensors
  # lsensors: $wavefront_sensors+science_sensors  # do all the sensors
  izk: "$@file_num//len(sensors)+4"
  iseqnum: $zk+100

input.telescope.fea:
  extra_zk:
    zk: $[0]*zk + [perturbation[zk]]
    eps: 0.612

output.dir: $f"intra/zk_{zk:02d}"
output.nfiles: $25*len(sensors)  # 4-28 inclusive
```

This runs 25 different "exposures" applying a different extra zernike phase to each exposure and placing each result in its own directory.  Each exposure produces 8 WF sensor CCDs as configured, but you can see how you'd modify to produce the 189 science sensors or 197 WF+science sensors.  In fact, I ran this earlier today on the science sensors to produce 25*189 = 4725 CCD images all from a single invocation of `galsim blah.yaml`.

cwwalter:
Ah, I see.  OK thanks.

One question: since file_num is a galsim builtin, why do you use the @ (which I thought was for things in the dictionary in the rest of the config file) instead of just using it as a variable like you did in your first example?

Is it related to being in the eval section?

rmjarvis:
The @ is unnecessary there, but allowed.  @file_num means get base['file_num'].  But this is one of the things that GalSim automatically makes available in eval statements too.  cf. https://github.com/GalSim-developers/GalSim/blob/releases/2.4/galsim/config/value_eval.py#L58

cwwalter:
OK, I see. Thanks.

Maybe if we have more variables than just det_num, it would be good to have an imsim only equivalent of galsim.config.eval_base_variables we could add things too which could then be added to the galsim one all at once.

Issue #413, opened 2023-08-30
Add FOCUSZ into amp headers


jbkalmbach:
For the AOS LSSTCam Full Array Mode we need to know the defocus distance from the FITS header to figure out which images are intra-focal and which are extra-focal.

cwwalter:
Can you tell us how this will be done in normal operations?  Does this just report the camera hexapod z-position?

cwwalter:
Also note: there is already a YAML command if you just want to add a header.

If we put this in the normal readout, it should be something that will be added in the normal headers in operation, + we need to understand how the value will filled.

jbkalmbach:
> Can you tell us how this will be done in normal operations? Does this just report the camera hexapod z-position?

Let me double check on slack from people who know better but I think it is the user-defined offset in the camera hexapod z-position beyond the corrections from the system.

> Also note: there is already a YAML command if you just want to add a header.

I tried to do this but the value only showed up in the eimage headers not the amp headers. Is there a way to pass it to both?

cwwalter:
> Let me double check on slack from people who know better but I think it is the user-defined offset in the camera hexapod z-position beyond the corrections from the system.

hmm... OK. I was thinking it would be tied to one of the yaml perturbation commands.  Is there a separate one for the user offset?

> I tried to do this but the value only showed up in the eimage headers not the amp headers. Is there a way to pass it to both?

Ah, @jmeyers314 added that option.  Josh, was your intention to only go to the eimage?

jbkalmbach:
> hmm... OK. I was thinking it would be tied to one of the yaml perturbation commands. Is there a separate one for the user offset?

It is when I'm running simulations. It is the same value I add to the aos degrees of freedom:
```
imsimConfigYaml["input"]["telescope"]["fea"]["aos_dof"]["dof"][5] += obsMetadata.focusZ
```

where the degrees of freedom are:
```
    # 0: M2 dz (um)
    # 1,2: M2 dx,dy (um)
    # 3,4: M2 rx,ry (arcsec)
    # 5: camera dz (um)
    # 6,7: camera dx,dy (um)
    # 8,9: camera rx,ry (arcsec)
    # 10-29: M1M3 bending modes (um)
    # 30-49: M2 bending modes (um)
```

cwwalter:
OK. Great. So then we should be able to just set the header to that value in the dictionary if it exists.

Did you confirm that this is a standard header that will normally be in the output?

[...]

PR #414, opened 2023-08-30
Add FOCUSZ to amp headers.

jbkalmbach:


rmjarvis:
What you have so far is just going to put FOCUSZ = 0 in all the chips.  That's probably not what you want.  Presumably for this to be useful, it should be the actual z height of each CCD.

jbkalmbach:
I think it should be 0.0 most of the time actually. It should describe a user defined offset in the focal plane. When I add it into the imsim config yaml under `output.header` then it gets passed through as the correct value (a value added to the camera z position in the AOS degrees of freedom).

jbkalmbach:
Closing this to add this in as a telescope parameter as proposed by Josh in #413

PR #415, opened 2023-09-11
modify treerings.py to defer loading data for a given CCD until it is requested

jchiang87:


imsim/treerings.py
```diff
@@ -22,14 +22,14 @@ class TreeRings():
...
-    def __init__(self, file_name, only_dets=None, logger=None):
+    def __init__(self, file_name, only_dets=None, logger=None, defer_load=True):
         """
         Constructor.
         Craig Lage UC Davis 19-Mar-18; cslage@ucdavis.edu
         This code reads in a file with tree ring parameters from file_name
         and assigns a tree ring model to each sensor.
         """
-        logger = galsim.config.LoggerWrapper(logger)
+        self.logger = galsim.config.LoggerWrapper(logger)
```
rmjarvis, line 32: This is likely to cause problems, since loggers aren't picklable.  Better to pass the logger to functions that need it each time, rather than save it as an attribute.

rmjarvis (changes requested):
This is a good idea.  My only quibble is that it's probably a bad idea to save the logger as an attribute.  Especially for input types, this tends to cause trouble.

Unfortunately, TreeRings are accessed via functions that don't get passed a logger.  So we can either (1) remove the logging commands from fill_dict, or (2) only use them when `defer_load = False`, or write custom `__setstate__`/`__getstate__` functions that pop off the logger from `__dict__` when pickling the input object.

jchiang87:
Thanks, Mike.   I updated the code so that the logger is no longer an attribute, removed the logging commands from `fill_dict`, and instead have relevant log messages emitted from the initializer.

PR #416, opened 2023-09-12
Add FocusZ to config and to headers

jbkalmbach:
* Add focusZ option to perturbations to easily set defocal distance of science sensors. 
* Add FOCUSZ to headers.

### DIFF
--- a/tests/test_telescope_loader.py
+++ b/tests/test_telescope_loader.py
@@ -662,6 +662,37 @@ def test_config_fea_file_num():
 
         assert configd == reference
 
+def test_config_focus_z():
+    """Test that focusZ is properly set from configs."""
+
+    telescope = imsim.load_telescope("LSST_r.yaml")
+    shifted_ref = (telescope
+        .withLocallyShiftedOptic('LSSTCamera', [0.0, 0.0, 1e-3])
+    )
+
+    focusz_shifted_function_call = imsim.load_telescope("LSST_r.yaml", focusZ=1e-3)
+
+    assert shifted_ref == focusz_shifted_function_call
+
+    shifted_config = textwrap.dedent(
+            """
+            input:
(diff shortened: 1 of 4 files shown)


rmjarvis:
I see you added a test image, but no test code.  There should be some kind of test of this new functionality.

jbkalmbach:
> I see you added a test image, but no test code. There should be some kind of test of this new functionality.

Thanks @rmjarvis. I've added a test for the telescope loader and I think this is ready for review.

rmjarvis:
If you rebase onto the current main, it should fix the rubin-sim.version problem you're currently getting.

jmeyers314 (approved):
Looks good to me, Bryce.  Thanks for the improvement.

tests/test_telescope_loader.py
```diff
@@ -662,6 +662,37 @@ def test_config_fea_file_num():
...
+    config = yaml.safe_load(shifted_config)
+    galsim.config.ProcessInput(config)
+    focusz_shifted_config = galsim.config.GetInputObj(
+        'telescope',
+        config['input']['telescope'],
+        config,
+        'telescope'
+    ).fiducial
+
+    assert shifted_ref == focusz_shifted_config
```
rmjarvis, line 695: This test doesn't seem to use the image you added to tests/data.  Do we need to add that to the repo?

jbkalmbach: I added `FOCUSZ` to the headers in the existing file because without it multiple tests in `test_readout.py` fail.

rmjarvis: Oh, I see.  I thought it was a new file.  It's just changes to the header in the existing one.  Got it.

rmjarvis (approved):

PR #417, opened 2023-09-13
process info extra output

jchiang87:
* Add extra output with memory and cpu info for each process as each stamp is rendered
* add realized flux to truth output
* clean up test outputs from `test_object_positions.py`

### DIFF
--- a/imsim/process_info.py
+++ b/imsim/process_info.py
@@ -31,8 +31,12 @@ class ProcessInfoBuilder(ExtraOutputBuilder):
         # User cpu time
         user_time = proc.cpu_times().user
 
-        self.scratch[obj_num] \
-            = ProcessInfo(object_id, pid, rss, uss, user_time, time.time())
+        process_info = ProcessInfo(object_id, pid, rss, uss, user_time,
+                                   time.time())
+        logger.info("Object %d, id %s, pid %d, RSS %.2f GB, USS %.2f GB, "
+                    "user_time %.2f, unix_time %.1f, det_name %s",
+                    obj_num, *process_info, base.get("det_name", ""))
+        self.scratch[obj_num] = process_info
 
     def finalize(self, config, base, main_data, logger):
         self.cat = galsim.OutputCatalog(names=_PROCESS_INFO_COLS,
--- a/imsim/process_info.py
+++ b/imsim/process_info.py
@@ -1,0 +1,49 @@
+import os
+import time
+from collections import namedtuple
+import psutil
+import galsim
+from galsim.config import ExtraOutputBuilder, RegisterExtraOutput
+
+
+_PROCESS_INFO_COLS = ['object_id', 'pid', 'rss', 'uss', 'user_time',
+                      'unix_time']
+_PROCESS_INFO_TYPES = [str, int, float, float, float, float]
+ProcessInfo = namedtuple('ProcessInfo', _PROCESS_INFO_COLS)
+
+
+class ProcessInfoBuilder(ExtraOutputBuilder):
+    """
+    Build output file with per-process cpu and memory info extracted
+    using psutil.
+    """
+    def processStamp(self, obj_num, config, base, logger):
+        object_id = base.get('object_id')
+
+        pid = os.getpid()
+        proc = psutil.Process(pid)
+        mem_full_info = proc.memory_full_info()
+
+        # Compute memory in GiB
+        rss = mem_full_info.rss/1024**3
+        uss = mem_full_info.uss/1024**3
+
+        # User cpu time
+        user_time = proc.cpu_times().user
+
+        self.scratch[obj_num] \
+            = ProcessInfo(object_id, pid, rss, uss, user_time, time.time())
+
+    def finalize(self, config, base, main_data, logger):
+        self.cat = galsim.OutputCatalog(names=_PROCESS_INFO_COLS,
+                                        types=_PROCESS_INFO_TYPES)
+        obj_nums = sorted(self.scratch.keys())
+        for obj_num in obj_nums:
+            self.cat.addRow(self.scratch[obj_num])
+        return self.cat
+
+    def writeFile(self, file_name, config, base, logger):
+        self.cat.write(file_name)
+
+
+RegisterExtraOutput('process_info', ProcessInfoBuilder())
--- a/imsim/process_info.py
+++ b/imsim/process_info.py
@@ -47,6 +47,7 @@ class ProcessInfoBuilder(ExtraOutputBuilder):
         return self.cat
 
     def writeFile(self, file_name, config, base, logger):
+        logger.warning("Writing process info output to %s", file_name)
         self.cat.write(file_name)
 
 
--- a/imsim/stamp.py
+++ b/imsim/stamp.py
@@ -410,8 +410,8 @@ class LSST_SiliconBuilder(StampBuilder):
         # Otherwise (high flux object), we might want to switch to fft.  So be a little careful.
         bandpass = base['bandpass']
         fft_psf = self.make_fft_psf(psf.evaluateAtWavelength(bandpass.effective_wavelength), logger)
-        logger.warning('Object %d has flux = %s.  Check if we should switch to FFT',
-                       base['obj_num'], self.realized_flux)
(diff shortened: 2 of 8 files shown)


imsim/stamp.py
```diff
@@ -434,14 +436,16 @@ def buildPSF(self, config, base, gsparams, logger):
                 flux = self.gal.flux*self.vignetting.at_sky_coord(
                     base['sky_pos'], self.image.wcs, pix_to_fp)
                 self.realized_flux = galsim.PoissonDeviate(self.rng, mean=flux)()
+                base['realized_flux'] = self.realized_flux
```
rmjarvis, line 439: I think this implies that realized_flux means two different things for vignetted objects depending on whether they were rendered with fft or not.  That's probably not ideal for an output catalog quantity.  I think we probably just want the other realized_flux always for output to be the number of photons from the object, before any possible loss from vignetting.
If you want to keep track of vignetting effects (for all objects), then you could additionally record `image.added_flux` after the `drawImage` call.  That records how much flux actually made it onto the stamp when drawing.

tests/test_psf.py
```diff
@@ -372,6 +376,7 @@ def test_atm_psf_fft(self):
                     'rawSeeing': self.opsim_data['rawSeeing'],
                     'band':  self.opsim_data['band'],
                     'screen_size': 409.6,
+                    'nproc': 1,
```
rmjarvis, line 379: Were these causing problems using multiple processes?  I'm not sure why you added nproc: 1 to these.  I think it's a good idea to have at least one test use it, just to make sure things work with that parameter.  But similarly, we should keep at least one tests without it to make sure things work with the default nproc: None.  I'd even suggest adding one that is explicitly nproc: 6 to make sure that works ok. (Since that's what None normally turns into.)

rmjarvis (changes requested):
This is a nice addition.  Just a couple minor comments.

imsim/process_info.py
```diff
@@ -0,0 +1,54 @@
...
+        # Compute memory in GiB
+        rss = mem_full_info.rss/1024**3
+        uss = mem_full_info.uss/1024**3
+
+        # User cpu time
+        user_time = proc.cpu_times().user
+
+        process_info = ProcessInfo(object_id, pid, rss, uss, user_time,
+                                   time.time())
+        logger.info("Object %d, id %s, pid %d, RSS %.2f GB, USS %.2f GB, "
```
cwwalter, line 36: So, if we set the log level to 2 it will write both to the log and the file correct?

If the level is set to 1 "warning" this won't be in the log but the file will still be created and written, is that correct?  Is there a way to tie these together so the extra output only shows up on 2?

cwwalter (approved):
Looks good but can we also add documentation to the config rst about these options? Two suggestion/questions below.

imsim/process_info.py
cwwalter: Thinking more about this, I can certainly imagine vases where we don’t want this coupled. But maybe in the YAML we could have a log level threshold for the the output files which is 2 by default.

rmjarvis: I don't understand the ask here.  If you don't want this as an extra_ouput, you just remove it from the config.  

Or do you want the option to have it in the config, so it makes logging statements, but not write a file?  If so, that's already possible by simply omitting the file_name from this extra_output.  Then it will run things, but not write to a file.

imsim/stamp.py
rmjarvis: Thinking more about this, I realized that the other realized_flux wouldn't actually be right anymore given the way the code works here.  I think what we want here is to keep the previous realized_flux, but then don't re-draw the poisson variate here.  Rather, just apply the vignetting.
```
flux = self.realized_flux*self.vignetting.at_sky_coord(
                    base['sky_pos'], self.image.wcs, pix_to_fp)
self.realized_flux = round(self.realized_flux)
```

imsim/process_info.py
cwwalter: Sorry.. I was unclear:

The option is now in the default config.  So, this means that users that don't understand they need to put a 

option: "" 

in their yaml files are always going to get this output (which can be hundreds of files).  So, I would like reasonable behavior for users while still being able to have the pattern etc setup in the the config file.  This is unlike checkpoints,  (which can always be usable If jobs need to restart), or centroids.  It's really a developer feature.

Ideally, it would be setup as now in the config file but the user would have to flip one option to have it start producing the output. I was suggestion perhaps tying it to a a minimum logging level so that by default they would start to show up when you asked for -v 2.

rmjarvis: I think we should not put it in the template config.  As developers, we know how to add it if we want it.

tests/test_psf.py
jchiang87: Yes, for some reason, I kept seeing the CI killed during test_psf.py, i.e., [here](https://github.com/LSSTDESC/imSim/actions/runs/6175279590/job/16762744913) and[ here](https://github.com/LSSTDESC/imSim/actions/runs/6176046866/job/16764176869).  I have no idea why it started to fail with my build, but I noted in the [github runner docs](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources)  that linux VMs only have 7GB of RAM, so I made `nproc=1` to avoid the RSS exceeding that number.

imsim/process_info.py
jchiang87: > I think we should not put it in the template config. As developers, we know how to add it if we want it.

That's fine with me, but I think the format needs to be documented somehow.   Keeping it in `imsim-config.yaml`  but commented-out seems like a convenient way to do that and have it handy.

rmjarvis: Well, we have a documentation page.  It would be appropriate to add information about this type there.

tests/test_psf.py
rmjarvis: In that case, maybe just add that for the one that uses `'screen_size': 409.6,`  That's the only one that should have a large memory footprint.  (Although I don't understand why using multiple processes to calculate it would be connected with higher memory...)

rmjarvis: Oh, I guess there are temporary Fourier planes during the initialization.  So with nproc=6, you have all 6 of those in memory at once, rather than one at a time.  That does make sense.

imsim/process_info.py
jchiang87: I added to the config.rst docs and put a full example there and deleted the entry from `imsim-config.yaml`

imsim/stamp.py
jchiang87: I've implemented this.

imsim/process_info.py
cwwalter: Great. Thanks.

Another thing we could do if you wanted something particularly easy to copy/paste start from would be to add a new

imsim-developer.yaml

to the examples directory that is set up for doing tests etc.

imsim/stamp.py
```diff
@@ -427,21 +429,22 @@ def buildPSF(self, config, base, gsparams, logger):
...
             if self.vignetting is not None:
                 pix_to_fp = self.det.getTransform(cameraGeom.PIXELS,
                                                   cameraGeom.FOCAL_PLANE)
-                flux = self.gal.flux*self.vignetting.at_sky_coord(
+                vignetted_flux = self.realized_flux*self.vignetting.at_sky_coord(
                     base['sky_pos'], self.image.wcs, pix_to_fp)
-                self.realized_flux = galsim.PoissonDeviate(self.rng, mean=flux)()
+                self.realized_flux = round(vignetted_flux)
 
-            logger.warning('Yes. Use FFT for this object.  max_sb = %.0f > %.0f',
```
cwwalter, line 438: Sorry... I thought I added a comment here before but somehow,  I don't see it now so maybe I didn't save it correctly.

Can we change these from WARNING to level 2?

cwwalter (approved):

tests/test_psf.py
rmjarvis: I still think this is the only one of these that should have `nproc: 1`.  The others should not have a significant memory footprint.

Issue #419, opened 2023-09-18
We are double counting the noise during FFT drawing.


cwwalter:
@matroxel noticed that in FFT drawing we use the realized flux and then add noise later. 

https://github.com/LSSTDESC/imSim/blob/main/imsim/stamp.py#L577
https://github.com/LSSTDESC/imSim/blob/main/imsim/stamp.py#L626

This seems like double counting the noise and it looks like line 577 should be

```
gal = gal.withFlux(self.flux, bandpass)
```

I think this might be a transcription error from when the code was being copied from the old system into the new GalSim config code base.  @rmjarvis can you confirm that's what happened and we should change this?

matroxel:
You still need that line with realized_flux within the second if branch for the nonFFT version.

rmjarvis:
Yes, Troxel and I already talked about it.  We should be using the non-Poisson flux for FFT drawing, since the addition of Poisson noise later does the job.

rmjarvis:
Except the change you mentioned isn't correct.  Rather, we need to move that line as it stands to inside the phot branch.  And use the regular flux in the fft branch.

rmjarvis:
Another thing we talked about is that in the Roman code, Troxel has been making decisions about when to switch to FFT based on the nominal flux, rather than the realized flux.  This is probably better, so a given object has the same behavior on all exposures, rather than possibly switching from phot to FFT (or in our case, real vs trivial SED) from one exposure to another. 

I think this makes more sense, so unless I get push back, let's do that in imSIm too.

cwwalter:
@rmjarvis and @jchiang87 @matroxel Checking: did this happen in both code bases?

rmjarvis:
Yes, I think so.

cwwalter:
Thanks.  Closing.

PR #420, opened 2023-09-18
Fix flux treatment of FFT renders to not double the Poisson noise

rmjarvis:
cf. Issue #419 

1. Switch to using the nominal flux for FFT renders, since we then add Poisson noise to the stamp image, which already does the work of Poisson-ifying the realized flux.
2. Switch the flux value used for all decisions (phot vs fft, faint vs not) to use the nominal flux, rather than the realized flux.
3. Changed the meaning of realized flux in the output file to be the actual number of photons that landed on the image, rather than the number of photons shot.  This way vignetting is consistently included in this definition regardless of whether the object used phot or fft.
4. Added phot_flux and fft_flux values in the output file.  Only one of these is non-zero, and it is the flux used for each rendering type.  So phot_flux is what used to be called realized_flux for photon shooting.  And fft_flux is what used to be called realized_flux for fft renders.  These each have slightly different meanings, but they are well-defined for each method.
5. Added tests of all these flux values in the output catalog.

### DIFF
--- a/imsim/stamp.py
+++ b/imsim/stamp.py
@@ -456,6 +456,7 @@ class LSST_SiliconBuilder(StampBuilder):
             else:
                 self.fft_flux = self.nominal_flux
             base['fft_flux'] = self.fft_flux
+            base['phot_flux'] = 0.  # Indicates that photon shooting wasn't done.
 
             logger.info('Yes. Use FFT for object %d.  max_sb = %.0f > %.0f',
                         base.get('obj_num'), max_sb, fft_sb_thresh)
--- a/imsim/stamp.py
+++ b/imsim/stamp.py
@@ -126,16 +126,17 @@ class LSST_SiliconBuilder(StampBuilder):
         self.nominal_flux = gal.flux
 
         # For photon shooting rendering, precompute the realization of the Poisson variate.
-        # Mostly so we can possibly abort early if realized_flux=0.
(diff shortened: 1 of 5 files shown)


imsim/stamp.py
```diff
@@ -123,10 +123,20 @@ def setup(self, config, base, xsize, ysize, ignore, logger):
...
+        base['nominal_flux'] = self.nominal_flux
+        base['phot_flux'] = self.phot_flux
+        base['fft_flux'] = 0.  # For fft drawing, this will be updated in buildPSF.
+        base['realized_flux'] = 0.  # Will update this after drawImage using im.added_flux
+
+        # Check if the Poisson draw for the photon flux is 0.
+        if self.phot_flux == 0:
             # If so, we'll skip everything after this.
             # The mechanism within GalSim to do this is to raise a special SkipThisObject class.
             raise galsim.config.SkipThisObject('realized flux=0')
```
jchiang87, line 142: I guess this should be
```
raise galsim.config.SkipThisObject('phot_flux=0')
```

jchiang87 (approved):
lgtm.  One small comment about updating a log message.

cwwalter:
Thanks Mike!

jchiang87:
Should we go ahead and merge this?  Everything looks good to me.

rmjarvis:
Sure.

PR #422, opened 2023-09-20
Update the memory used by the atmospheric screen.

cwwalter:
Use the production instead of test value in the default config.


cwwalter:
Did the last commit mean to say "ensure that the imsim versions.." are used?

jchiang87:
No, we want to use the GalSim versions for the regression test so that it's consistent with [this](https://github.com/LSSTDESC/imSim/blob/main/tests/test_sky_model.py#L37).

cwwalter:
OK, that is what I thought you were trying to do (so it was robust against these sort of changes) but I was confused because I saw this:

image


and it looked like you were switching from GalSim to the imsim bandpass.

jchiang87:
I think that was just a coding error (not sure where it came from).  As far as I can tell, `galsim.RubinBandpass` doesn't exist, at least not in GalSim 2.4.11, but `imsim.RubinBandpass` does.

cwwalter:
Ah.. OK. I see. I think that change was made when Mike added the new bandpass stuff. It's odd it made it through the tests.

jchiang87:
That script doesn't get run as part of the tests.  It needs to be run by hand to generate the reference file.  It would have failed much earlier when the skybrightness interface changed.

tests/sky_level_reference_values.py
```diff
@@ -19,12 +22,14 @@
...
 with warnings.catch_warnings():
     warnings.simplefilter('ignore')
-    sky_model.setRaDecMjd(ra, dec, mjd, degrees=True)
+    sky_model.set_ra_dec_mjd(ra, dec, mjd, degrees=True)
 
 sky_levels = {}
 for band in 'ugrizy':
-    bandpass = galsim.RubinBandpass(band)
-    wave, spec = sky_model.returnWaveSpec()
+    with mock.patch('os.getenv', return_value=''):
```
rmjarvis, line 29: I think there is probably no need to do this.  Should probably just use the regular RubinBandpass, rather than the old GalSim ones.

rmjarvis: Then also remove the mocking in test_sky_model.

jchiang87: Then we need to change the code in test_sky_model also.

rmjarvis: Right.  Since we learned we can't actually do the regression test we thought we were doing, I don't see any point in using the galsim Bandpass for that test.

rmjarvis: BTW, I added a test to test_bandpass.py that checks the fallback to GalSim's files when RUBIN_SIM_DATA is not defined.

config/imsim-config.yaml
```diff
@@ -80,7 +80,7 @@ input:
         t0: 0               # seconds
         exptime: $exptime
         kcrit: 0.2          # in units of 1/r0
-        screen_size: 409.6  # Default=812.2, which takes a lot of memory, so use this for testing.
+        screen_size: 812.2  # To use less memory for testing/development, you can try 409.6
```
jmeyers314, line 83: Should be 2^13 * 0.1 = 819.2.  812.2 must be a typo.

jchiang87 (approved):
This all seems fine to me, though someone else should check that parts that I changed.

rmjarvis (approved):

PR #423, opened 2023-09-28
Update documentation after latest changes.

cwwalter:
Update the instructions to point at the new throughput files.

Also, add real note to Apple Silicon users and fix a few things that changed recently but weren't reflected here.

If there are any other documentation changes, we could add them here before merging.


rmjarvis (approved):
lgtm

PR #425, opened 2023-09-28
Parse LSSTBuilder args using config machinery

jmeyers314:
I added a couple new arguments to the batoid_rubin:LSSTBuilder constructor.  They didn't get picked up by ImSim config though, and in fact, I don't think the previously `bend_dir` and `fea_dir` were getting picked up either.  So this PR adds galsim config processing for the LSSTBuilder arguments.  

This will depend on a not-yet-pushed branch in batoid_rubin, so I don't expect CI to pass yet.  But I'm hopeful I'll be able to merge the batoid_rubin stuff tonight or tomorrow.  Just wanted to stake this PR now before any new versions are tagged.

### DIFF
--- a/imsim/telescope_loader.py
+++ b/imsim/telescope_loader.py
@@ -16,7 +16,7 @@ def parse_xyz(xyz, base):
 
 
 def apply_fea(
-    fea_perturbations, telescope, fea_dir=None, bend_dir=None
+    fea_perturbations, telescope, **kwargs
 ):
     """ Parse a finite element analysis config dict.  Used to add detailed
     effects like mirror print through or temperature originating figure
@@ -29,10 +29,8 @@ def apply_fea(
         The perturbations dictionary to parse.
     telescope : batoid.Optic
         Telescope to perturb
-    fea_dir : str, optional
-        Directory containing batoid_rubin FEA data.
-    bend_dir : str, optional
-        Directory containing batoid_rubin bending mode data.
+    **kwargs: dict
+        Additional arguments to pass to the LSSTBuilder constructor.
 
     Examples of fea config dicts
     ----------------------------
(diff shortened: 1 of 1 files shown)


jmeyers314:
Hmmm....  Not sure what I changed that would affect the fringing unit tests; I'll try to dig in later today.  Assuming that's solvable though, this is now ready for review (new version of batoid-rubin is now available on conda-forge).

jmeyers314:
Reran and this time it passed.  Hmmm....

rmjarvis:
I missed what the failure was, but maybe we're missing a deterministic seed in one of the tests.

jmeyers314:
I think it was [this line](https://github.com/LSSTDESC/imSim/blob/main/tests/test_fringing.py#L53).  Not obviously probabilistic.

imsim/telescope_loader.py
```diff
@@ -256,16 +247,23 @@ def load_telescope(
...
-    bend_dir = None
-    if 'fea_dir' in config:
-        fea_dir = ParseValue(config, 'fea_dir', base, str)
-    if 'bend_dir' in config:
-        bend_dir = ParseValue(config, 'bend_dir', base, str)
 
-    safe = True
+    skip = list(req.keys()) + list(opt.keys()) + list(single.keys()) + ['_get']
     for k, v in config.items():
+        if k in skip:
```
rmjarvis, line 265:
```suggestion
        if k in skip or k[0] == '_':
```
I think would be safer to not put `_get` in the skip list, but rather skip all leading underscore keys here.  The config processing puts other leading underscore values in various places, so depending on how this gets called, there might sometimes be other items to skip here.

imsim/telescope_loader.py
```diff
@@ -256,16 +247,23 @@ def load_telescope(
...
     perturbations = {}
-    fea_dir = None
-    bend_dir = None
-    if 'fea_dir' in config:
-        fea_dir = ParseValue(config, 'fea_dir', base, str)
-    if 'bend_dir' in config:
-        bend_dir = ParseValue(config, 'bend_dir', base, str)
 
-    safe = True
+    skip = list(req.keys()) + list(opt.keys()) + list(single.keys()) + ['_get']
```
rmjarvis, line 263:
```suggestion
    skip = list(req.keys()) + list(opt.keys()) + list(single.keys())
```

rmjarvis (approved):
Just one suggested change.

PR #427, opened 2023-10-06
add option to disable IERS downloads

jchiang87:


cwwalter:
I don't think I see the file:: URL to tell it where to get the local file in this case.  Does it go somewhere else, or not checked in?

jchiang87:
Doing
```
iers.conf.auto_download = False
iers.conf.iers_degraded_accuracy = 'warn'
```
disables the download entirely, so there doesn't need to be any URLs specified and we don't need to keep local copies of the IERS corrections and leap seconds files around.

jchiang87:
Unfortunately, I just found out for the multiprocessing runs, we need to have these lines called for each process, in addition to the main process.  Adding the lines to batoid_wcs.py seems to work.

cwwalter:
Ah OK.  For my understanding: Do having the file around do anything different at all if the calculation is for the future?

cwwalter (approved):
LGTM

jchiang87:
> Do having the file around do anything different at all if the calculation is for the future?

According to the docs, the IERS download data are predictive 1 year into the future.  So if we are computing for a date less than a year into the future, the most recent file would be provide numbers within the IERS accuracy spec.

rmjarvis (approved):
Seems fine to me.  The only thought I had was maybe the default should be True.  I can't imagine we ever need to be maximally up to date on IERS leap seconds.

jchiang87:
@jmeyers314 I added some code to `batoid_wcs.py` to disable the IERS downloads and also to silence those erfa warnings.   Could you have a look please?   Also, I see that batoid_rubin wants to download a `bend.yaml` file to its `data/bend` folder.  This is causing my tests to fail when I run using a shared stack at S3DF.  Those downloads may also be a problem for running on Theta in the batch queues, where remote downloads seemingly aren't allowed, but I guess they happen only in certain circumstances?

jmeyers314:
Looks good to me.

We can manually do the batoid_rubin data download somewhere appropriate instead of letting it try to auto-download.

jchiang87:
Unless I hear otherwise, I'll merge this at 2pm PT today.

imsim/opsim_data.py
```diff
@@ -365,6 +374,10 @@ def OpsimData(config, base, value_type):
     """
     meta = galsim.config.GetInputObj('opsim_data', config, base, 'OpsimData')
 
+    if 'disable_iers_downloads' not in galsim.config.eval_base_variables:
+        galsim.config.eval_base_variables.append('disable_iers_downloads')
+        base['disable_iers_downloads'] = meta.disable_iers_downloads
```
rmjarvis, line 379: I don't think this is the way we want to do this.  You're effectively making our default template (which uses "$disable_iers_downloads") implicitly dependent on OpsimData.  If a user doesn't include that anywhere, then the default template file will be broken in a very hard to diagnose way.

You should either have the template explicitly include this in eval_variables and have both batoid_wcs and opsim_data use that as the source information, or have one of them use a "Current" item to get it from the other one, so at least the user can see the way they are dependent on each other.  E.g. 
```
wcs:
    type: Batoid
    disable_iers_downloads: "@input.opsim_data.disable_iers_downloads"
```
If the latter, then you should add a note that if the user isn't using either of our two forms of opsim_data (skycat or instcat) then this should be set by hand in the wcs section.

rmjarvis: Also, do we really need it in both places?  It seems like the implementation is such that if either one has it, then it will get implemented and the astropy global variables will be set correctly.  So maybe the better solution is just to identify which one needs it to be set correctly first.  (I would expect opsim_data, but I might be wrong.)  Then just do it there and don't bother with having it also done in the other place.

I'll also reiterate my suggestion to make the default True.  I don't think we probably have any use cases where we want to guarantee that these are fully up to date, so letting it always be disabled seems like the better option to me.

jchiang87: The problem is that each process needs to have the download disabled, so if `output.nproc > 1`, the distributed processes, i.e., the ones running batoid for each CCD, would still try to do the download even if it's already been disabled for opsim_data in the head process.

rmjarvis: OK.  Is there a need to allow `disable_iers_downloads=False`?  Why not just hard-code this entirely and always make it disabled when using imsim?

jchiang87: Making `disable_iers_downloads=False` the default seems fine, and I don't personally have a problem with hard-coding it here, but maybe someone would want it enabled in some cases? I really couldn't say.  @jmeyers314 @cwwalter Do you have an opinion?

jmeyers314: I can't imagine why I'd ever care about the precise iers numbers.  This is basically degenerate with a pointing error when aiming the telescope, but probably much smaller.

cwwalter:
I can imagine we want to do something where we recreate real observations after operations start.  

But, as long as the errors are much smaller than the pointing accuracy this should be find.  It would probably be good to have a sense of how big an effect this was.

rmjarvis:
I think there is zero chance that any use case will require the iers updates.  Even trying to reproduce actual observations.  We don't know the actual telescope pointing well enough for this to matter.
But even if I'm wrong, it's trivial for a user to get the update if they really want it.
```
python -c "from astropy.time import Time; Time.now().tai"
```
will download the necessary data if the current data is out of date.  If you want, we can provide that as a standalone script people can use to get the update if they want it.

jmeyers314 (approved):
LGTM.

jchiang87:
Thanks everyone for helping on this one!

PR #428, opened 2023-10-26
workaround for faint objects with zero-valued SEDs at bandpass.effective_wavelength

jchiang87:


rmjarvis (changes requested):
LGTM.  But it would probably be worth writing a short test that draws a couple objects that hit this check and use red_limit (presumably).  It wouldn't surprise me greatly if the ones that used red_limit still have trouble farther along somewhere in the GalSim code.  We may have made a similar assumption that effective_wavelength would work for evaluating somewhere in there.

jchiang87:
@rmjarvis I've added some test code that exercises these changes.

tests/test_stamp.py
```diff
@@ -257,35 +257,35 @@ def new_toplevel_only(self, object_types):
...
-    obj_num = 2610
+    obj_num = 2611
     image, _ = galsim.config.BuildStamp(config, obj_num, logger=logger, do_noise=False)
     print(obj_num, image.center, image.array.shape, image.array.sum())
     assert image.array.shape == (350,350)
     assert 2_400_000 < image.array.sum() < 2_700_000  # 2591010
 
     # 5. Extremely bright star.  Maxes out size at _Nmax.  (And uses fft.)
-    obj_num = 2696
+    obj_num = 2697
```
rmjarvis, line 288: Why did all these obj_nums increase by 1?

rmjarvis (approved):

tests/test_stamp.py
jchiang87: I added a galaxy with the desired properties to the galaxy catalog file.  Since the stars are handled by skyCatalogs after the galaxies, I had to increment each obj_num by 1 in order to have the same stars for these tests.

rmjarvis: Got it.  That makes sense.

PR #433, opened 2023-11-08
Add support for ComCam

jmeyers314:
Adds support for ComCam.

Note, possibly only works on `main` branches of `batoid` and `batoid_rubin` at the moment.  I had a bug in my ComCam optics yaml where both the camera and the camera+telescope were named `ComCam`.  This collision meant when you tried to rotate the camera, you ended up rotating the entire telescope instead.  Will be fixed in a point release soonish...

### DIFF
--- a/imsim/readout.py
+++ b/imsim/readout.py
@@ -9,6 +9,7 @@ import galsim
 from galsim.config import ExtraOutputBuilder, RegisterExtraOutput, GetAllParams
 from lsst.afw import cameraGeom
 import lsst.obs.lsst
+from lsst.obs.lsst.translators.lsst import SIMONYI_TELESCOPE
 from .bleed_trails import bleed_eimage
 from .camera import Camera, get_camera
 from .batoid_wcs import BatoidWCSBuilder
@@ -29,6 +30,18 @@ LSSTCam_filter_map = {'u': 'u_24',
                       'z': 'z_20',
                       'y': 'y_10'}
 
+# There are multiple u, g, and z filters for ComCam.
+# See https://github.com/lsst/obs_lsst/blob/w.2023.26/python/lsst/obs/lsst/filters.py#L352-L361
+# and https://jira.lsstcorp.org/browse/DM-21706
+# I chose the ones that made the most sense to me (JM)
+ComCam_filter_map = {
+    'u': 'u_05',
+    'g': 'g_01',
+    'r': 'r_03',
+    'i': 'i_06',
+    'z': 'z_03',
+    'y': 'y_04',
+}
 
 def make_batoid_wcs(ra0, dec0, rottelpos, obsmjd, band, camera_name,
                     logger=None):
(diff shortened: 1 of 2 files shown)


imsim/readout.py
```diff
@@ -239,13 +252,32 @@ def get_primary_hdu(eimage, lsst_num, camera_name=None,
...
         phdu.header['FILTER'] = band
         phdu.header['TESTTYPE'] = 'IMSIM'
         phdu.header['RAFTNAME'] = raft
         phdu.header['SENSNAME'] = sensor
         phdu.header['RATEL'] = ratel
         phdu.header['DECTEL'] = dectel
+    elif camera_name == 'LsstComCam':
+        phdu.header['FILTER'] = ComCam_filter_map.get(band, None)
+        phdu.header['TELESCOP'] = SIMONYI_TELESCOPE
+        phdu.header['INSTRUME'] = 'ComCam'
```
jchiang87, line 274: Is this the right value for `INSTRUME`?  Based on the convention for LSSTCam and the existing datasets in `/repo/main` at S3DF:
```
[lsstccs@sdfrome001 LSSTCam w_2023_49] butler query-collections /repo/main */raw/all
       Name        Type
------------------ ----
DECam/raw/all      RUN 
HSC/raw/all        RUN 
LATISS/raw/all     RUN 
LSST-TS8/raw/all   RUN 
LSSTCam/raw/all    RUN 
LSSTComCam/raw/all RUN 
```
I would have expected the desired value to be `LSSTComCam`.

jchiang87 (changes requested):
This looks good to me, but we should clarify what the desired the `INSTRUME` keyword value should be.

imsim/readout.py
jchiang87: I checked an existing ComCam raw file in /repo/main, and it does indeed use `ComCam`, so this is ok.

jchiang87 (approved):
lgtm.

rmjarvis (approved):

cwwalter:
Jim, how does this instrument name choice match up with the desire to have a new instrument for the rehearsal with ITL constants?  

Do you want to do this first and then make a new PR for that? Or do you want to have this as the default for when we get real constants and then make a new instrument just for the temporary usage?

jchiang87:
> Jim, how does this instrument name choice match up with the desire to have a new instrument for the rehearsal with ITL constants?

> Do you want to do this first and then make a new PR for that? Or do you want to have this as the default for when we get real constants and then make a new instrument just for the temporary usage?

We'd probably need to iterate with various people on a separate simulated ComCam model, so I'd be inclined to move forward with this, relying for now on the existing LSSTComCam model for testing, then make everything consistent with the final conventions, etc. after that process.   I don't think using `ComCam` here along with the existing model should be a problem as long as we use test repos for these data until things have finalized.

cwwalter:
Great thanks.

PR #435, opened 2023-11-15
Radec wrap

jmeyers314:
Improves instcat filtering of radec near ra=0,360 and the pole.

### DIFF
--- a/imsim/instcat.py
+++ b/imsim/instcat.py
@@ -14,8 +14,32 @@ import galsim
 import pickle
 
 
-def get_radec_limits(wcs, xsize, ysize, logger, edge_pix):
-    """Min and max values for RA, Dec given the wcs."""
+def get_radec_limits(wcs, xsize, ysize, logger, edge_pix, threshold=0.5*galsim.degrees):
+    """Min and max values for RA, Dec given the wcs.
+
+    Parameters
+    ----------
+    wcs: galsim WCS object
+        The WCS for the image.
+    xsize, ysize: int
+        The size of the image.
+    logger: galsim.Logger
+        A logger for logging debug statements.
+    edge_pix: int
+        The number of pixels to allow objects to be off the image.
+    threshold: galsim.Angle
+        The threshold for how close to the pole we are before we don't bother
+        to check the ra range.
+
+    Returns
+    -------
+    min_ra, max_ra, min_dec, max_dec: float
+        The min and max values for RA and Dec.
+    min_x, min_y, max_x, max_y: float
+        The min and max values for x and y on the image.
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -37,14 +37,28 @@ jobs:
...
+
+            - name: Install conda deps b3
+              run: |
+                conda install -y stackvana>=0.2023.32
+
+            # - name: Install conda deps c
+            #   run: |
+            #     conda install -y --file etc/standalone_conda_requirements.txt
+            #     conda info
+            #     conda list
```
rmjarvis, line 61: Why all this?  I think we don't want to have to edit this file whenever we add a new requirement.  I think if we add something to our conda_requirements file, this should automatically get the update.

imsim/instcat.py
```diff
@@ -147,12 +219,15 @@ def __init__(self, file_name, wcs, xsize=4096, ysize=4096, sed_dir=None, edge_pi
...
-                    ra = float(tokens[2])
-                    dec = float(tokens[3])
+                    ra = float(tokens[2])*galsim.degrees
+                    dec = float(tokens[3])*galsim.degrees
                     #logger.debug('object at %s,%s',ra,dec)
-                    if not (min_ra <= ra <= max_ra and min_dec <= dec <= max_dec):
+                    if not (
+                        min_ra <= ra.wrap(ref_ra).deg <= max_ra
+                        and min_dec <= dec.deg <= max_dec
+                    ):
```
rmjarvis, line 228: Maybe cleaner to have get_radec_limits return Angles rather than floats.  Then we don't need this baroque combination of `.deg` applications.  Can do all <= tests with the angles themselves.

rmjarvis (changes requested):

.github/workflows/ci.yml
jmeyers314: This is me trial-and-erroring to fix up the CI.  Still haven't got it, but I'll clean it up after I get to a solution.

imsim/instcat.py
jmeyers314: Yep.  I think that is indeed cleaner.  

Might have been slightly cleaner still if Angle had an `__abs__` method enabling `max(np.abs([angle1, angle2]))` instead of `max(np.abs([angle1.deg, angle2.deg]))*galsim.degrees`.  Should I go add one, or is there a good reason not to have this?

rmjarvis: abs would be reasonable to add.  Just didn't occur to me that it would be useful.

rmjarvis (approved):

PR #436, opened 2023-11-18
stamp handling and config tweaks

jchiang87:
* ~The original issue the stamp handling changes are meant to address were addressed in [GalSim #1258](https://github.com/GalSim-developers/GalSim/pull/1258), but they seem useful to keep regardless.~
* Chris added some fixes to enable the skyCatalogs interface code to use the band set at the config level and to incorporate the band info in the checkpoint file names to enable simulating the same exposure in different bands.

### DIFF
--- a/config/imsim-config-skycat.yaml
+++ b/config/imsim-config-skycat.yaml
@@ -23,7 +23,7 @@ template: imsim-config
 
 input.sky_catalog:
     file_name: default_sky_cat.yaml
-    band: { type: OpsimData, field: band }
+    band: $band
     mjd: { type: OpsimData, field: mjd }
 
 input.opsim_data:
--- a/config/imsim-config.yaml
+++ b/config/imsim-config.yaml
@@ -104,9 +104,10 @@ input:
         dir: checkpoint
         file_name:
           type: FormattedStr
-          format : checkpoint_%08d-%s.hdf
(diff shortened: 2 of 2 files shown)


imsim/stamp.py
```diff
@@ -629,6 +629,12 @@ def draw(self, prof, image, method, offset, config, base, logger):
             # don't draw anything.
             return image
 
+        my_bounds = image.bounds & base['current_image'].bounds
+        if my_bounds.area() == 0:
+            # The stamp for this object doesn't overlap with the
+            # current image, so skip it.
+            return image
```
rmjarvis, line 636: Does this ever happen?  I thought the galsim config code already checked for this.

jchiang87: Yes, it does happen.   I saw right away it while trying to debug the original issue.  Since we are selecting a 100 pixel region around the edge of the CCD for bright stars not actually landing on the CCD, the faint ones like this also get included.

jchiang87: Is  GalSim #1258 going to be part of a release soon?  If so, I would just as soon omit the stamp-related changes and only include Chris' config changes in this PR.

rmjarvis: It's already released as part of 2.5.1.

rmjarvis (approved):

PR #441, opened 2024-01-04
use lsst function to disable implicit threading

jchiang87:


imsim/__init__.py
```diff
@@ -1,4 +1,7 @@
 
+from lsst.utils.threads import disable_implicit_threading
+disable_implicit_threading()
```
rmjarvis, line 3: A quick inline comment about why this is useful is probably appropriate here.
Something like:
```
# We never use threading in imsim, and without this, some systems try to run all the 
# multiprocessing jobs on a single core, which leads to terrible performance.
# We think disabling this is never a problem, and sometimes extremely important.
# It's best if this is done as soon as possible, hence here before any other imports.
```

rmjarvis (approved):

PR #444, opened 2024-01-17
Add airmass dependent Bandpasses

jmeyers314:
Work in progress, looking for feedback.

Adds 
- an `airmass` kwarg to `RubinBandpass`.  Performs linear interpolation of log(throughput) vs airmass X, which seems to be pretty good from investigation in a separate notebook.
- `EnvelopeRatio` photon operator.  This takes an "envelope" (X=1.0) bandpass object, and reads the target (X > 1.0) bandpass function from the base config.  It then sets photon fluxes to zero with probability `1 -  bandpass(wavelength)/envelope(wavelength)`, effectively changing the filter from X=1 to X>1 and still getting the Poisson statistics right.

I'm still not sure how to:
- Ensure that any precomputed skyCatalog fluxes are using the X=1.0 throughputs.
- Coerce galsim.drawImage() to use the envelope bandpass when initially sampling photons behind-the-scenes with WavelengthSampler.

I'll dig in tomorrow, but any advice appreciated.

### DIFF
--- a/config/imsim-config.yaml
+++ b/config/imsim-config.yaml
@@ -133,7 +133,12 @@ image:
 
     det_name: "$det_name"
 
-    bandpass: { type: RubinBandpass, band: "$band" }
+    bandpass:
+        type: RubinBandpass
+        band: $band
+        # You can optionally add an airmass here to incorporate into the bandpass.  If you leave
+        # this out, then the fiducial bandpass with airmass = 1.2 will be used.
+        airmass: { type: OpsimData, field: airmass }
 
     wcs:
         type: Batoid
--- a/config/imsim-config.yaml
+++ b/config/imsim-config.yaml
@@ -139,6 +139,8 @@ image:
         # You can optionally add an airmass here to incorporate into the bandpass.  If you leave
         # this out, then the fiducial bandpass with airmass = 1.2 will be used.
         airmass: { type: OpsimData, field: airmass }
+        camera: "@output.camera"
+        det_name: $det_name
 
     wcs:
         type: Batoid
--- a/imsim/bandpass.py
+++ b/imsim/bandpass.py
@@ -35,7 +35,9 @@ class AtmInterpolator:
             frac = (X - self.Xs[idx-1]) / (self.Xs[idx] - self.Xs[idx-1])
             out = (1-frac)*np.array(self.logarr[idx-1])
             out += frac*self.logarr[idx]
-        return np.exp(out)
+        out = np.exp(out)
+        out[~np.isfinite(out)] = 0.0  # Clean up exp(log(0)) => 0
+        return out
 
 
 def RubinBandpass(band, airmass=None, logger=None):
--- a/imsim/bandpass.py
+++ b/imsim/bandpass.py
@@ -126,7 +126,12 @@ def RubinBandpass(
         )
 
         if camera is not None:
-            old_path = Path(os.getenv("OBS_LSST_DATA_DIR"))
+            try:
+                old_path = Path(os.getenv("OBS_LSST_DATA_DIR"))
+            except TypeError:
+                raise ValueError(
+                    "Unable to find OBS_LSST_DATA; required if using camera or det_name for bandpass."
+                )
             old_path = old_path / camera / "transmission_sensor" / det_name.lower()
             det_files = list(old_path.glob("*.ecsv"))
             if len(det_files) != 1:
--- a/imsim/bandpass.py
+++ b/imsim/bandpass.py
@@ -44,7 +44,7 @@ class AtmInterpolator:
 def RubinBandpass(
     band,
     airmass=None,
-    instrument_name=None,
+    camera=None,
     det_name=None,
     logger=None
 ):
@@ -59,8 +59,8 @@ def RubinBandpass(
     airmass : `float`, optional
         The airmass at which to evaluate the bandpass.  If None, use the
         standard X=1.2 bandpass.
-    instrument_name : `str`, optional
-        Name of the instrument for which to incorporate the detector-specific
+    camera : `str`, optional
+        Name of the camera for which to incorporate the detector-specific
         quantum efficiency.  If None, use the standard hardware bandpass.
     det_name : `str`, optional
         Name of the detector for which to incorporate the quantum efficiency.  If
@@ -68,10 +68,17 @@ def RubinBandpass(
     logger : logging.Logger
         If provided, a logger for logging debug statements.
     """
-    if sum([instrument_name is not None, det_name is not None]) == 1:
-        raise ValueError("Must provide both instrument_name and det_name if using one.")
+    if sum([camera is not None, det_name is not None]) == 1:
+        raise ValueError("Must provide both camera and det_name if using one.")
+    match camera:
+        case 'LsstCam':
+            camera = 'lsstCam'
+        case 'LsstComCam':
+            camera = 'comCamSim'
+        case _:
+            camera = camera
     tp_path = Path(os.getenv("RUBIN_SIM_DATA_DIR")) / "throughputs"
-    if airmass is None and instrument_name is None:
+    if airmass is None and camera is None:
         file_name = tp_path / "baseline" / f"total_{band}.dat"
         if not file_name.is_file():
             logger = galsim.config.LoggerWrapper(logger)
@@ -100,9 +107,9 @@ def RubinBandpass(
             wave_type='nm'
         )
 
-        if instrument_name is not None:
+        if camera is not None:
             old_path = Path(os.getenv("OBS_LSST_DATA_DIR"))
-            old_path = old_path / instrument_name / "transmission_sensor" / det_name.lower()
+            old_path = old_path / camera / "transmission_sensor" / det_name.lower()
             det_files = list(old_path.glob("*.ecsv"))
             if len(det_files) != 1:
                 raise ValueError(f"Expected 1 detector file, found {len(det_files)}")
@@ -173,7 +180,7 @@ class RubinBandpassBuilder(galsim.config.BandpassBuilder):
         req = { 'band' : str }
         opt = {
             'airmass' : float,
-            'instrument_name' : str,
+            'camera' : str,
             'det_name' : str
         }
         kwargs, safe = galsim.config.GetAllParams(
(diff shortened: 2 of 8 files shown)


jmeyers314:
Thinking more about how to coerce `.drawImage` to use the X=1.0 "envelope" bandpass instead of the actual `@image.bandpass` when using the `EnvelopeRatio` photon operator: we could place that logic directly into `LSST_SiliconBuilder`.  That'd make this work for our most important case I think, but errors might creep in if using a different stamp type.

Alternatively, we could refactor to require that `image.bandpass` be the X=1 envelope, and only build the X>1 actual bandpass when configuring `LSST_SiliconBuilder`.  That'd have the downside though that image.bandpass is slightly a lie.

@rmjarvis , do you have an opinion here?

cwwalter:
Sorry have been busy with first trip to summit after moving.  @rmjarvis did you see Josh's question above?

rmjarvis:
I thought you would just specify the config bandpass as being the one without the airmass factor and let the effective bandpass with the atmosphere be emergent from the combination of that with the EnvelopeRatio.  But I admit I haven't thought about this enough to understand why that doesn't work for wavelength sampling as Josh implied above.

cwwalter:
Hi Josh,

What's your current status on this and what do you still need help/input on?

I think we wanted to start the rehearsal production really soon so we should figure out how to do it:

You asked above for advice on how to:

- Ensure that any precomputed skyCatalog fluxes are using the X=1.0 throughputs.
- Coerce galsim.drawImage() to use the envelope bandpass when initially sampling photons behind-the-scenes with WavelengthSampler.

Did you get any out of this issue?

jmeyers314:
Jim, Eli and I chatted a bit about this last Friday.  I think we decided that instead of initially generating photons from the outer envelope bandpass and then probabilistically culling them to match the realized bandpass, we'd instead generate from the good 'ole X=1.2 throughputs and just modify photon fluxes (weights) to get to the appropriate bandpass.  That's much easier than trying to coordinate envelope+realized bandpass.  The downside is the Poisson statistics don't end up quite right, but since the airmass and QE variations are pretty small, I bet we can live with that.  (And it's still better than what we had before).

I can probably deliver the above by ~tomorrow.

cwwalter:
Intersting.  So, the interpolation would return a weights table?

I thought I remembered the Galsim photon list can take a weight for entry, but can't find that now.  Will GalSim itself apply the weight?

jmeyers314:
Okay, I think this one is ready to look at now.  Tests pass for me locally (with branch tickets/DM-42513 of obs_lsst_data).

jmeyers314:
Okay, looks like the one CI failure is where it can't find the file that lives on an unmerged branch of obs_lsst_data.  Good.

imsim/bandpass.py
```diff
@@ -48,9 +178,22 @@ def buildBandpass(self, config, base, logger):
...
+            'det_name' : str
+        }
+        kwargs, safe = galsim.config.GetAllParams(
+            config, base, req=req, opt=opt
+        )
         kwargs['logger'] = logger
         bp = RubinBandpass(**kwargs)
+
+        # Also, store the kwargs=None version in the base config.
+        if 'airmass' in kwargs:
```
jchiang87, line 193: Should this also check for `camera` in `kwargs`?

config/imsim-config.yaml
```diff
@@ -133,7 +133,12 @@ image:
 
     det_name: "$det_name"
 
-    bandpass: { type: RubinBandpass, band: "$band" }
+    bandpass:
+        type: RubinBandpass
+        band: $band
+        # You can optionally add an airmass here to incorporate into the bandpass.  If you leave
+        # this out, then the fiducial bandpass with airmass = 1.2 will be used.
+        airmass: { type: OpsimData, field: airmass }
```
jchiang87, line 141: Does using this config as-is mean that the default QE curve (as captured in `baseline/total_{band}.dat`) will be used?  If we want to use the per-CCD QE curves, what would the corresponding entries for `camera` and `det_name` be?  I guess it would be something like
```
camera: "$output.camera"
det_name: "$det_name"
```
?

tests/test_stamp.py
```diff
@@ -449,6 +451,126 @@ def new_toplevel_only(self, object_types):
...
+            print(f"{ref_flux = }")
+            print()
+
+            ref_fluxes.append(ref_flux)
+            realized_fluxes.append(realized_flux)
+
+        print("\n"*3)
+        return np.array(ref_fluxes), np.array(realized_fluxes)
+
+    ref_QE, realized_QE = get_fluxes(None, 'LsstComCam', 'R22_S11')
```
jchiang87, line 542: Could also just use `LsstCam` here and since those data are already in `obs_lsst_data`.  Note that DM-40164--the ticket with the updated QE curves for LSSTCam-- still isn't merged, as of this writing, but will be soon.

jchiang87 (changes requested):
Thanks, Josh.  I have some minor comments/requests having to do with handling the per-CCD QE curve side of things, otherwise lgtm.

config/imsim-config.yaml
jmeyers314: I think something like that will work (though maybe it's `@output.camera`?)  I'll double check.  Do you think we should make this (+ using the opsim airmass) the default?

imsim/bandpass.py
jmeyers314: Yep.  But it's even easier and more future proof to just skip the check entirely always, computing the fiducial_bandpass (can't imagine this is actually a time sink).

tests/test_stamp.py
jmeyers314: Of course.  Hopefully that'll get this through CI.

jmeyers314: > Hopefully that'll get this through CI.

Nope, not quite.  test_object_positions refers to ComCam and now fails b/c I added `camera` to imsim-config.yaml.  But getting closer...

config/imsim-config.yaml
cwwalter: Yes, I think so.  We should always be specify the airmass.  I double checked that the instanceCat interface also calculates the airmass.  So, assuming it gets correctly into the opsim entry it should be good everywhere.

jmeyers314: Okay, I fiddled with some of the tests to make this work, but I think it's okay now.  (Apparently the throughput of the real LsstCam R22_S11 is better than what's in the throughputs repo).

tests/test_stamp.py
jmeyers314: Fiddled with test_object_positions to not use non-existent LsstCamImSim QE curves.

jmeyers314:
@jchiang87 , I think I've addressed all your comments now.  Let me know if there's anything else I should look at.

@rmjarvis , could you take a peek at this, especially the tests to make sure they address some of the concerns you raised at the last imsim technical meeting?

jchiang87 (approved):
Thanks, Josh.  LGTM.

tests/test_stamp.py
```diff
@@ -366,15 +368,15 @@ def new_toplevel_only(self, object_types):
...
+    assert 230_000 < image.array.sum() < 260_000  # 241_136
 
     # There is also a code path where the xsize,ysize is dictated by the calling routine.
     del config['stamp']['size']
     image, _ = galsim.config.BuildStamp(config, obj_num, logger=logger, do_noise=False,
                                         xsize=128, ysize=128)
     print(obj_num, image.center, image.array.shape, image.array.sum())
     assert image.array.shape == (128,128)
-    assert 290_000 < image.array.sum() < 320_000  # 306675
+    assert 320_000 < image.array.sum() < 350_000  # 338_265
```
rmjarvis, line 379: These all changed because of something in the sims code I suppose?  New version?

tests/test_stamp.py
```diff
@@ -446,7 +448,127 @@ def new_toplevel_only(self, object_types):
     image, _ = galsim.config.BuildStamp(config, obj_num, logger=logger, do_noise=False)
     print(obj_num, image.center, image.array.shape, image.array.sum())
     assert image.array.shape == (32, 32)
-    assert 5 < image.array.sum() < 10  # 8
+    assert 7 < image.array.sum() < 13  # 10
```
rmjarvis, line 451: How about this?  Do you understand why this changed?

rmjarvis (approved):
I have a few questions, but this looks fine to me too.

tests/test_stamp.py
jmeyers314: Not the sims code, this is b/c I added kwargs to imsim-config.yaml to use the measured QE curves in the obs_lsst_data repo instead of the fiducial QE curve from the throughputs repo.  Here's the difference (for R22):
qe

tests/test_stamp.py
jmeyers314: I think same as above: real QE measurements are better than the baseline.

imsim/bandpass.py
```diff
@@ -15,24 +74,113 @@ def RubinBandpass(band, logger=None):
...
+        tput = interpolator(airmass)
+        bp_atm = galsim.Bandpass(
+            galsim.LookupTable(
+                wave, tput, interpolant='linear'
+            ),
+            wave_type='nm'
+        )
+
+        if camera is not None:
+            old_path = Path(os.getenv("OBS_LSST_DATA_DIR"))
```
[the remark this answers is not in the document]
jmeyers314: Turns out `Path(None)` raises a TypeError.  I think it's okay to panic here; user asked for a specific detector bandpass but we're unable to comply.  Reraised a ValueError with a more helpful message.

imsim/bandpass.py
```diff
@@ -15,24 +74,113 @@ def RubinBandpass(band, logger=None):
...
-    sims_dir = os.getenv("RUBIN_SIM_DATA_DIR")
-    file_name = os.path.join(sims_dir, "throughputs", "baseline", f"total_{band}.dat")
-    if not os.path.isfile(file_name):
-        # If the user doesn't have the RUBIN_SIM_DATA_DIR defined, or if they don't have
-        # the correct files installed, revert to the GalSim files.
-        logger = galsim.config.LoggerWrapper(logger)
-        logger.warning("Warning: Using the old bandpass files from GalSim, not lsst.sims")
-        file_name = f"LSST_{band}.dat"
-    bp = galsim.Bandpass(file_name, wave_type='nm')
+    if sum([camera is not None, det_name is not None]) == 1:
```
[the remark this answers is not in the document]
jmeyers314: Yeah, I like that.  Shortened it to `if (camera is None) != (det_name is None):`.

PR #446, opened 2024-02-02
Fix negative pixels that could happen from bleed trails

rmjarvis:
Fix a bug in the bleed trail algorithm that would subtract too much from excess charge when dealing with off-the-bottom pixels and less than a full well of charge was left to deal with.


jchiang87 (approved):
Thanks!  LGTM.

imsim/lsst_image.py
```diff
@@ -293,11 +294,15 @@ def addNoise(self, image, config, base, image_num, obj_num, current_var, logger)
             camera = get_camera(self.camera_name)
             det_name = base['det_name']
             serial_number = camera[det_name].getSerial()
+            # Note: the regular Python hash function is non-derterministic, which is not good.
```
erykoff, line 297: non-derterministic -> non-deterministic

rmjarvis:
```suggestion
            # Note: the regular Python hash function is non-deterministic, which is not good.
```

jchiang87:
The most recent set of changes look fine to me.  Unless I hear otherwise, I'll go ahead and merge this at 5pm PT today.

PR #448, opened 2024-02-15
add LsstComCamSim as a camera option

jchiang87:
* Include file with reduced strength tree rings for ComCam simulations.  All CCDs have `B` parameters in the weaker regime specified in the [build_tree_ring_file.py](data/tree_ring_data/build_tree_ring_file.py) script.
* Update the `OBSID` and `CONTRLLR` header keywords to follow expected Rubin conventions.


imsim/readout.py
```diff
@@ -268,16 +267,21 @@ def get_primary_hdu(eimage, lsst_num, camera_name=None,
         phdu.header['SENSNAME'] = sensor
         phdu.header['RATEL'] = ratel
         phdu.header['DECTEL'] = dectel
-    elif camera_name == 'LsstComCam':
+        telcode = 'MC'
+    elif camera_name in ('LsstComCam', 'LsstComCamSim') :
```
jchiang87, line 271: I'm wondering if we should omit `LsstComCam` as an option since the DM folks would like to distinguish between simulated and real data by the instrument model.  @jmeyers314 Any thoughts?

jmeyers314: That seems like a reasonable default position to me.  We can revisit if it breaks any workflows later.  Presumably the same applies to `LsstCam` / `LsstCamSim` ?

jchiang87: Thanks, I'll remove the `LsstComCam` option then, and once I find time to make a `LsstCamSim` instrument model, I'll replace `LsstCam` similarly. BTW, feel free to comment and/or review this PR. I just picked on Chris because he's familiar with the header and tree ring discussions.

imsim/readout.py
```diff
@@ -286,6 +287,11 @@ def get_primary_hdu(eimage, lsst_num, camera_name=None,
         phdu.header['RA'] = ratel
         phdu.header['DEC'] = dectel
         phdu.header['ROTCOORD'] = 'sky'
+        telcode = 'MC'
+    dayobs = eimage.header['DAYOBS']
+    seqnum = eimage.header['SEQNUM']
+    contrllr = eimage.header['CONTRLLR']
+    phdu.header['OBSID'] = f"{telcode}_{contrllr}_{dayobs}_{seqnum:06d}"
```
jmeyers314, line 294: I'm surprised `{dayobs}` isn't `{dayobs:03d}`, but since this was working before I assume that's right.

jmeyers314 (approved):
LGTM.

imsim/readout.py
jmeyers314: nvm.  it's always 8 digits isn't it.  (not 3!)

PR #449, opened 2024-02-28
Add ComCam vignetting

jmeyers314:
I updated the files in devel-tools/vignetting to handle both LsstCam and ComCam.  I made one algorithmic tweak, which was to reflect the input (radius, flux) data about the radius=0 axis so when fitting the spline, the symmetry of the input data and knots forces the spline to have zero derivative at radius=0.

ComCam and (updated) LsstCam pngs below:

LSSTCam_vignetting_profile_fit
LSSTComCamSim_vignetting_profile_fit

To use the new vignetting profile for ComCam, update your yaml config with:
```yaml
input:
  vignetting:
    file_name: LSSTComCamSim_vignetting_data.json
```


devel-tools/vignetting/README.md
```diff
@@ -14,10 +15,12 @@ Instructions for generating vignetting spline fits.
    photons per star.  Currently, `magnorm=22.25` will produce ~100k
    incident photons per star.
 
-4. To run the simulation run `galsim imsim-vignetting.yaml`.
+4. To run the simulation run `galsim imsim-vignetting.yaml`.  For
+ComCam, you can add `eval_variables.bisComCam=True`.
```
jchiang87, line 19: What does adding this do?

jchiang87 (approved):
Look good.  I do have one question which may warrant more explanation in the README.

devel-tools/vignetting/README.md
jmeyers314: It sets [this](https://github.com/LSSTDESC/imSim/blob/comcam_vignetting/devel-tools/vignetting/imsim-vignetting.yaml#L17) eval variable, which is then used in the yaml file to select between comcam/lsstcam settings in a few places.  E.g., [here](https://github.com/LSSTDESC/imSim/blob/comcam_vignetting/devel-tools/vignetting/imsim-vignetting.yaml#L30).  

The `b` in `bisComCam` (for boolean) is unfortunate.  (Would have been nice to just have `isComCam` which is more readable.). But that's a limitation of the galsim config system.  I can add a bit more text to the README.

jmeyers314: Maybe `bIsComCam` or `bUseComCam` is better?

jmeyers314: Okay, I updated the README text and went with `bIsComCam`.

PR #451, opened 2024-03-01
Add sensor height map to batoid telescope model

jmeyers314:
Adds sensor height map data to batoid telescope model.  The height data comes from https://ls.st/doc-37242.  The more detailed measurements there are fit, for each sensor, to a Zernike model which I evaluate here.  Also included is a script to make a focal plane height map for comparison to the docushare document.  Here's the actual comparison:

From Doc-37242
Fig 21

From imsim script:

fp_height_map

There's enough similarity to convince me I got the signs and coordinates right, even if there are some offsets too (though note, it was hard to match colorbars).

Finally, I had to tweak `batoid` a bit to make this, so it won't pass CI until that's updated.  I wanted to make the PR while this was fresh in my memory though.


cwwalter:
Do we have the info to do the same for the ComCam model?

aaronroodman:
Ask Andy Rasmussen, he should be able to locate height data if this was collected for the ComCam raft

_________________________________________________________________________
 Prof. Aaron Roodman
 Deputy Director Vera C. Rubin Observatory
 Dept. of Particle Physics & Astrophysics
 Kavli Institute for Particle Astrophysics & Cosmology
 SLAC National Accelerator Laboratory
 Stanford University


 SLAC National Accelerator Laboratory           E-mail: ***@***.******@***.***>
 2575 Sand Hill Rd.                                          Phone:  650-926-2705
 MS 29
 Menlo Park, CA 94025                                    URL: http://www.slac.stanford.edu/~roodman
_________________________________________________________________________



From: Chris Walter ***@***.***>
Date: Thursday, March 7, 2024 at 8:15 AM
To: LSSTDESC/imSim ***@***.***>
Cc: Subscribed ***@***.***>
Subject: Re: [LSSTDESC/imSim] Add sensor height map to batoid telescope model (PR #451)

Do we have the info to do the same for the ComCam model?

cwwalter:
@arasmssn This is the GitHub issue I asked you about.

arasmssn:
@jmeyers314 I think something's wonky in the trascription between the focal plane sensor fit zernike coefficients. First of all it looks almost like there's a vignetting pattern that drops off at field positions >1.75deg or so that makes the surface height of the sensors drop off. I'll need to check these maps against the actual in-situ measurements for which we also didn't have full reach because of the displacement sensor's plane of reflection.
I'll post other problems I think I'm seeing but need to be sure first.

PR #454, opened 2024-03-18
Unpin rubin-sim in conda reqs

jmeyers314:


cwwalter (approved):
LGTM.. but do we want to pin it at 2.0.0 or leave it free?

jchiang87:
I think we want to pin it at some version to avoid problems with the unit tests, especially the sky model tests.   That's why it was pinned at 1.1.1.

rmjarvis:
If possible, we should try to leave it unpinned and only pin it again if there are problems.  People in the wild will continue to have whatever the latest is, so we will want to know when that is not compatible with how we are using it.

jmeyers314:
I don't understand.  If I `conda install ...` packages one at a time, things work.  If I `conda install --file standalone_conda_requirements.txt"` the same packages, then conda stdouts that it's done, but returns a non-zero exit status anyway so GHA thinks error.  Gonna try with ` || true`...

jmeyers314:
Seemed to get farther?  But similar error when installing the test requirements?  adding more `|| true`...

jmeyers314:
Okay, that seems to have worked (rebasing now to clean up the history).  Note I fixed a deprecation warning too in case anyone wants to review that.

rmjarvis (approved):

cwwalter:
The pickle to fits change of the fringing file is for the deprecation warning?

jmeyers314:
> The pickle to fits change of the fringing file is for the deprecation warning?

Exactly.  The pickle was holding a deprecated interpolation object, so I extracted the values-to-be-interpolated and put them in a fits instead.

Issue #452, opened 2024-03-13
Ensuring consistent yaml configs for specifying the Camera to use


jchiang87:
Currently, to switch from `LsstCam` to `LsstComCamSim`, (at least?) three entries in the yaml config need to be changed:
* The input camera to use for reading the focal plane geometry and per-amp parameters from `obs_lsst`:
```
input.camera:  LsstComCamSim
```
* The tree rings file to use
```
input.tree_rings.file_name: LsstComCamSim_tree_rings_2024-02-03.txt
```
* The optics model to use
```
eval_variables.bIsComCam: True
```
These can all be set independently to inconsistent values and simulations will run to completion without any warnings (though an info message regarding the tree rings file used will be emitted).

It would be great if all of this could be handled with a single config parameter, i.e.,
```
input.camera: LsstComCamSim
```

jmeyers314:
Looking through imsim-config.yaml, I see:

```yaml
output.camera: LsstCam
```

```yaml
output.readout.bias_levels_file: LSSTCam_bias_levels_run_13421.json
```

```yaml
input.vignetting.file_name: LSSTCam_vignetting_data.json
```

```yaml
input.telescope.file_name: $f"LSST_{band}.yaml"
```

and potentially stuff like

```yaml
output.nfiles: 189 # 9 for comcam
```

Some ideas on how to handle:

1
--

In galsim-config-language something like:

```yaml
input.telescope.file_name: $ f"LSST_{band}.yaml" if @output.cameraName == 'LsstCam' else f"ComCam_{band}.yaml"
```

though that doesn't really scale past 2 cameras (maybe that's enough?).

2
--

Or, in imsim-config.yaml at least, shove all the camera-dependent stuff into various `eval_variables` and have one section uncommented for ComCam and another for LsstCam:

```yaml
eval_variables:
    sband: r
    # For LSSTCam
    scameraName: LsstCam
    sopticsFileName: LSST_{band}.yaml
    inumdet: 189
    # # For ComCam
    # scameraName: LsstComCamSim
    # sopticsFileName: ComCam_{band}.yaml
    # inumdet: 9
```

That keeps everything related to the camera choice localized at least.

3
--

Or there's the approach in https://github.com/LSSTDESC/imSim/blob/main/devel-tools/vignetting/imsim-vignetting.yaml where I made a boolean config variable and used a bunch of `$ ... if IsComCam else ...` lower in the config.
[...]

PR #456, opened 2024-03-22
Add template configs for simulating ComCam

jchiang87:
This addresses #452


cwwalter:
During the last technical meeting, the idea that we could use a dictionary of all the camera dependent items that classes drew from instead was also suggested as an alternative approach to solving this problem.

rmjarvis (approved):

PR #459, opened 2024-04-05
Use a template dict for all the camera-specific info

rmjarvis:
This is a proposal for IMO a neater way to handle the camera-specific information in the config file, rather than #456, which is currently an open PR.  It reads in a config dict into a camera_info dictionary that other items can use.  So the user only needs to update output.camera to get all the appropriate files for that camera.  This seems to be what at least some users were expecting to happen when they changed that, so it feels like an API that would be intuitive.

Unfortunately, it is currently blocked by a GalSim bug, which has been fixed [#1281](https://github.com/GalSim-developers/GalSim/pull/1281), but will need to flow through to a bugfix release before this PR will work.  But I think OK for people to take a look and comment on the design pattern already.

The PR is directed into u/jchiang/ComCam_configs to more easily see the differences with that one.  If this is approved, we should merge this first, then PR #456.


jmeyers314:
Did you forget to commit the `camera + '_info.yaml` files possibly?

rmjarvis:
Yes. Thanks!  (The peril of relying on git add -p.)

jchiang87 (changes requested):
A couple changes are needed for ComCam, but otherwise things look fine.

rmjarvis:
OK.  Fixed those items.  Thanks.

jchiang87 (approved):
Thanks for the changes, Mike.

PR #463, opened 2024-04-17
Use the minimum of image.nobjects and the number of objects available to be rendered. 

jchiang87:
Mike's suggested fix seems to work.  However, I don't know how to write an isolated unit test for this.  The only way I can see to test it would be to run a full simulation and checking the outputs.  Even running with just a couple of objects takes a long time.  Suggestions welcome.

### DIFF
--- a/imsim/lsst_image.py
+++ b/imsim/lsst_image.py
@@ -37,13 +37,15 @@ class LSST_ImageBuilder(ScatteredImageBuilder):
         logger.debug('image %d: Building LSST_Image: image, obj = %d,%d',
                      image_num,image_num,obj_num)
 
-        if 'sky_catalog' in base['input'].keys():
-            nobjects = LoadInputObj(base, 'sky_catalog').getNObjects()
-        elif 'instance_catalog' in base['input'].keys():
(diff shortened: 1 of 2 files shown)


rmjarvis:
I think you can use the truth catalog output to test.
Case 1:
instcat with say 10 objects, and output.nobjects=3 -> check that there are 3 items in the output truth catalog.
Case 2:
instcat with 3 objects and output.nobjects=10 -> check that there are again 3 items in the truth catalo
Case 3:
instcat with 3 objects and no output.nobjects specified -> same output as case 2.
And if the first 3 objects in case 1 are the same as the 3 in cases 2,3, all three output files should be identical.  Both the truth catalog and image.

jchiang87:
Sure, but I think that still involves running full simulations.  I was hoping there would be something that could be done to test the setting of `LSST_ImageBuilder.nobjects` directly for the different yaml inputs.

rmjarvis:
There are ways to do that with mock, but I always prefer just running a full integration test if possible.  You can make the fluxes quite small, so the run time could be very quick.

rmjarvis (approved):

Issue #464, opened 2024-04-22
instance catalog read twice


esheldon:
I'm seeing that instance catalogs are read twice.

When galaxies are present its taking about 20minutes to read the catalog, so doubling this is significant.

Below I added a timer to show the read time.  You can see that it happens twice
```
Reading config file gauss-nodcr-01.yaml
Overriding default configuration file with /gpfs/mnt/gpfs02/astro/desdata/esheldon/mambaforge/envs/imsim/share/eups/Linux64/dustmaps_cachedata/gbb0a0c949e+81bc2a20b4/config/.dustmapsrc
Reading visit info from instance catalog 01032304/instcat.txt
Done reading meta information from instance catalog
Reading instance catalog 01032304/instcat.txt
49884838it [22:37, 36749.65it/s]
Total objects in file = 49884816
Found 199874 objects potentially on image
Reading instance catalog 01032304/instcat.txt
49884838it [22:25, 36947.85it/s]
...snip...
```

rmjarvis:
I think the intended use of instance catalogs was to have only the objects that will appear on the image (or not too far off of the image) in each file.  So reading that twice isn't so much overhead.  I don't remember off hand the design constraints that led us to have it read twice, but it was probably just that the way we have been using the instance catalogs in the past, it wasn't much of a tall pole.  I'm sure it's possible to have it only read once, but it's not a trivial change, I don't think.

esheldon:
I'm using the instcats that were used in real production I believe.  These cover the full focal plane

esheldon:
Could we put an `@lru_cache` on the function?

rmjarvis:
We don't use instcats anymore in production.  We use SkyCatalogs.

jchiang87:
I think the initial read is simply to get the number of objects for galsim to set seeds etc. for multiprocessing.  We got around this problem with skyCatalogs by adding an `approx_nobjects` option to the `input.sky_catalog` yaml section: https://github.com/LSSTDESC/imSim/blob/main/imsim/skycat.py#L67
We could add the same option for the instance catalog code to avoid reading in the entire file.

rmjarvis:
Ah, that makes sense.  Yeah, we could add that for instcat too.  That would be pretty straightforward.

PR #466, opened 2024-04-23
Delay read of instcat until data is needed

rmjarvis:
Address #464 take two.

I removed the approx_nobject keyword, and now implement getApproxNObjects() by just counting the lines in the instance catalog file.  This will always be >= the number of objects in that file that fall on the image, so it works as a good approx value.  And it's quite fast, so should address the original problem about reading the file twice taking a long time.


esheldon:
Now the catalog is read once for a single CCD

However it is still reading it again for each CCD that is simulated.

rmjarvis:
> However it is still reading it again for each CCD that is simulated.

Yes, it selects the galaxies in the input catalog that will fall on (or near) the CCD, so that is required.

esheldon (approved):

jmeyers314 (approved):
LGTM.

PR #467, opened 2024-05-04
Enable test multiprocessing

jmeyers314:
test_psf.py had a race condition where one test could remove another tests's test directory before that test was finished.  I added some tempfile logic to avoid this.


rmjarvis (approved):
lgtm

PR #469, opened 2024-05-10
Simplify rotSkyPos calculation

jmeyers314:
Updates readout.py:compute_rotSkyPos.

This routine had previously instantiated a complete WCS, marched separate test points a little north, and a little towards the CCD +Y to determine RotSkyPos empirically for insertion into the image header.  The new routine follows notes I made at https://smtn-019.lsst.io/v/DM-44258/index.html, developing something closer to an analytic formula for determining RotSkyPos.  The key insight is that RotSkyPos is intended to relate the camera orientation to _ICRF north_ and not true north.  Previous attempts I've seen at analytic relations use the telescope rotator angle (rotTelPos) and the parallactic angle, the latter being the position angle of zenith measured from true North (not ICRF North). This is the wrong north! So I invented the pseudo-parallactic angle which just swaps true north -> ICRF north.  With that, plus acknowledging a few realities about how telescopes flip images around and other coordinate conventions, the analytic formula for Rubin becomes:

RotSkyPos = 270 - RotTelPos + pseudo-parallactic-angle.

I also clarified the documentation a bit, and fixed one line where the MJD was being interpreted in the (default) UTC time scale instead of the TAI scale.

Note that in the end, I've concluded we _didn't_ make a mistake in simulating OR3.  We were correct to use the regular parallactic angle inside batoid_wcs since the place where we use it relates true-of-date "observed" ra/dec coordinates with altitude and azimuth.  A separate transformation handles true north <-> ICRF north, so we didn't need the pseudo parallactic angle for creating the WCS.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -54,14 +54,25 @@ jobs:
                 pip install -e .
                 cd ..
 
+            - name: Cache data
+              uses: actions/cache@v4
+              with:
+                path: rubin_sim_data
+                key: "2024-05-10"  # Update the key if we ever change the data.
+
             - name: Install rubin_sim_data
               run: |
-                mkdir rubin_sim_data
-                mkdir rubin_sim_data/sims_sed_library
-                # Just get the skybrightness, throughputs, and SED data for now.
-                curl https://s3df.slac.stanford.edu/groups/rubin/static/sim-data/rubin_sim_data/skybrightness_may_2021.tgz | tar -C rubin_sim_data -xz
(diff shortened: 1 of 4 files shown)


jchiang87 (approved):
lgtm, though I'm wondering how we should test this.  For the older implementation, I simulated a handful of CCDs around the periphery of the focal plane for a couple visits, and ran single frame processing through `calibrate` to check the astrometry.  It would be good to add some tests within the scope of this package.

jmeyers314:
New code is still subject to the old test: https://github.com/LSSTDESC/imSim/blob/main/tests/test_readout.py#L198

If you think we need more than this though, I can probably figure out how to do a one-off calibrate.

jchiang87:
Ah, I forgot about that test (even though I wrote it).  I guess it's fine, for now. It would be good to get rid of `LsstCamImSim` entirely, but that doesn't have to happen here.

Issue #453, opened 2024-03-13
Sky background calculation using rubin_sim.skyBrightness SEDs should not include atmospheric absorption


jchiang87:
There's a thread in [slack](https://lsstc.slack.com/archives/C01JTSDV7CH/p1710367626198229?thread_ts=1710347138.892389&cid=C01JTSDV7CH) where Lynne Jones mentions this and references [SMTN-002](https://smtn-002.lsst.io/#skynote), particularly footnote 1.

cwwalter:
I'm just putting this here for future reference in case anyone is confused by why this is an issue or needs a reminder of the issue.  As Jim says this is explained in the tech note above.

Fundamentally the sky model comes from a model from ESO which is measured light coming *from* the sky including airglow etc. So the light is leaving the atmosphere and going into the pupil. Therefore it doesn't make sense to include the extinction going *through* the atmosphere.

jchiang87:
Fixed in #470

PR #470, opened 2024-05-13
Use hardware bandpass for sky level

jchiang87:
This addresses #453

### DIFF
--- a/imsim/bandpass.py
+++ b/imsim/bandpass.py
@@ -96,6 +96,7 @@ def RubinBandpass(
         case _:
             camera = camera
     tp_path = Path(os.getenv("RUBIN_SIM_DATA_DIR")) / "throughputs"
+    bp_hardware = None
     if airmass is None and camera is None:
         file_name = tp_path / "baseline" / f"total_{band}.dat"
         if not file_name.is_file():
@@ -183,6 +184,12 @@ def RubinBandpass(
     bp = bp.truncate(relative_throughput=1.e-3)
     bp = bp.thin()
     bp = bp.withZeropoint('AB')
+    if bp_hardware is not None:
+        bp_hardware.truncate(relative_throughput=1.e-3)
+        bp_hardware.thin()
+        bp_hardware.withZeropoint('AB')
+        bp.bp_hardware = bp_hardware
+
     return bp
 
 
--- a/imsim/sky_model.py
+++ b/imsim/sky_model.py
@@ -9,7 +9,8 @@ from scipy.interpolate import RegularGridInterpolator, RectBivariateSpline
 import os
 from .meta_data import data_dir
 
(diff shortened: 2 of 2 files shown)


jmeyers314 (commented):
Generally looks good, but I'm wondering about the best way to set RUBIN_AREA.  From the default config [here](https://github.com/LSSTDESC/imSim/blob/main/config/imsim-config.yaml#L263-L264), photons are initially sampled from an annulus with an effective diameter of d = 2 * sqrt(4.18**2 - 2.55*2) = 703 cm, not 642 or 649 cm.  I'm wondering if that later figure is some kind of FOV averaged effective diameter that integrates the vignetting.  Since the vignetting is emergent, I'd have guessed we want to use the 703 cm figure for determining the initial flux.

jchiang87:
Based on the description in [SMTN-002](https://smtn-002.lsst.io/#source-counts), it looks like we should be using 703 cm.

jmeyers314:
Yep, I agree.  I wonder what the best way to compute this programmatically is...  We could put the inner/outer diameter into `eval_variables`, and then wire those up consistently to both `PupilAnnulusSampler` and `SkyModel`([here]( https://github.com/LSSTDESC/imSim/blob/main/imsim/sky_model.py#L247)) maybe?

Using 703 cm as the default seems reasonable to me.  (I suspect the median-over-FOV value is closer to this than to 642).

jchiang87:
> an effective diameter of d = 2 * sqrt(4.18**2 - 2.55*2) = 703 cm, not 642 or 649 cm. 

There's a typo here.  Using `d = 2 * sqrt(4.18**2 - 2.55**2)`, I get `d = 662.4` cm.

jchiang87:
I'll also fix the pupil area in the skycat.py code in this PR.

jchiang87:
@jmeyers314 I've hooked up everything to use the `eval_variables` for the pupil annulus radii.  Can you have another look?

config/imsim-config.yaml
```diff
@@ -39,6 +39,9 @@ eval_variables:
         type: Degrees
         theta: { type: OpsimData, field: rotTelPos }
 
+    fpupil_R_outer: 4.18
+    fpupil_R_inner: 2.55  # M1 inner diameter is 2.558, but we need a bit of slack for off-axis rays
+
```
jmeyers314, line 44: Maybe just put `fpupil_area` here once rather than duplicating the code?

jmeyers314 (approved):
LGMT.  One last comment/question.

config/imsim-config.yaml
jchiang87: Sure, I had thought that we might want to access those values individually, at some point, but the skycat and instcat implementations indeed probably don't need that.

PR #471, opened 2024-05-14
Use .withFluxDensity to normalize the SEDs in instcat.py

jchiang87:
I also cleaned up a couple of no-longer relevant hot-fixes related to the SED normalization in skyCatalogs, and I removed the old, disabled test code in test_instcat_parser.py since it all relied on lsst.sims functionality.

### DIFF
--- a/imsim/instcat.py
+++ b/imsim/instcat.py
@@ -164,10 +164,6 @@ class InstCatalog(object):
 
     The other "phosim commands" are handled by OpsimDataLoader.
     """
-    # Using area-weighted effective aperture over FOV
-    # from https://confluence.lsstcorp.org/display/LKB/LSST+Key+Numbers
-    _rubin_area = 0.25 * np.pi * 649**2  # cm^2
-
     def __init__(self, file_name, wcs, xsize=4096, ysize=4096, sed_dir=None,
                  edge_pix=100, sort_mag=True, flip_g2=True, approx_nobjects=None,
                  pupil_area=RUBIN_AREA, min_source=None, skip_invalid=True,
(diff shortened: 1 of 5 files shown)


imsim/instcat.py
```diff
@@ -359,7 +353,16 @@ def getSED(self, index):
...
             sed = galsim.SED(full_name, wave_type='nm', flux_type='flambda')
-            sed = sed.withMagnitude(0, self._bp500)  # Normalize to mag 0
+
+            # Normalize the SED to magnorm=0 at 500 nm.
+            magnorm = 0.0
+            wl = 500.0*u.nm
+            fnu = (magnorm * u.ABmag).to_value(u.erg/u.s/u.cm**2/u.Hz)
+            flambda = fnu * (astropy.constants.c/wl**2).to_value(u.Hz/u.nm)
+            hnu = (astropy.constants.h * astropy.constants.c / wl).to_value(u.erg)
+            flux_density = flambda / hnu
```
rmjarvis, line 363: These are all constants, so no point in redoing this calculation every time.  Should make this flux_density a class attribute so it only needs to be computed once.

Also, I was playing around with it a bit, and I think this is the cleanest way to compute this quantity with astropy.units:
```
fnu = (0 * u.ABmag).to(u.erg/u.s/u.cm**2/u.Hz)
flux_density = fnu.to_value(u.ph/u.nm/u.s/u.cm**2, u.spectral_density(500*u.nm))
```

rmjarvis (approved):
Just an efficiency suggestion to not repeat the calculation unnecessarily.

PR #472, opened 2024-05-15
Add TableRow

jmeyers314:
Adds a custom InputType for the galsim config machinery to parse an astropy QTable.  My current use case is to grab one row of the table for each AOS operations rehearsal 4 (OR4) exposure (example [here](https://github.com/jmeyers314/aos_block_planner/blob/main/OR4_AOS_schedule.ecsv)), so this is similar to the OpsimData class.  I wanted to be able to leverage units and structured array columns though, and lose some of the other peculiarities of OpSimData.

It's possible this would fit in better in GalSim than ImSim.  Wherever it ends up, it should be part of a tag/release soonish I think for OR4 reproducibility.

### DIFF
--- a/imsim/__init__.py
+++ b/imsim/__init__.py
@@ -36,3 +36,4 @@ from .opd import *
 from .vignetting import *
 from .sag import *
 from .process_info import *
+from .table_row import *
--- a/imsim/table_row.py
+++ b/imsim/table_row.py
@@ -136,9 +136,7 @@ def RowData(config, base, value_type):
     return val, safe
 
 
-RegisterInputType(
-    input_type="table_row", loader=InputLoader(TableRow, file_scope=True)
-)
+RegisterInputType(input_type="table_row", loader=InputLoader(TableRow, file_scope=True))
 RegisterValueType(
     type_name="RowData",
     gen_func=RowData,
--- a/imsim/table_row.py
+++ b/imsim/table_row.py
@@ -1,0 +1,147 @@
+import astropy.units as u
+import galsim
+from astropy.table import QTable
+from galsim.config import (
+    GetAllParams,
+    GetInputObj,
+    InputLoader,
+    RegisterInputType,
+    RegisterValueType,
+)
+
+
+class TableRow:
+    """Class to extract one row from an astropy QTable and make it available to the
+    galsim config layer.
+
+    Parameters
+    ----------
+    file_name: str
+    keys: list
+        Column names to use as keys.
+    values: list
+        Values to match in the key columns.
+    """
+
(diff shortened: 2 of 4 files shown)


tests/test_exposure_table.py
```diff
@@ -0,0 +1,225 @@
...
+
+            # Check multi indexing during table load
+            for idx0 in range(2):
+                for idx1 in range(2):
+                    config["input"]["exposure_table"]["keys"] = ["idx0", "idx1"]
+                    config["input"]["exposure_table"]["values"] = [idx0, idx1]
+                    galsim.config.RemoveCurrent(config["input"]["exposure_table"])
+                    galsim.config.ProcessInput(config)
+                    idx = idx0 * 2 + idx1
+                    check_exp_data(config, idx)
```
rmjarvis, line 219: Rather than just checking explicit calls to ExposureData, it's probably also worth adding some values in the config itself that use the input type.  Things that use `{ 'type': 'ExposureData', 'field': ... }`, since that's the syntax that users will actually use.

rmjarvis (changes requested):
This feels a little to specific a use case to put in GalSim proper.  Probably imSim is a good home for it.  But if you want, you could also submit it to galsim_extra if you think people outside of the imSim user community might want to use it.

tests/test_exposure_table.py
jmeyers314: Sure.  I added a test that actually creates a batoid telescope, similar to what I'm doing for OR4.

imsim/exposure_table.py
```diff
@@ -0,0 +1,148 @@
...
+
+
+RegisterInputType(
+    input_type="exposure_table", loader=InputLoader(ExposureTable, file_scope=True)
+)
+RegisterValueType(
+    type_name="ExposureData",
+    gen_func=ExposureData,
+    valid_types=[float, int, bool, str, galsim.Angle, list],
+    input_type="exposure_table",
```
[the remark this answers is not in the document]
jmeyers314: Sounds good.  I think I actually had AstropyTableRow at one point during development and changed it (can't remember why now).  I landed on TableRow and RowData now.

jmeyers314:
> This feels a little to specific a use case to put in GalSim proper. Probably imSim is a good home for it. But if you want, you could also submit it to galsim_extra if you think people outside of the imSim user community might want to use it.

Sure, ImSim is fine for now.  Can always migrate later if that's useful.

jmeyers314:
Assuming this passes CI, I think it's ready for another look @rmjarvis .

rmjarvis (approved):

PR #473, opened 2024-05-15
Header fixes and LSSTCamSim support

jchiang87:


imsim/readout.py
```diff
@@ -270,16 +269,31 @@ def get_primary_hdu(eimage, lsst_num, camera_name=None,
...
         phdu.header['CCDSLOT'] = sensor
         phdu.header['RA'] = ratel
         phdu.header['DEC'] = dectel
         phdu.header['ROTCOORD'] = 'sky'
+        phdu.header['ROTPA'] = rotang
         telcode = 'MC'
     dayobs = eimage.header['DAYOBS']
     seqnum = eimage.header['SEQNUM']
     contrllr = eimage.header['CONTRLLR']
+    phdu.header['TELESCOP'] = SIMONYI_TELESCOPE
```
jmeyers314, line 286: Maybe factor out some of the other common keywords too?  E.g., 'RAFTBAY', 'CCDSLOT', 'RA', 'DEC', 'ROTCOORD', 'ROTPA', ...?

imsim/readout.py
```diff
@@ -270,16 +269,31 @@ def get_primary_hdu(eimage, lsst_num, camera_name=None,
...
+    phdu.header['TELCODE'] = telcode
+    phdu.header['RASTART'] = ratel
+    phdu.header['DECSTART'] = dectel
+    phdu.header['ELSTART'] = eimage.header['ALTITUDE']
+    phdu.header['AZSTART'] = eimage.header['AZIMUTH']
+    if eimage.header['IMGTYPE'] == 'SKYEXP':
+        phdu.header['RADESYS'] = 'ICRS'
+        phdu.header['TRACKSYS'] = 'RADEC'
+    else:
```
jmeyers314, line 295: When does this get used?  Simulated flats?

jmeyers314 (approved):
LGTM.

imsim/readout.py
jmeyers314: (by this, I mean TRACKSYS = LOCAL).

imsim/readout.py
jchiang87: In order to do that more cleanly, I'd like to get rid of support for `LsstCamImSim`, otherwise I suspect the number of `if` statements will actually increase.  In fact, in addition to getting rid of `LsstCamImSim` support, I'd like to support just `*Sim` cameras explicitly, e.g., `LsstCamSim` and `LsstComCamSim`, so also removing `LsstCam` as an option.  I'll post an issue with this proposal.

imsim/readout.py
jchiang87: Yes, for anything not on-sky, specifically any calibration frames.

PR #480, opened 2024-07-29
satellite streaks from instance catalogs

jchiang87:


rmjarvis:
I think this implementation is fine for the ways that most people want to use streaks.  However, it is probably worth documenting somewhere that it doesn't use the right PSF.  Streaks are not sources at infinity, so they are out of focus, and hence have a different PSF.  This implementation doesn't account for that detail.

cwwalter:
Yes, I happen to be looking at satellites now for some TMA tests with a small telescope we had mounted to it.  The streaks are pretty wide. Can that be adjusted by the "box size".

rmjarvis (approved):

PR #484, opened 2024-12-16
Correct sky_pos and local_wcs in photon op tests

welucas2:
Some of the photon op tests appear not to be self-consistent: the `sky_pos` and `local_wcs` are set independently though both should actually be derived from the image's WCS and the object position within it. This wouldn't be an issue except that the incoming PR #424 implements a new `XyToV` which is self-consistent through use of the `img_wcs` alone and so breaks the regression tests here. This PR updates the tests to calculate the the `sky_pos` and `local_wcs` together from the `image_pos` and `img_wcs`.

Note this isn't an issue in the photon ops themselves, just in the test method. This also changes the expected photon position centre in the regression tests from (-564.5, -1431.4) to (-960.28, -308.09). This is borne out in the tests in PR #424 using the new `XyToV` implementation which preserves these results.

### DIFF
--- a/tests/test_photon_ops.py
+++ b/tests/test_photon_ops.py
@@ -65,6 +65,22 @@ def create_test_photon_array(t=0.0, n_photons=10000):
         time=np.full(n_photons, t),
     )
 
+def create_test_img_wcs(boresight, rottelpos=np.pi / 3 * galsim.radians):
+    telescope = create_test_telescope(rottelpos)
+    wcs_factory = BatoidWCSFactory(
+        boresight,
+        obstime=Time.strptime("2022-08-06 06:50:59.337600", "%Y-%m-%d %H:%M:%S.%f"),
+        telescope=telescope,
+        wavelength=622.3195217611445,  # nm
+        camera=get_camera(),
+        temperature=280.0,
+        pressure=72.7,
+        H2O_pressure=1.0,
+    )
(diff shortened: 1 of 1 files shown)


tests/test_photon_ops.py
```diff
@@ -596,12 +649,15 @@ def test_config_rubin_optics():
                 },
             ]
         },
+        "sky_pos": create_test_img_wcs(
+            boresight=galsim.CelestialCoord(1.1047934165124105 * galsim.radians, -0.5261230452954583 * galsim.radians),
+            rottelpos=np.pi / 3 * galsim.radians
+        ).toWorld(image_pos),
```
rmjarvis, line 655: Slightly cleaner here would be to define boresight once outside the config, and then use both `"boresight": boresight` above in photon_ops and use that same value here for the `boresight` parameter.  Otherwise there are two places where a bunch of random numbers need to be visually confirmed to match up.

rmjarvis (approved):
Thanks William. This all looks right to me.  Just one small suggestion.

tests/test_photon_ops.py
welucas2: Thanks Mike, nice spot! I've made that change now.

welucas2:
Merging as agreed at yesterday's meeting.

PR #487, opened 2025-05-10
ci with cmvfs

jchiang87:


cwwalter (commented):
LGTM

jchiang87:
@cwwalter I think this is ready to go.  I had to change a couple of test conditions slightly in `test_psf.py` and `test_stamp.py`.

cwwalter (approved):
OK LGTM.

GH has to be separately told to use this yml correct?

jchiang87:
Yes, I added it as a separate github action, so after this merge we'll have both this version and the stackvana versions running for each PR.

PR #488, opened 2025-06-24
fix docker build

jchiang87:
* Use `ghcr.io/lsst/scipipe:al9-w_latest` to build against latest weekly
* Use skyCatalogs v2.0.1
* Remove `sncosmo` from conda reqs.


cwwalter (approved):
LGTM.

PR #490, opened 2025-07-17
use skyCatalogs v2.1 in CI testing

jchiang87:


jchiang87:
I've included these changes in `u/jchiang/v2.1_update_docs`

PR #491, opened 2025-07-23
update installation docs for v2.1.0

jchiang87:
Also
* use skyCatalogs v2.1.0 in CI and docs
* handle numpy deprecations


cwwalter (approved):
OK looks good now. Thanks.

PR #493, opened 2025-09-25
New sensor models and auto-switch for ITL or e2v

welucas2:
Craig Lage has provided us with new models of the ITL and e2v sensors used in LSSTCam which use four points along each pixel edge; previous models (provided with GalSim) use 8 and 32 points. Using lower resolution sensor models saves memory, which is particularly helpful for the photon pooling pipeline in which the full detector has to be retained in memory.

I've added all the models, including those we had previously, to the imSim repository. A new pair of tests (heavily inspired by GalSim's test_sensor.py) check the various models and ensure that the spot size doesn't differ more than would be expected across the different model resolutions, and regression tests are there on the second moments and ellipticities of generated in the test images.

Finally, I noticed while doing this that imSim was always using the default `SiliconSensor` model from GalSim, which is the 8-point ITL model. I added a little logic to `lsst_image.py` in the image setup that switches the sensor between the 4-point ITL and e2v models according to which LSSTCam detector is being simulated -- unleass a sensor model name has already been given in the config, in which case that's used instead.

### DIFF
--- a/tests/test_sensor_models.py
+++ b/tests/test_sensor_models.py
@@ -77,14 +77,13 @@ def run_sensor_tests(sensor_type):
     sensor4 = galsim.SiliconSensor(rng=rng4, name=sensor_path('lsst_' + sensor_type + '_50_32'))
     image4 = draw_obj_with_sensor(obj, sensor4, rng4)
 
-    # The max flux in the images using a SiliconSensor model should always be
-    # less than or equal to the flux when not using a sensor or using the simple Sensor.
+    # The max flux in the images using no sensor or a simple Sensor should be
+    # the same. Then we'd expect that the max in the different SiliconSensor
+    # models should always be less than or equal to that.
+    assert image0.array.max() == image1.array.max()
     assert image2.array.max() <= image0.array.max()
     assert image3.array.max() <= image0.array.max()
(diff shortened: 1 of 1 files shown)


rmjarvis:
On my machine with galsim 2.6.1, I get
```
FWHM of reference PSF:  0.9679986011305037
FWHM of FFT PSF:  0.9590216810458915
```
So it's not something with the latest galsim version.  Maybe numpy changed something?  Seems weird, but I can't think what else in the environment would be a likely culprit.  We don't go too crazy with our dependencies in GalSIm, and the AtmosphericPSF class being used in that test doesn't seem to use any lsst stack details.  So I don't think there should be too much in the dependency list that is even relevant to this.

If you can reproduce this locally, it would probably be helpful to try to figure out what env changes make the problem appear.  (Sorry, I know that might be quite tedious.)  Maybe that will give us a clue.

jmeyers314:
Looks like there's a kind="stable" argument to [argsort](https://numpy.org/doc/stable/reference/generated/numpy.argsort.html). Bet that'd help.

jmeyers314:
Ah. I see you already mentioned that in the other thread...

rmjarvis (changes requested):
LGTM.  Just one very minor suggestion.

PR #494, opened 2025-10-06
Remove test using calculateFWHM from test_atm_psf_fft

welucas2:
In #493 we found an issue with `calculateFWHM()` as used in one of the PSF tests. This is now being fixed in GalSim (https://github.com/GalSim-developers/GalSim/pull/1337) but in the meantime @rmjarvis has suggested removing the lines calculating and testing the FWHM in the imSim test as there are already assertions that the size is as expected using `calculateMomentRadius()`. This PR simply removes the references to the FWHM and the assertion.

Once this is merged into main, I'll rebase #493 onto it and push again.


jchiang87 (approved):
LGTM.  Thanks!

PR #495, opened 2025-10-07
Full focal plane infrastructure

welucas2:
This isn't quite ready yet, but I'd like to get a draft PR in place. This will add to imSim the capability to simulate the full focal plane. Currently, we can only simulate detectors in isolation: for a given set of observation parameters and input, imSim will run through tasks drawing each detector in the list. This neglects any 'crosstalk' between detectors in the form of photons that are generated for one but which (for whatever reason) end up raining down on another.

We're building towards a two-pass full focal plane model which will work like the following:
1) Pass one will work like a current imSim run, with the addition of an extra output field `off_detector_photons`. While working on each detector, imSim will gather the photons that end up outside its bounds and write them to file for use in pass two. The other output is the same image that would normally have been produced, of those objects which lie within the detector's bounds and have been drawn on it.
2) Pass two takes both of those outputs from pass one. While working on a detector, it will read its image from pass one as well as the off-detector photons that were originally created by tasks working on other detectors. In this way, the task working on detector X is able to draw the photons that were originally created while working on the other 188 tasks.

As of the moment there are still things to do here:
* The current implementation is for the photon pooling pipeline with `LSST_PhotonPoolingImage`. I'd like to see how easy it would be to implement for the original `LSST_Image` pipeline, too. This might be something to do later on separately.
  * If not doing now, preventing or providing a warning for users who try to use the `off_detector_photons` extra output with the original pipeline.
* Provide a switch when reading `off_detector_photons` in pass two that refactors the input at the very start, prior to drawing the images.
  * Each task in pass one produces one file. which could potentially contain photons destined for all 188 other detectors. So, to be sure that all photons are captured, *not* refactoring the files would require each task in pass two to read and work with 188 files.
  * Refactoring would take the 189 files, each containing all the photons originating from a given detector but destined elsewhere, and produce a different set of 189 files, each containing the photons from any origin but destined for a given detector. With this, each task in pass two would only need to read the one off-detector photon file.
* We should provide template configuration files for users, and probably expand the current `imsim-user-instcat-ffp-pass1.yaml` and `imsim-user-instcat-ffp-pass2.yaml` examples.


examples/imsim-user-instcat-ffp-pass2.yaml
```diff
@@ -0,0 +1,81 @@
...
+    #         - $band
+    #         - $det_name
+    #         - "@output.det_num"
+    file_pattern:
+        type: FormattedStr
+        format: off_det_%08d-%1d-%s-DETNAME-detDETNUM.fits
+        items:
+            - { type: OpsimData, field: observationId }
+            - { type: OpsimData, field: snap }
+            - $band
```
rmjarvis: A maybe simpler API would be to just have `file_name`, but allow `*`.  Then later use `glob.glob(file_name)` to get the list of files matching that pattern.  I use this in Piff when the input can involve multiple files.

imsim/instcat.py
```diff
@@ -244,8 +274,16 @@ def _load(self, logger=None):
...
                     ):
                         continue
+
                     world_pos = galsim.CelestialCoord(ra, dec)
                     #logger.debug('world_pos = %s',world_pos)
+                    if self.full_focal_plane:
+                        # Check whether the object is closer to this detector's center
+                        # than any of the adjacent detectors. If not, then skip it.
+                        obj_dist = {det: detector_centers[det].distanceTo(world_pos) for det in adjacent_detectors}
+                        closest_detector = min(obj_dist, key=obj_dist.get)
```
rmjarvis: This is very nice.  Took me a minute to understand how it worked, but it's quite elegant.

rmjarvis (commented):
Looks good.  Some suggestions here, but I think all minor edits.

imsim/full_focal_plane.py
```diff
@@ -0,0 +1,238 @@
...
+    def photons(self, photon_array):
+        # Allow the photons to be set manually from outside the class if necessary.
+        self._photons = photon_array
+
+    def _read_photons(self, logger=None):
+        logger = galsim.config.LoggerWrapper(logger)
+        if len(self.file_names) > 0:
+            photon_storage = []
+            logger.info(f"Detector {self.det_name} reading {len(self.file_names)} files for off-detector photons.")
+            for file_name in self.file_names:
```
jchiang87, line 165: For a full focal plane simulation, using the diffsky + star catalogs,  I'm finding that the job memory blows up here to at least 40 GBs if we just read in all of the off-detector photons from the other 188 detectors (in my initial test it looks like it OOM'd on a perlmutter login node).  I've made some modifications [here](https://github.com/LSSTDESC/imSim/compare/u/welucas2/full-focal-plane...u/jchiang/ffp_off_det_photon_memory_fixes) to downselect for the CCD being processed.  With this change, the increase in memory is only a few hundred MB in my test.

examples/imsim-user-instcat-ffp-pass2.yaml
```diff
@@ -0,0 +1,81 @@
...
+
+image.type: LSST_FocalPlaneImage
+image.sensor.treering_center: ""
+image.sensor.treering_func: ""
+
+output.det_num.first: 95
+output.nfiles: 1
+output.file_name.format: "eimage_full_focal_%08d-%1d-%s-%s-det%03d.fits"
+
+output.dir: output
```
jchiang87, line 81: Since the sky background and cosmic rays are added to the eimage in pass1, I'd add these lines here:
```
image.sky_level: 0
output.cosmic_ray_rate: 0.0
```
I added these to the [branch](https://github.com/LSSTDESC/imSim/compare/u/welucas2/full-focal-plane...u/jchiang/ffp_off_det_photon_memory_fixes) I made off of this one.

Edit:  Something else I just noticed:  the centroid file from pass1 was overwritten with an empty file as an outcome of pass2.  So I would also add
```
output.photon_pooling_truth: ""
```
to avoid that.

jchiang87 (commented):
Here are some observations and suggested code updates based on my running a full focal plane simulation at NERSC using full diffsky + stars skyCatalogs.

One other thing I noticed after running pass2 on my test CCD (R22_S11):  Objects near the edge of the CCD (but definitely on that CCD) got brighter after pass2 vs pass1, which suggests double counting of those objects.  I can search through the centroid files to see which objects were simulated in pass1 for each CCD, but I'm wondering if it's possible to include object ID information in the photon arrays that are written out to the off-detector files.   That would be useful if we find that we need to filter those photons in pass2.

I confirm that I see the same objects in the pass1 centroid files for adjacent CCDs at the ~5-7% level.  For one pair, I see ~3000 objects in common out of 44k and 48k objects in the two files, respectively.  (This seems to have occurred because I didn't set `input.sky_catalog.full_focal_plane: True` in my pass1 yaml file.  I've noted below that an entry like this should be in the instcat-based yaml example.)

I'm seeing `nan`s for the `x` and `y` values in the off-detector photon files.  It looks like there are just a handful of photons per file (~10), but they are in every file.

examples/imsim-user-instcat-ffp-pass1.yaml
```diff
@@ -0,0 +1,56 @@
...
+
+# Put your own commands that override the defaults below here.  For example
+# input.instance_catalog.file_name: ./imsim_cat_197356.txt
+# input.instance_catalog.sort_mag: False
+# input.tree_rings.only_dets: [R22_S11]
+# image.nobjects: 5
+
+input.instance_catalog.file_name: $os.environ.get('IMSIM_HOME')+'/imSim/examples/example_instance_catalog_ffp.txt'
+input.instance_catalog.sort_mag: False
+
```
jchiang87, line 21: There needs to be an entry
```
input.instance_catalog.full_focal_plane: True
```
here. The same goes for a skycatalogs-based yaml file:
```
input.sky_catalog.full_focal_plane: True
```
Otherwise, the code in the relevant catalog implementations for calculating computing the closest detector won't be executed, and we'll get multiple renderings of objects at the CCD edges.

imsim/skycat.py
```diff
@@ -217,13 +273,30 @@ def getKwargs(self, config, base, logger):
                'approx_nobjects': int,
                'mjd': float,
                'pupil_area': float,
+               'full_focal_plane': bool,
```
jchiang87, line 276: If we specify `output.off_detector_photons` in the yaml config, then I think `full_focal_plane` should be `True`.  Conversely, if we don't do the `full_focal_plane` calculation in the skycat or instcat code, then setting `output.off_detector_photons` in the yaml isn't useful.  So I'd omit `full_focal_plane` from `opt` here, and instead infer the value from the config, e.g., adding code like this below:
```
        try:
            galsim.config.GetFromConfig(base, 'output.off_detector_photons')
            kwargs['full_focal_plane'] = True
        except galsim.config.GalSimConfigError:
            kwargs['full_focal_plane'] = False
```

rmjarvis: I'd prefer full_focal plane remain in the config as an explicit specification, but have it be an error if the value of full_focal_plane is incompatible with the presence or not of off_detector_photons.

jchiang87:
@rmjarvis I think I've largely addressed your comments as well as the items I brought up.  (There's some consolidation of the common code in `instcat.py` and `skycat.py` that I'd like to defer to a different PR.)  Please have a look and let me know if there are any remaining issues.

imsim/full_focal_plane.py
```diff
@@ -128,52 +131,29 @@ def __init__(self, camera, det_name, file_name=None, file_pattern=None, photons=
...
-            # where it's controlled with det_num.first, nfiles and only_dets.
-            det_num_start = 0
-            det_num_end = 189
-            file_names = [file_pattern.replace("DETNAME", camera[i].getName()).replace("DETNUM", "{:03d}".format(i))
-                          for i in range(det_num_start, det_num_end)
-                          if camera[i].getName() != det_name
-                          ]
+            if dir is not None:
+                file_name = os.path.join(dir, file_name)
+            self.file_names = sorted(fname for fname in glob.glob(file_name))
```
rmjarvis: I would have thought simply `sorted(glob.glob(file_name))` would suffice here.

rmjarvis (approved):
Sorry to take so long on this.  I just have a slight simplification suggestions.  The code looks fine to me.

PR #503, opened 2025-12-15
set the default camera to LsstCamSim and remove non *Sim camera types

jchiang87:
These changes are a belated follow up to #473 :
* Set default camera to `LsstCamSim` in tests and yaml configs.
* Remove support for `LsstCam` and `LsstCamImSim`.

### DIFF
--- a/examples/flat.yaml
+++ b/examples/flat.yaml
@@ -75,7 +75,7 @@ output:
     nproc: 1    # Change this to work on multiple CCDs at once.
     nfiles: 1   # Default is all 189 CCDs.  Set to 1 while testing.
 
-    camera: LsstCam
+    camera: LsstCamSim
 
     exptime: 30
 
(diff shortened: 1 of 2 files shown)


imsim/bandpass.py
```diff
@@ -89,7 +89,7 @@ def RubinBandpass(
     if (camera is None) != (det_name is None):
         raise ValueError("Must provide both camera and det_name if using one.")
     match camera:
-        case 'LsstCam' | 'LsstCamSim' :
+        case 'LsstCamSim' :
             camera = 'lsstCam'
```
cwwalter: I have a hard time keeping up with the names... but just checking:  We are supposed to use the normal lsstCam here?

cwwalter (commented):
LGTM.

But (along with a question) I found one possible missing camera change I called out.

imsim/bandpass.py
cwwalter: Ah I see... do you prefer to leave this as is? Or rename the directory?

PR #516, opened 2026-04-02
Update tree ring parameters based on LSSTCam data

jchiang87:
In this PR, I've performed the following steps to update the model tree ring parameters for each science CCD:

- Boxcar smoothed the LSSTCam tree ring profile data provided by HyeYun on the 10 pixel scale to reduce the noise, while retaining tree ring scales of interest.
- Computed the standard deviation, `data_std`, of the LSSTCam pixel values in the radius range 4500-5200 pixel units.  This will be used as a measure of the tree ring amplitudes.
- Computed the predicted model profiles of the current tree ring model (i.e., `tree_ring_parameters_2018-04-26.txt`) for the CCD at issue evaluated at the same radius locations as the LSSTCam data.
- Computed the standard deviation, `model_std`, of the tree ring model profiles in the 4500-5200 radius range.
- Rescaled the `A` and `B` parameters of the tree ring model radial scaling function, `A + B * r**4`, by the ratio `data_std/model_std`. The rescaled `A` and `B` values are updated in the new tree ring parameter model.
- Replaced the tree ring centers with the tree ring centers measured by HyeYun from the LSSTCam data (where they are available).
- Wrote the updated tree ring parameters to `tree_ring_parameters_2026-04-02.txt`.

This procedure has been implemented in the `refit_treerings.py` script. I've updated the `treerings.py` module to provide access to the derivative of the pixel distortion function, `dfdr`, which is the predicted pixel count modulation from the effect of the tree ring distortions. I've also added interfaces to update the tree ring parameters and to write the modified model to a text file with the expected format.

The following plots show the boxcar smoothed LSSTCam tree ring profiles (blue curves) for a selection of LSSTCam CCDs, compared with the rescaled model profiles (orange curves).  The vertical dotted lines indicate the range over which the standard deviations were computed for the LSSTCam data and the model profiles.  The corner legend shows the scaling factor that was applied to the A and B parameters.
R33_S10_tree_ring_profile
R23_S02_tree_ring_profile

The plot below shows the measured tree ring profile from a simulated flat using the new parameters (blue curve) with the corresponding model prediction overlaid (orange curve).
R22_S11_simulation


rmjarvis (approved):
LGTM.  Math all seems correct. (dfdr in particular).

PR #518, opened 2026-06-09
Cleanup some test errors and warnings.

rmjarvis:
I noticed that my Mac wasn't passing tests and have a bunch of warnings.  Some of the warnings are in dependencies, so I didn't bother with them, but others are actionable.  Also, one of the test errors was happening in the CI system too, not just on Mac.

This PR includes the following fixes:

1. Unpin the conda version, which was causing problems installing conda on the CI system.
2. Fixed some of the flux regression tests in test_image.py to match the current results.  These were failing on the CI as well as my Mac.  I think due to changes in the dependencies.  The LSST stack was part of the difference, but not the whole thing.  Not sure which other dependencies were relevant, but I updated several values to be the average of the linux and Mac values, and I relaxed the tolerance to 5 sigma, rather than 4, so hopefully it will stay passing.
3. Fixed the boresight usage in test_photon_ops.py.  I'm actually surprised this passed on the CI system, since there were inconsistent boresights used by the WCS and the icrs function.  Now the test makes sure to use a consistent boresight for both, which matches whatever boresight is used for each test.  (There are two choices, which are used among the various tests in this file.)
4. Relaxed a cross_rot_angle test that was at rtol=0.03, but on my laptop needed 0.033.  I bumped the tolerance to 0.04.
5. Catch the expected warning for missing tree ring information in one of the tests that exercises this path on purpose, so now it's not so noisy.
6. Set some defaults for batoid_rubin, since the default values will change in a future version.

There are still some warnings emitted by batoid and skycatalogs, but I left those alone.  The correct place to fix them is in those packages.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -54,7 +54,7 @@ jobs:
               uses: conda-incubator/setup-miniconda@v2
               with:
                   auto-update-conda: false
-                  conda-version: "25.11.1"
+                  #conda-version: "25.11.1"
                   activate-environment: stack
                   python-version: ${{ matrix.py }}
                   condarc-file: etc/.condarc
--- a/.gitignore
+++ b/.gitignore
@@ -98,6 +98,7 @@ ENV/
 bin/
 .sconf_temp
 .sconsign.dblite
+AGENTS.md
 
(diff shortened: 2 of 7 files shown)


jchiang87 (approved):
lgtm

